PR #11, opened 2012-10-16
Noop Extractor

coderoshi:
https://github.com/rzezeski/yokozuna/issues/7

### DIFF
--- a/src/yz_noop_extractor.erl
+++ b/src/yz_noop_extractor.erl
@@ -18,15 +18,8 @@
 %%
 %% -------------------------------------------------------------------
 
-%% @doc A noop extractor. Creates no index field for the
-%%      registered mime type.
-%%
-%% Example:
-%%   
-%%   Value is opaque data, such as an MPG file. Extraction
-%%   will do nothing, and return:
-%%   
-%%   []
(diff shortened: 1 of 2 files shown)


src/yz_noop_extractor.erl
```diff
@@ -0,0 +1,43 @@
...
+%% "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+%% KIND, either express or implied.  See the License for the
+%% specific language governing permissions and limitations
+%% under the License.
+%%
+%% -------------------------------------------------------------------
+
+%% @doc A noop extractor. Creates no index field for the
+%%      registered mime type.
+%%
```
rzezeski: For the second sentence just say that this extractor performs no work and always returns the empty list.

src/yz_noop_extractor.erl
```diff
@@ -0,0 +1,43 @@
...
+%% Example:
+%%   
+%%   Value is opaque data, such as an MPG file. Extraction
```
rzezeski: The value can be anything, doesn't have to be opaque data.  The user could just as well register this extractor for JSON data.  I'd say just remove the example part.  See above comment.

PR #13, opened 2012-10-24
Creation of indexes via HTTP

rzezeski:
Made a PR on behalf of @coderoshi so I can comment on the code.

This PR addresses issue #8.

### DIFF
--- a/misc/bench/src/yz_driver.erl
+++ b/misc/bench/src/yz_driver.erl
@@ -15,7 +15,7 @@
          mfa_valgen_i/1]).
 
 -include_lib("basho_bench/include/basho_bench.hrl").
--record(state, {pb_conns, index, iurls, surls}).
+-record(state, {pb_conns, index, rurls, surls}).
 -define(DONT_VERIFY, dont_verify).
 
 %% ====================================================================
@@ -27,17 +27,19 @@ new(_Id) ->
     Index = basho_bench_config:get(index, "test"),
     HTTP = basho_bench_config:get(http_conns, [{"127.0.0.1", 8098}]),
     PB = basho_bench_config:get(pb_conns, [{"127.0.0.1", 8087}]),
-    IPath = basho_bench_config:get(index_path, "/riak/test"),
+    RPath = basho_bench_config:get(riak_path, "/riak/test"),
     SPath = basho_bench_config:get(search_path, "/search/test"),
-    IURLs = array:from_list(lists:map(make_url(IPath), HTTP)),
+    IPath = basho_bench_config:get(index_path, "/yz/index/test"),
+    RURLs = array:from_list(lists:map(make_url(RPath), HTTP)),
     SURLs = array:from_list(lists:map(make_url(SPath), HTTP)),
+    IURLs = array:from_list(lists:map(make_url(IPath), HTTP)),
     Conns = array:from_list(lists:map(fun make_conn/1, PB)),
     N = length(HTTP),
     M = length(PB),
 
     {ok, #state{pb_conns={Conns, {0,M}},
                 index=list_to_binary(Index),
-                iurls={IURLs, {0,N}},
+                rurls={RURLs, {0,N}},
                 surls={SURLs, {0,N}}}}.
 
 run(search, _KeyGen, ValGen, S=#state{surls=URLs}) ->
@@ -69,22 +71,22 @@ run({search, Qry, FL, Expected}, _, _, S=#state{surls=URLs}) ->
             {error, Reason, S2}
     end;
 
-run({index, CT}, _KeyGen, ValGen, S=#state{iurls=URLs}) ->
+run({index, CT}, _KeyGen, ValGen, S=#state{rurls=URLs}) ->
     Base = get_base(URLs),
     {Key, Line} = ValGen(index),
     Key2 = mochiweb_util:quote_plus(Key),
     URL = ?FMT("~s/~s", [Base, Key2]),
-    S2 = S#state{iurls=wrap(URLs)},
+    S2 = S#state{rurls=wrap(URLs)},
     case http_put(URL, CT, Line) of
         ok -> {ok, S2};
         {error, Reason} -> {error, Reason, S2}
     end;
 
-run(load_fruit, KeyValGen, _, S=#state{iurls=URLs}) ->
+run(load_fruit, KeyValGen, _, S=#state{rurls=URLs}) ->
     Base = get_base(URLs),
     {Key, Val} = KeyValGen(),
     URL = ?FMT("~s/~p", [Base, Key]),
-    S2 = S#state{iurls=wrap(URLs)},
+    S2 = S#state{rurls=wrap(URLs)},
     case http_put(URL, "text/plain", Val) of
         ok -> {ok, S2};
         {error, Reason} -> {error, Reason, S2}
--- a/misc/bench/src/yz_driver.erl
+++ b/misc/bench/src/yz_driver.erl
@@ -15,7 +15,7 @@
          mfa_valgen_i/1]).
 
 -include_lib("basho_bench/include/basho_bench.hrl").
--record(state, {pb_conns, index, iurls, surls}).
+-record(state, {pb_conns, index, rurls, surls, iurls}).
 -define(DONT_VERIFY, dont_verify).
 
 %% ====================================================================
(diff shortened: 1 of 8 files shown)


misc/bench/src/yz_driver.erl
```diff
@@ -27,18 +27,21 @@ new(_Id) ->
     Index = basho_bench_config:get(index, "test"),
     HTTP = basho_bench_config:get(http_conns, [{"127.0.0.1", 8098}]),
     PB = basho_bench_config:get(pb_conns, [{"127.0.0.1", 8087}]),
-    IPath = basho_bench_config:get(index_path, "/riak/test"),
+    RPath = basho_bench_config:get(riak_path, "/riak/test"),
     SPath = basho_bench_config:get(search_path, "/search/test"),
-    IURLs = array:from_list(lists:map(make_url(IPath), HTTP)),
+    IPath = basho_bench_config:get(index_path, "/yz/test"),
```
rzezeski: This should be `yz/index/test`, no?

misc/bench/src/yz_driver.erl
```diff
@@ -15,7 +15,7 @@
          mfa_valgen_i/1]).
 
 -include_lib("basho_bench/include/basho_bench.hrl").
--record(state, {pb_conns, index, iurls, surls}).
+-record(state, {pb_conns, index, rurls, surls, iurls}).
```
rzezeski: I notice `iurls` isn't used anywhere.  Why add it?

src/yz_solr.erl
```diff
@@ -24,7 +24,8 @@
 
 -define(CORE_ALIASES, [{index_dir, instanceDir},
                        {cfg_file, config},
-                       {schema_file, schema}]).
+                       {schema_file, schema},
+                       {remove_index, deleteIndex}]).
```
rzezeski: I think we might want to use `deleteInstanceDir` so that everything is killed.  I'm not sure, I haven't tested yet.

src/yz_index.erl
```diff
@@ -87,6 +99,17 @@ local_create(Ring, Name) ->
...
 
+%% @doc Remove the index `Name' locally.
+-spec local_remove(string()) -> ok.
+local_remove(Name) ->
+    CoreProps = [
+                 {core, Name},
+                 {remove_index, "true"}
+                ],
+    {ok, _, _} = yz_solr:core(remove, CoreProps),
+    IndexDir = index_dir(Name),
```
rzezeski: I think using `deleteInstanceDir` with Solr's core UNLOAD would do this for you.

src/yz_index.erl
```diff
@@ -123,6 +146,19 @@ add_to_ring(Name, Info) ->
...
+remove_index(Indexes, Name) ->
+    orddict:erase(Name, Indexes).
+
+-spec remove_from_ring(index_name()) -> ok.
+remove_from_ring(Name) ->
+    Indexes = get_indexes_from_ring(yz_misc:get_ring(transformed)),
+    Indexes2 = remove_index(Indexes, Name),
+    yz_misc:set_ring_meta(?YZ_META_INDEXES, Indexes2),
+    % XXX: ring event is not fired
+    ok = local_remove(Name),
```
rzezeski: Don't call `local_remove` here.  Do the same thing as `local_create` and call it during a ring event.  You do it this way because the ring is the authoritative source.

src/yz_wm_index.erl
```diff
@@ -0,0 +1,232 @@
...
+%% software distributed under the License is distributed on an
+%% "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+%% KIND, either express or implied.  See the License for the
+%% specific language governing permissions and limitations
+%% under the License.
+%%
+%% -------------------------------------------------------------------
+
+
+%% @doc Resource for serving Riak objects over HTTP.
```
rzezeski: copy/paste monster

src/yz_wm_index.erl
```diff
@@ -0,0 +1,232 @@
...
+%%
+%% Available operations:
+%% 
+%% GET /yz/index
+%%   Get information about every index in this ring in JSON format.
```
rzezeski: Mentioning the ring here confuses things.  Not sure I understand the second statement.  Is it one day going to be information different from `yz/index/Index`?

src/yz_wm_index.erl
```diff
@@ -0,0 +1,232 @@
...
+%% GET /yz/index/Index
+%%   Gets information about a specific index in JSON format.
+%%   Returns the following information:
+%%   {
+%%      "name"  : IndexName,
+%%      "bucket": IndexName,
+%%      "schema": SchemaName
+%%   }
+%%   IndexName is the same value passed into the URL. Schema
+%%   is the solr schema type already installed on the server,
```
rzezeski: Schema is the name of the schema to associate with the index.

src/yz_wm_index.erl
```diff
@@ -0,0 +1,232 @@
...
+%% DELETE /yz/index/Index
+%%   Deletes the index with the given index name.
+%%   
+
+-module(yz_wm_index).
+-compile(export_all).
+-include("yokozuna.hrl").
+-include_lib("webmachine/include/webmachine.hrl").
+
+-record(ctx, {api_version,  %% atom() - Determine which version of the API to use.
```
rzezeski: If you're going to spec there then use the `:: type()` convention.  E.g. https://github.com/rzezeski/yokozuna/blob/master/include/yokozuna.hrl#L25

src/yz_wm_index.erl
```diff
@@ -0,0 +1,232 @@
...
+
+%% Responds to a PUT request by creating an index
+%% and hook for the "index" name given in the route
+%% Returns "ok" if created, or "exists" if an index
+%% of that name already exists. Returns a 500 error if
+%% the given schema does not exist.
+create_index(RD, S) ->
+    IndexName = S#ctx.index_name,
+    BodyProps = S#ctx.props,
+    case proplists:get_value(<<"schema">>, BodyProps) of
```
rzezeski: Use `proplists:get_value/3`

src/yz_wm_index.erl
```diff
@@ -0,0 +1,232 @@
...
+        undefined  -> SchemaName = ?YZ_DEFAULT_SCHEMA_NAME;
+        SchemaName -> SchemaName = SchemaName
+    end,
+    BodyResp = wrq:set_resp_header("Content-Type", "text/plain", RD),
+    case create_install_index_if_exists(IndexName, SchemaName) of
+        notfound ->
+            {{halt, 500},
+                wrq:set_resp_header("Content-Type", "text/plain",
+                    wrq:append_to_response_body(
+                        io_lib:format("Schema does not exist~n",[]),
```
rzezeski: Return the schema name in the response

src/yz_wm_index.erl
```diff
@@ -0,0 +1,232 @@
...
+%% of that name already exists. Returns a 500 error if
+%% the given schema does not exist.
+create_index(RD, S) ->
+    IndexName = S#ctx.index_name,
+    BodyProps = S#ctx.props,
+    case proplists:get_value(<<"schema">>, BodyProps) of
+        undefined  -> SchemaName = ?YZ_DEFAULT_SCHEMA_NAME;
+        SchemaName -> SchemaName = SchemaName
+    end,
+    BodyResp = wrq:set_resp_header("Content-Type", "text/plain", RD),
```
rzezeski: You could change `BodyResp` to `RD2` and pass it to calls below to save some space and repetition.

src/yz_wm_index.erl
```diff
@@ -0,0 +1,232 @@
...
+    case (RDBody == <<>>) or (RDBody == []) of
+      true  -> [];
+      false -> 
+          case mochijson2:decode(RDBody) of
+              {struct, BodyData} -> BodyData;
+              _ -> []
+          end
+    end.
+
+% If the index exists, return "exists".
```
rzezeski: Single `%` comments are for EOL.  These should be `%%`.

misc/bench/src/yz_driver.erl
coderoshi: Yes. These driver changes were a failed experiment. I didn't mean to check them in.

src/yz_index.erl
coderoshi: I understand, and this is the line I wanted to ask you about. The event never gets fire, even after setting the ring meta. I wanted to ask you if there was another step I'm missing here

src/yz_wm_index.erl
coderoshi: I appreciate your pedantry :+1:

Issue #1, opened 2012-08-14
colocated replicas squash each other


rzezeski:
## Why is it bad?

Collocated replicas cause **incorrect and non-deterministic** query
results.
## When can it happen?

Any time the number of active nodes is less than the `target_n_val`.
Given the default value of `4` this could happen with a cluster of 3
or less.  It could happen in a 5 node cluster when 2 nodes go down and
new or existing data is written.  It could happen in a 10 node cluster
where a partition occurs leaving two cluster of 7 and 3.  The 7 node
partition won't have collocated replicas because it has more than
`target_n_val` nodes but the 3 node side doesn't.
## What is the cause?

The fundamental cause is the assumption that each partition has a
dedicated backend instance.  Normally, every partition has it's own
instance of bitcask, leveldb, etc.  Given a preflist where two
replicas are on the same node there are still two distinct copies of
that object on two different instances of the backend.  Partitions do
not share backend instances.

Yokozuna **DOES** share one Solr instance and one Core instance **for
all partitions** on the node.  By default, each Solr document must
have a unique identifier.  When a document is written with an existing
id then it overwrites the old one.  Yokozuna writes documents using
the Riak key as the unique id and a field that lists the **single
owning partition**.  The following is an example.

```
<doc>
  <id>riak_key</id>
  <_pn>2<_pn>
  <text>I'll take bolth, please.</text>
</doc>
```

Querying will work as long as this is the only replica of key
`riak_key` on this node.  But if another replica is written then it
will squash the previous value for `_pn`.  This means that queries
which filter by partition 2 will come up empty when they shouldn't.

rzezeski:
## What is the fix?

The crux of the issue is that the Riak key is being used as a unique
key in Solr but it's not unique among partitions.  This seemed like
the most straight forward way to key documents in Solr but it causes
issues as described.

The easiest fix is to **use a concatenation of the Riak key and
partition number as the unique key**.  In order to still have access
to the Riak key without doing string manipulation a new field, `_rk`
(Riak Key), will be added.  Now the doc will look like the following.

```
<doc>
  <id>riak_key_2</id>
  <_rk>riak_key</_rk>
  <_pn>2</_pn>
  <text>I'll take bolth, please.</text>
</doc>
```

**This will prevent collocated replicas from stomping on each other
while still working with the current query filter.**

Issue #3, opened 2012-10-05
remove vnode


rzezeski:
Since yokozuna piggy-backs off the KV vnode there is no need to have a vnode in yokozuna.  Removing the vnode will cause the yokozuna service to never be marked as up.  I'm not sure that matters but even so a call can be made directly to `riak_core_node_watcher:service_up`.

rzezeski:
Whoops.  I removed the vnode but forgot to close this issue.

PR #2, opened 2012-10-04
Active anti entropy

rzezeski:
TODO:
- Rebase my KV branch against latest AAE branch when it lands.  Search code for differences between current branch.
- Implement `yz_index_hashtree:remove` so that ownership changes don't cause undef failures.
- Rename `yz_solr:get_vclocks` to `get_entropy_data`.
- Move date stuff from `yz_entropy` to `yz_misc`.
- Change `get_index_n` to use cached `{Bucket,N}` map.
- Make the `before` argument part of the `Filter` passed to `yz_solr:get_entropy_data`.
- Make entropy data chunk size configurable.
- Add `yz_index_hashtree:insert` and `delete` calls to `yz_kv:index` hook. (Currently it relies on the full build and is not updating tree on fly)
- Move logical partition stuff into `yz_misc`.
- Grep code for TODOs.

PR #8, opened 2012-10-24
Index Creation via HTTP

coderoshi:
For Yokozuna to be useful it needs to support the ability to perform
various administration tasks remotely.  This issue deals with creating
an index via HTTP.  However, it may lay the ground work for other
administration tasks such as adding/modifying schemas.
## Specification
- The resource should be something like `<host-port>/yokozuna/index/<index_name>`.
### PUT
- Create the index if it doesn't already exist.
- If it does exist then return something indicating it already exists.
  Changing the schema of an index which contains data is not simple.
  Fields could be removed.  The associated field type for a given name
  could change.  The analyzer for a given field type could change.
  Etc.  If a user wants to change the schema then I think that will
  require deleting the index, creating it anew with a different
  schema, and then letting AAE re-index.
- The body may optionally contain JSON providing arguments to pass to
  index creation.
- Block until index creation has finished.
### DELETE

Delete the index.  The code to delete an index needs to be written.

TODO: Should it be smart and check that nothing uses this index first
or just assume the user knows what is best?
- Block until deletion is complete.
### GET

Return JSON representing information about index.
### JSON Body
- `schema_name` - The name of the schema to associate with the index

### DIFF
--- a/src/yz_wm_index.erl
+++ b/src/yz_wm_index.erl
@@ -1,0 +1,233 @@
+%% -------------------------------------------------------------------
+%%
+%% Copyright (c) 2012 Basho Technologies, Inc.  All Rights Reserved.
+%%
+%% This file is provided to you under the Apache License,
+%% Version 2.0 (the "License"); you may not use this file
+%% except in compliance with the License.  You may obtain
+%% a copy of the License at
+%%
+%%   http://www.apache.org/licenses/LICENSE-2.0
+%%
+%% Unless required by applicable law or agreed to in writing,
+%% software distributed under the License is distributed on an
+%% "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+%% KIND, either express or implied.  See the License for the
+%% specific language governing permissions and limitations
+%% under the License.
+%%
+%% -------------------------------------------------------------------
+
+
+%% @doc Resource for managing Yokozuna Indexes over HTTP.
+%%
+%% Available operations:
+%% 
+%% GET /yz/index
+%%   Get information about every index in JSON format.
+%%   Currently the same information as /yz/index/Index,
+%%   but as an array of JSON objects.
+%%   
+%% GET /yz/index/Index
+%%   Gets information about a specific index in JSON format.
+%%   Returns the following information:
+%%   {
+%%      "name"  : IndexName,
+%%      "bucket": IndexName,
+%%      "schema": SchemaName
+%%   }
+%%   IndexName is the same value passed into the URL. Schema
+%%   is the name of the schema associate with this index. That
+%%   schema file must already be installed on the server.
+%%   Defaults to "_yz_default".
+%%
+%% PUT /yz/index/Index
+%%   Creates a new index with the given name, and also creates
+%%   a post commit hook to a bucket of the same name.
+%%   A PUT request requires this header:
+%%     Content-Type: application/json
+%%   A JSON body may be sent. It currently only accepts
+%%   { "schema" : SchemaName }
+%%   If no "schema" is given, it defaults to "_yz_default".
+%%   Returns a '409 Conflict' code if the index already exists.
+%%
+%% DELETE /yz/index/Index
+%%   Deletes the index with the given index name.
+%%   
+
+-module(yz_wm_index).
+-compile(export_all).
+-include("yokozuna.hrl").
+-include_lib("webmachine/include/webmachine.hrl").
+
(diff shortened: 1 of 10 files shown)


rzezeski:
Revert the changes to `yz_driver` completely since they aren't being used.

src/yz_wm_index.erl
```diff
@@ -0,0 +1,230 @@
...
+%% Returns "ok" if created, or "exists" if an index
+%% of that name already exists. Returns a 500 error if
+%% the given schema does not exist.
+create_index(RD, S) ->
+    IndexName = S#ctx.index_name,
+    BodyProps = S#ctx.props,
+    SchemaName = proplists:get_value(<<"schema">>, BodyProps, ?YZ_DEFAULT_SCHEMA_NAME),
+    case create_install_index_if_exists(IndexName, SchemaName) of
+        notfound ->
+            {{halt, 500},
```
rzezeski: I think the schema name should be checked in malformed request and return 400 if it doesn't exist.

src/yz_wm_index.erl
```diff
@@ -0,0 +1,230 @@
...
+                wrq:set_resp_header("Content-Type", "text/plain",
+                    wrq:append_to_response_body(
+                        io_lib:format("Schema ~p does not exist~n",[SchemaName]),
+                        RD)),
+                S};
+        Body ->
+            {Body,
```
rzezeski: I'm of the opinion that the code is easier to read if the RD is built outside of the returned tuple.  E.g.

``` erlang
RD = wrq:set_resp_header("content-type", "text/plain", ...),
{Body, RD, S}
```

Also, why is the body both set explicitly and part of the return code?

src/yz_wm_index.erl
```diff
@@ -0,0 +1,230 @@
...
+                wrq:set_resp_header("Content-Type", "text/plain",
+                    wrq:append_to_response_body(Body, RD)),
+                S}
+    end.
+
+%% Responds to a GET request by returning index info for
+%% the given index as a JSON response.
+read_index(RD, S) ->
+    {ok, Ring} = riak_core_ring_manager:get_my_ring(),
```
rzezeski: Use `yz_misc:get_ring(transformed)`

src/yz_wm_index.erl
```diff
@@ -0,0 +1,230 @@
...
+    case S#ctx.index_name of
+        undefined  ->
+            Indexes = yz_index:get_indexes_from_ring(Ring),
+            Details = {array,[index_body(Ring, IndexName)
+              || IndexName <- orddict:fetch_keys(Indexes)]};
+        IndexName ->
+            Details = index_body(Ring, IndexName)
+    end,
+    {mochijson:encode(Details), RD, S}.
```
rzezeski: Use mochijson2.  This switch means that values must be binary and an array is just a list.

coderoshi: I actually switched to mochijson from mochijson2. the actual output sent to the client was "erlangified". Single character strings were converted to integers. I couldn't find a way to prevent it.

rzezeski: That's because in mochijson2 a list is considered an array and in erlang a string is just a list of integers.  If you want a string the value must be a binary.

rzezeski:
- Deleting an existing index removes the index but doesn't remove the hook.  You'll want to look at `riak_kv_vnode:add_obj_modified_hook` and instead of adding to the `Hooks` list you'll want to remove and call `riak_core_bucket:set_bucket`.  Don't worry about putting it in KV, just put it in `yz_kv` for now.  Also, removing the hook should be done first.
- A PUT for an existing index returns a 204.  The problem with this is that the user could set a different schema name and the 204 gives the expectation that the schema name was changed.  This isn't the case and it shouldn't be the case until thought has been giving to modifying schemas on the fly.  At this point if the index already exist a 400 or 409 should be returned (or maybe some other status code I'm not aware of).
- I notice we return `content-type` with our 204 responses.  Probably doesn't hurt anything but seems odd.  Furthermore some of the 204s return `text/plain` and some `application/json`.

rzezeski:
Don't fret, you are almost there :)

coderoshi:
I resolved #1 and #2 in your list, but I'm not sure about half of #3. delete_resource seems to inherit from content_types_provided, and it's not clear one how to remove that header. PUT resulting in 204 was a bug.

rzezeski:
+1 to merge.  Great job!

PR #16, opened 2012-10-28
JSON extractor that adds suffixes to fields based on value type

dreverri:
The extractor behaves very much like the existing JSON extractor. Arrays create multi-value fields and nested object field names are concatenated by a field separator (`_` by default). The only difference is the field names are suffixed according to the value type. The suffixes added to fields by this extractor correspond to the dynamic field definitions in the default schema.
- String - `*_t`
- Number - `*_d`
- Boolean - `*_b`
- Multi value
  - String - `*_txt`
  - Number - `*_ds`
  - Boolean - `*_bs`

Also, the function `yz_solr:prepare_xml/1` has been added to allow tests to generate XML data from extracted fields. While indexing documents I found that boolean and number values were not generated correctly. All values should be converted to binary in the extractor or the XML data will be incorrect. The included test extracts fields from sample JSON data and generates XML data.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,4 @@ priv/lucene-solr
 riak_test/ebin
 tests/20*
 tests/current
+.eunit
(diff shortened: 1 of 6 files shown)


rzezeski:
It would be nice to reuse the existing JSON extractor and then transform the result but this approach is fine for now.  I'll be making some changes to extraction, in general, this month to support Solr Cell and maybe I'll try adding this as a feature of the regular json extractor.

Anyways, tests pass, dialyzer doesn't complain.  +1 to merge.

PR #21, opened 2012-11-02
Encapsulate set_instance_data function by pass in client_token. Allows skipping execution of previous lines if need be.

coderoshi:


rzezeski:
+1 to merge

PR #4, opened 2012-11-16
handle object siblings

coderoshi:
Currently Yokozuna punts on object siblings.  It grabs all the values and takes the head.  It would be ideal to index all siblings so that a query matching any of them returns the object key.  There are probably multiple ways to handle this.  Here are two potential solutions off the cuff.
1. Use multiValued fields for everything.  All siblings would generate one document but that document could contain same field many times.
2. Generate a document for each sibling.  This would require changing the solr id key to include a sibling identifier.  After sibling resolution how are the documents cleaned up?  You could probably do delete by query and use `_yz_rk` (riak key) field.

### DIFF
--- a/src/yz_doc.erl
+++ b/src/yz_doc.erl
@@ -45,7 +45,7 @@ doc_id(O, Partition, Sibling) ->
 has_siblings(O) -> riak_object:value_count(O) > 1.
 
 %% @doc Given an object generate the doc to be indexed by Solr.
--spec make_docs(riak_object:riak_object(), binary(), binary()) -> doc().
+-spec make_docs(riak_object:riak_object(), binary(), binary()) -> [doc()].
 make_docs(O, FPN, Partition) ->
     [make_doc(O, Content, FPN, Partition) || Content <- riak_object:get_contents(O)].
 
@@ -98,7 +98,7 @@ extract_fields({MD, V}) ->
 %% @private
 %%
 %% @doc Extract tags from object metadata.
--spec extract_tags(obj()) -> fields().
+-spec extract_tags(dict()) -> fields().
 extract_tags(MD) ->
     MD2 = get_user_meta(MD),
     TagNames = get_tag_names(MD2),
--- a/src/yz_doc.erl
+++ b/src/yz_doc.erl
@@ -22,6 +22,8 @@
 -compile(export_all).
 -include("yokozuna.hrl").
 
+-define(MD_VTAG, <<"X-Riak-VTag">>).
+
 %% @doc Functionality for working with Yokozuna documents.
 
 %%%===================================================================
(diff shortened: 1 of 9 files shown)


src/yz_doc.erl
```diff
@@ -31,33 +33,59 @@
 add_to_doc({doc, Fields}, Field) ->
     {doc, [Field|Fields]}.
 
--spec doc_id(riak_object:riak_object(), binary()) -> binary().
+% -spec doc_id(riak_object:riak_object(), binary()) -> binary().
```
rzezeski: Why was this spec commented out?

src/yz_doc.erl
```diff
@@ -31,33 +33,59 @@
...
 doc_id(O, Partition) ->
     <<(riak_object:key(O))/binary,"_",Partition/binary>>.
+doc_id(O, Partition, none) ->
```
rzezeski, line 40: Add a space here so it's easier to tell it's two different functions.

src/yz_kv.erl
```diff
@@ -83,30 +82,51 @@ get_md_entry(MD, Key) ->
...
+    {Bucket, Key} = BKey = {riak_object:bucket(Obj), riak_object:key(Obj)},
+    BProps = riak_core_bucket:get_bucket(Bucket, Ring),
+    AllowMult = proplists:get_value(allow_mult, BProps, false),
+    Partition = get_partition(VNodeState),
+    LP = yz_cover:logical_partition(LI, Partition),
     DocID = binary_to_list(yz_doc:doc_id(Obj, ?INT_TO_BIN(LP))),
     try
-        yz_solr:delete(binary_to_list(Bucket), DocID)
+        yz_solr:delete(binary_to_list(Bucket), DocID),
+        delete_siblings_by_key(Obj, Bucket, Key, AllowMult)
```
rzezeski: We actually don't need both deletes.  Deleting by the riak key is a superset of deleting by doc id.  Let's get rid of the first delete call.

src/yz_solr.erl
```diff
@@ -228,8 +229,12 @@ convert_action(remove) -> "UNLOAD".
 encode_commit() ->
 	xmerl:export_simple([{commit, []}], xmerl_xml).
 
+encode_delete({key,Key})->
+    Query = "_yz_rk:" ++ binary_to_list(Key) ++ " AND _yz_vtag:[* TO *]",
```
rzezeski, line 236: The `_yz_vtag` field is not needed here.  No?

src/yz_kv.erl
```diff
@@ -83,30 +82,51 @@ get_md_entry(MD, Key) ->
...
+    Docs = yz_doc:make_docs(Obj, ?INT_TO_BIN(LFPN), ?INT_TO_BIN(LP)),
+
     try
-        ok = yz_solr:index(binary_to_list(Bucket), [Doc])
+        ok = yz_solr:index(binary_to_list(Bucket), Docs),
+        case riak_object:value_count(Obj) of
+            2 ->
+                case AllowMult of
+                    true  ->
+                        % An object has crossed the threshold from being a single value
```
rzezeski: Single percent is meant for EOL comments.  Just saying cause this will confuse emacs erlang-mode.  Deal with my nitpicking comments :)

src/yz_solr.erl
```diff
@@ -139,7 +139,8 @@ index(Core, Docs) ->
     Opts = [{response_format, binary}],
     case ibrowse:send_req(URL, Headers, post, XML, Opts) of
         {ok, "200", _, _} -> ok;
-        Err -> throw({"Failed to index docs", Docs, Err})
+        Err -> throw({"Failed to index docs", Err})
+        % Err -> throw({"Failed to index docs", Docs, Err})
```
rzezeski: What's the reason for this change?

rzezeski:
Two dialyzer errors:

```
yz_kv.erl:56: Invalid type specification for function yz_kv:get_obj_ct/1. The success typing is (dict()) -> any()
yz_kv.erl:61: Invalid type specification for function yz_kv:is_tombstone/1. The success typing is (dict()) -> boolean()
```

src/yz_solr.erl
coderoshi: No reason, just cruft from an old test

src/yz_solr.erl
coderoshi, line 236: Yes it's needed. We don't want to delete every object that has a _yz_rk, only the siblings with the given key. I'll change the function sig to be clearer.

https://github.com/rzezeski/yokozuna/pull/4/files#L6R124

rzezeski:
+1 to merge

PR #24, opened 2012-12-05
Add jmx jvm properties

coderoshi:


coderoshi:
Related: https://github.com/rzezeski/riak/pull/1

src/yz_solr_proc.erl
```diff
@@ -110,11 +118,21 @@ build_cmd(SolrPort, Dir) ->
     PortArg = "-Djetty.port=" ++ SolrPort,
     LoggingProps = filename:join([Dir, "logging.properties"]),
     LoggingArg = "-Djava.util.logging.config.file=" ++ LoggingProps,
+    case JMXPort == undefined of
+        false ->
+            % JMXOptions = "-DOPTIONS=jmx",
```
rzezeski: remove this line

src/yz_solr_proc.erl
```diff
@@ -110,11 +118,21 @@ build_cmd(SolrPort, Dir) ->
...
+            JMXPortArg = "-Dcom.sun.management.jmxremote.port=" ++ JMXPort,
+            JMXAuthArg = "-Dcom.sun.management.jmxremote.authenticate=false",
+            JMXSSLArg = "-Dcom.sun.management.jmxremote.ssl=false",
+            JMXArgs = [JMXPortArg, JMXAuthArg, JMXSSLArg];
+        _ ->
+            JMXArgs = []
+    end,
```
rzezeski, line 129: Instead of matching against the value of `JMXPort == undefined` just directly match against `JMXPort`.

include/yokozuna.hrl
```diff
@@ -97,6 +97,7 @@
 -define(PARTITION_BINARY(S), S#state.partition_binary).
 
 -define(YZ_DEFAULT_SOLR_PORT, "8983").
+-define(YZ_DEFAULT_SOLR_JMX_PORT, "8985").
```
rzezeski: I don't see this being used anywhere.  Seems like we don't want a default if the default is to not have JMX enabled.

src/yz_solr.erl
```diff
@@ -175,6 +175,9 @@ ping(Core) ->
 port() ->
     app_helper:get_env(?YZ_APP_NAME, solr_port, ?YZ_DEFAULT_SOLR_PORT).
 
+jmx_port() ->
+    app_helper:get_env(?YZ_APP_NAME, solr_port_jmx, undefined).
```
rzezeski: Please rename the app var to `solr_jmx_port`.  That sounds better and is the name used in the rest of the code.

PR #31, opened 2013-01-02
Use JSON Update Instead of XML Update

coderoshi:
Use the Solr JSON update endpoint instead of XML.

My gut says this should reduce latency, increase throughput, and make the code slightly less complicated.  Make sure to benchmark to verify it's at least the same if not better.


src/yz_solr.erl
```diff
@@ -41,24 +41,19 @@
 
 -spec build_partition_delete_query(ordset(lp())) -> term().
 build_partition_delete_query(LPartitions) ->
-    Queries = [?QUERY(?YZ_PN_FIELD_S ++ ":" ++ ?INT_TO_STR(LP))
+    Deletes = [{delete, ?QUERY(?YZ_PN_FIELD_S ++ ":" ++ ?INT_TO_STR(LP))}
```
rzezeski: This won't work because mochijson2 treats a string as an array (since a string is just a list).  Try the following

``` erlang
?QUERY(<<?YZ_PN_FIELD_S,":",(?INT_TO_BIN(LP))/binary>>)>>.
```

rzezeski:
Looks like JSON update slightly nudged out XML.

### JSON

json-load

### XML

xml-load

rzezeski:
Please squash the commits into the first one since the first one is broken without the other two (this will help when using git-bisect).

coderoshi:
I can live with 15% improvement

rzezeski:
Forgot to add the +1 to merge on last comment.

PR #41, opened 2013-01-03
Update README to use yz-setup rather than manual steps

coderoshi:
PR basho/riak#264 adds a yz-setup build to the makefile that just encodes any manual steps needed to set up Yokozuna. This updates the README's build section to reflect that.


rzezeski:
+1 to merge

Issue #10, opened 2012-10-16
Schema Admin via HTTP [JIRA: RIAK-1715]


rzezeski:
Yokozuna should have the ability to create and modify schemas
remotely, HTTP for this specific issue.

There are still a lot of questions regarding this issue.  The
fundamental ones are:
- Can a running system cope with schema changes?  If so, how can it be
  done safely?
- Can schemas be modified piecemeal or must it be all-or-nothing?
- Can JSON be used to read/modify/write the schema?
- Should concurrent writers/siblings be accounted for?  I'm a little
  less worried about concurrent writers in a healthy cluster and more
  worried about partitioned writes.  Would PW/PR/DW/W/R=N be good
  enough?
## Specification
- The resource: `<host-port>/yokozuna/schema/<schema-name>`
### GET
- Return the schema with content type of `application/xml`.
- <del>TODO: allow to pick-out subset of schema to return, e.g. a list of
  fields?</del>
- <del>TODO: allow to return in JSON format?</del>
### PUT
- Accepts `text/xml` or `application/xml`.
- The body is a properly formed Solr schema.  See the
  [example schema](http://svn.apache.org/viewvc/lucene/dev/trunk/solr/example/solr/collection1/conf/schema.xml?view=markup).
- <del>If the schema name already exists then don't replace the current
  one.  Instead return an error to user stating it already exists.</del>  Need 
  to be able to overwrite a schema in case a bad schema is uploaded.
- TODO: Think about adding param `overwrite=true` to bypass the
  previous check allowing the user to overwrite the current schema
  definition.  This has to be thought about carefully because changing
  schemas could cause issues.
### POST

TODO: Think about allowing POSTs to add to or modify a subset of a
schema.  E.g. adding a new field without read/modify/write of entire
schema.
### DELETE

TODO: Do we allow deletes of schemas?

abhinavsingh:
@rzezeski My 2 cents over above concerns:
1. Yokozuna should explore the idea of adapting to schema changes i.e. running system will eventually cope up with schema changes all by itself
2. There can also be a flag using which developer can decide whether Yokozuna should adapt to schema changes for old data by itself. If this flag is turned on, Yokozuna can do relevant reindexing job in the background
[...]

PR #42, opened 2013-01-04
Timeouts during bulk data load, possible tie in to yz_events crash and a bad state on core create [JIRA: RIAK-1585]

neuhausler:


coderoshi:
This looks to be a partial implementation of #10.

src/yz_wm_schema.erl
```diff
@@ -0,0 +1,104 @@
...
+        undefined -> {{halt, 404}, RD, S};
+        _ -> {false, RD, S}
+    end.
+
+%% Responds to a PUT request by storing the schema
+%% Will overwrite schema with the same name
+store_schema(RD, S) ->
+    SchemaName = S#ctx.schema_name,
+    Schema = wrq:req_body(RD),
+    yz_schema:store(list_to_binary(SchemaName), Schema),
```
coderoshi, line 95: Overwriting a schema can be dangerous. I'd say at minimum by default PUT should refuse to overwrite an existing schema without a specific intent on the user's part. The schema should also be verified correct, and alert the user if invalid.

There's ample discussion of possible solutions here #10

neuhausler, line 95: Really good discussion! The only goal I had for that pull request was to make it at least possible to upload a schema via http, just to be able to start get experienced with all the issues related to schemas, and there are many, as discussed separately, and as experienced by myself. So it should just be seen as an enabler to start building prototypes.

Btw I tend to agree with one of Ryan's questions/suggestions, that at some point that "raw access" to the schemas has to get replaced by some more high-level API that behind the scene based on some JSON-config puts together a consistent and correct schema and! config file, as required by Yokozuna and Solr.

src/yz_wm_schema.erl
```diff
@@ -0,0 +1,104 @@
...
+%%
+
+-module(yz_wm_schema).
+-compile(export_all).
+-include("yokozuna.hrl").
+-include_lib("webmachine/include/webmachine.hrl").
+
+-record(ctx, {api_version :: atom(),   % Determine which version of the API to use.
+              schema_name :: string(), % name the schema
+              method :: atom()         % HTTP method for the request
```
rzezeski: The `api_version` and `method` fields aren't really used, please remove them.

src/yz_wm_schema.erl
```diff
@@ -0,0 +1,104 @@
...
+store_schema(RD, S) ->
+    SchemaName = S#ctx.schema_name,
+    Schema = wrq:req_body(RD),
+    yz_schema:store(list_to_binary(SchemaName), Schema),
+    {true, RD, S}.
+
+%% Responds to a GET request by returning schema
+read_schema(RD, S) ->
+    SchemaName = S#ctx.schema_name,
+    {yz_schema:get(list_to_binary(SchemaName)), RD, S}.
```
rzezeski: The `get` call returns `{ok, raw_schema()}` which will cause an error here.  It needs to be matched first.

src/yz_wm_schema.erl
rzezeski, line 95: In the name of pragmatism I'd like to see this change committed (with the fixes I've mentioned) and I'll add specific issues for the points above.

rzezeski:
Marcel, thanks for the contribution.  Please address the two comments I made, rebase/squash into the original commit, and I'll re-test the PR.

neuhausler:
will do, thanks.

rzezeski:
+1 to merge

Basho-JIRA:
We have a separate GH issue open in the backlog for bulk load performance. This seems specific to the issue fixed in RIAK-1567 (#437) and covered in https://basho.zendesk.com/agent/tickets/10352, which has seemed to work for the client. 

PR #45, opened 2013-01-25
Upgrade to Solr 4.1

rzezeski:
- http://wiki.apache.org/solr/ReleaseNote41
- https://issues.apache.org/jira/browse/SOLR/fixforversion/12321141#selectedTab=com.atlassian.jira.plugin.system.project%3Aversion-issues-panel
- http://lucene.apache.org/solr/4_1_0/changes/Changes.html

Run smartos benchmark to see if there is any query performance difference.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,7 +2,7 @@ deps
 ebin
 priv/*.jar
 priv/java_lib
-priv/solr
+priv/solr*
 priv/apache*
 priv/solr-jars
 priv/lucene-solr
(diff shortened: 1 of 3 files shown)


rzezeski:
My local, 5-node, SmartOS benchmarks show no regression in query perf.  Based on the release notes there have been improvement to the lucene codec that should help query performance in some cases.  I didn't see that in these benchmarks but I have a feeling it's because the distributed search apache http stalecheck is hiding them, see #26.

### Alpha Query -  Solr 4.0

solr40-alpha

### Alpha Query -  Solr 4.1

solr41-alpha

### Lima Query -  Solr 4.0

solr40-lima

### Lima Query - Solr 4.1

solr41-lima

PR #14, opened 2013-01-29
Files Overwritten on Core Failure

coderoshi:
If a Solr Core fails to initialize, or is not up for any reason, and a
ring event comes in then Yokozuna will try to create it because
`yz_index:exists` returns `false` if the Core is not running.  In most
cases this is harmless because the index data will not be lost.  But
in the case where modifications have been made to files living under
the `instanceDir` they will be lost.  This is because Yokozuna will
re-copy the configuration files overwriting any modifications made to
those local to the instance.
## Reproduce
- Run against Yokozuna commit `448fe02b1691b7943dcbdc87df4074d1b1fbde4a`
- Build a Riak rel, `make stage`, and start it
- Create a new index `yz_index:create("foo")`
- Stop Riak
- Edit `rel/riak/data/yz/foo/conf/config.xml` and remove the
  `startup="lazy"` from the `ExtractingRequestHandler`.
- Start Riak and tail -f `rel/riak/log/solr-0.log`
- You'll see a failure to load the class and then shortly after a
  successful creation of the core
- Reopen `rel/riak/data/yz/foo/conf/config.xml` and you'll see the
  `startup="lazy"` is there again.
## Potential Solutions
- Have `yz_index:exists` distinguish between a downed Solr Core and
  one that doesn't exist at all.  Take different actions depending on
  the situation.
- During file copy only overwrite files that don't already exist.
- Same as previous but only overwrite if the source file is newer.

### DIFF
--- a/src/yz_misc.erl
+++ b/src/yz_misc.erl
@@ -22,6 +22,7 @@
 -compile(export_all).
 -include("yokozuna.hrl").
 -include_lib("webmachine/include/webmachine.hrl").
+-include_lib("kernel/include/file.hrl").
 
 %%%===================================================================
 %%% API
@@ -42,10 +43,14 @@ convert_pl_entry({Partition,Node}, logical, Map) ->
     {LPartition,Node}.
 
 %% @doc Recursively copy each file to `Dir'.
--spec copy_files([string()], string()) -> ok.
-copy_files([], _) ->
+%% Defaults to `overwrite' existing files. You may also
+%% set to only `update' with more recent files,
+%% which first compares timestamps. Finally, you can
+%% `skip' any duplicate files.
+-spec copy_files([string()], string(), overwrite | update | skip) -> ok.
+copy_files([], _, _) ->
     ok;
-copy_files([File|Rest], Dir) ->
+copy_files([File|Rest], Dir, Action) ->
     Basename = filename:basename(File),
     case filelib:is_dir(File) of
         true ->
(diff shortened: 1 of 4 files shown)


src/yz_misc.erl
```diff
@@ -54,10 +59,47 @@ copy_files([File|Rest], Dir) ->
...
 
+%% @doc Recursively copy each file to `Dir',
+%% overwriting any exiting files
+-spec copy_files([string()], string()) -> ok.
+copy_files([], _) ->
+    ok;
+copy_files([File|Rest], Dir) ->
+    copy_files([File|Rest], Dir, overwrite).
+
+%% If destination exists, skip the copy
```
rzezeski, line 79: Given this function is a predicate I think it would be better named `should_copy`.

test/yz_misc_tests.erl
```diff
@@ -0,0 +1,60 @@
...
+%%--------------------------------------------------------------------
+
+%% @doc EUnit and Exercise yz_misc.
+-module(yz_misc_tests).
+
+-compile(export_all).
+
+-include_lib("eunit/include/eunit.hrl").
+
+prop_does_copy_action_skip_test() ->
```
rzezeski: Might want to consider removing the `prop_` prefix.  In all the Riak code I've seen we use `prop_` to indicate an EQC test.

test/yz_misc_tests.erl
```diff
@@ -0,0 +1,60 @@
...
+    Thisfile = atom_to_list(?MODULE) ++ ".erl",
+    false = yz_misc:does_copy_action(skip, Thisfile, Thisfile),
```
rzezeski: At the end of the day this probably doesn't matter so much, but if you use the eunit macros such as ?assert and ?assertNot the error message is a bit more descriptive.  Feel free to stick with matching, but thought I'd point this out in case you haven't seen it.

### Matching

```
in function yz_misc_tests:prop_does_copy_action_overwrite_test/0 (test/yz_misc_tests.erl, line 58)
**error:{badmatch,true}
```

### ?assertNot

```
    yz_misc_tests: prop_does_copy_action_overwrite_test...*failed*
in function yz_misc_tests:'-prop_does_copy_action_overwrite_test/0-fun-0-'/0 (test/yz_misc_tests.erl, line 58)
in call from yz_misc_tests:prop_does_copy_action_overwrite_test/0 (test/yz_misc_tests.erl, line 58)
**error:{assertion_failed,[{module,yz_misc_tests},
                   {line,58},
                   {expression,"not ( yz_misc : does_copy_action ( overwrite , \"<nofile>\" , \"<nofile>\" ) )"},
                   {expected,true},
                   {value,false}]}
```

rzezeski:
- Compiles + eunit passes
- No new dialyzer errors
- Verified it does not overwrite core

test/yz_misc_tests.erl
coderoshi: will do. case of repurposing, this started life as an eqc test

rzezeski:
Looks like you accidentally added the `log` dir.

rzezeski:
+1 to merge

PR #57, opened 2013-01-29
Handle {error, Reason} in make_doc/index

coderoshi:
Refer to #112 as a replacement for this issue.

The `yz_doc:extract_fields` function can return `{error, any()}` yet `make_doc` assumes success.  This would blow up on the call to `lists:append` which would bubble to `yz_kv:index` which at that point is outside the try/catch and would crash vnode.  This is a sign of some sloppiness on my part in thinking about error handling during indexing.

The entire code path from the KV vnode call of `yz_kv:index` to its return should be gone over to check for common error conditions that might arise do to user input and decide how to deal with them.

### DIFF
--- a/src/yz_kv.erl
+++ b/src/yz_kv.erl
@@ -150,13 +150,14 @@ index(Obj, Reason, VNodeState) ->
     LFPN = yz_cover:logical_partition(LI, first_partition(PrimaryPL)),
     P = get_partition(VNodeState),
     LP = yz_cover:logical_partition(LI, P),
-    Docs = yz_doc:make_docs(Obj, ?INT_TO_BIN(LFPN), ?INT_TO_BIN(LP),
+    AllDocs = yz_doc:make_docs(Obj, ?INT_TO_BIN(LFPN), ?INT_TO_BIN(LP),
                             index_content(BProps)),
+    ValidDocs = [D || D <- AllDocs, element(1, D) == doc],
     IdxN = {first_partition(PrimaryPL), NVal},
 
     try
-        ok = yz_solr:index(Index, Docs),
-        ok = cleanup(length(Docs), {Index, Obj, Key, LP}),
+        ok = yz_solr:index(Index, ValidDocs),
+        ok = cleanup(length(ValidDocs), {Index, Obj, Key, LP}),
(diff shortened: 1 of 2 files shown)


src/yz_kv.erl
```diff
@@ -150,13 +150,14 @@ index(Obj, Reason, VNodeState) ->
     LFPN = yz_cover:logical_partition(LI, first_partition(PrimaryPL)),
     P = get_partition(VNodeState),
     LP = yz_cover:logical_partition(LI, P),
-    Docs = yz_doc:make_docs(Obj, ?INT_TO_BIN(LFPN), ?INT_TO_BIN(LP),
+    AllDocs = yz_doc:make_docs(Obj, ?INT_TO_BIN(LFPN), ?INT_TO_BIN(LP),
                             index_content(BProps)),
+    ValidDocs = [D || D <- AllDocs, element(1, D) == doc],
```
coderoshi, line 155: Having valid docs is great. What to do with invalid docs?

Should we quash the entire update, or just do a partial index? If partial, what's a good way of communicating that. Letting this sit for now.

rzezeski:
I started writing a riak_test with the goal to test various failure scenarios.  The test can be found on the [rz-errors-test](https://github.com/basho/yokozuna/compare/master...rz-errors-test) branch.

My first test was to write malformed JSON by simply truncating the last few characters.  Ideally you'd expect this to return with a `400` but it doesn't.  The problem is that the KV webmachine resource, KV PUT FSM, and KV VNode drive the request.  How does Yokozuna signal that the data is bad to the client?  Remember, KV treats the data as opaque and doesn't care what it is.  Yokozuna, on the other hand, has to look at the data so that it can index it.

This has big ramifications.  The KV and YZ write are not atomic in relation to one another.  What should we do in failure cases?
- Return failure to user even if KV wrote fine?
- Return success even though indexing failed?
- Roll back KV write + return error to client?
- Fall back to indexing the content as plain text?
- Log that indexing failed and hope user notices? (this is current solution except the `try/catch` net needs to be cast wider)

This topic also becomes very important if we look into different partitioning schemes for YZ.  In the future indexing may not be local to the KV write.  Whatever approach we take should keep that in mind.

coderoshi:
Closing this PR. Replaced with #112. Sadly, this will also close the issue, because github.

PR #61, opened 2013-02-03
Port Latest AAE Changes

rzezeski:
More changes were made during Riak 1.3 testing which must be ported to Yokozuna.  These are blockers on 0.3.0 release since yokozuna_essential is having AAE related failures.  Not all of these changes necessarily apply to Yokozuna AAE.
- <del>basho/riak_kv#460 (AAE Exchange Bug for N=1)</del>
- <del>basho/riak_kv#472 (Various AAE Changes Prompted by QA)</del>
- <del>basho/riak_kv#473 (Don't Repair Over and Over)</del>
- <del>basho/riak_kv#474 (Fix AAE Rehash When Data is Missing)</del>

### DIFF
--- a/src/yz_index_hashtree.erl
+++ b/src/yz_index_hashtree.erl
@@ -180,9 +180,8 @@ handle_call({compare, Id, Remote, AccFun}, From, S) ->
     {noreply, S};
 
 handle_call(destroy, _From, S) ->
-    {_,Tree0} = hd(S#state.trees),
-    hashtree:destroy(Tree0),
-    {stop, normal, ok, S};
+    S2 = destroy_trees(S),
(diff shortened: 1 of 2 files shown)


rzezeski:
basho/riak_kv#460 doesn't apply because Yokozuna AAE exchanges a given partition between KV and Yokozuna, it doesn't compare adjacent partitions of Yokozuna.  Thus, N=1 makes no difference.

rzezeski:
basho/riak_kv#473 does not apply.  Yokozuna already handles this by re-indexing regardless if read-repair occurred or not.

rzezeski:
basho/riak_kv#474 does not apply as it fixed something in 473 which also didn't apply.

PR #69, opened 2013-03-11
Update port numbers in the README

timdoug:
riak/rel/gen_dev generates configurations where the Riak and Solr nodes
listen on different ports than those currently in the documentation.
Considering step 4 in the building section suggests make stagedevrel,
update the subsequent port numbers used in the examples accordingly.


rzezeski:
Thanks for the patch.

PR #76, opened 2013-03-15
Add filter query compression

rzezeski:
After running some benchmarks I discovered that query throughput can
be increased by reducing the number of term queries in the 'fq' param
passed to Solr.

Currently Yokozun passes an 'fq' like so:

(nodeA AND p1) OR (nodeB AND p4) OR (nodeC AND p7) OR (nodeA AND p10) ...

Notice that the nodes repeat for each node-partition combo.  This
"compresses" the 'fq' to the following:

(nodeA AND (p1 OR p10 OR ...)) OR (nodeB AND (p4 OR ...)) OR ...

Notice that the partitions are now grouped with the corresponding
owning node.  On my local workstation I noticed a ~17% improvement in
throughput.  As a lot of the filter query work is done in memory it
should be CPU bound so the slower the CPU the more this change should
benefit.

### DIFF
--- a/include/yokozuna.hrl
+++ b/include/yokozuna.hrl
@@ -47,7 +47,8 @@
 
 %% Shorthand for existing types
 -type datetime() :: calendar:datetime().
--type orddict(K,V) :: [{K,V}].
+-type orddict(K,V) :: orddict:orddict(K,V).
+-type od() :: orddict:orddict().
 -type ordset(T) :: ordsets:ordset(T).
 -type proplist() :: proplists:proplist().
 -type ring() :: riak_core_ring:riak_core_ring().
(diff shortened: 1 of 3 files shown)


rzezeski:
The following benchmark was run on my local workstation.
- core i7-3970X, 32GB DDR3 1333, Intel 520 6Gb/s SSD
- Ubuntu 12.10
- Erlang R15B02
- Java HotSpot(TM) 64-Bit Server VM (build 23.7-b01, mixed mode)

The lima query is a single-term query that matches 1 document out of 1 million.

```
+---------------+---------------------+--------------------+------------------+
|               |Throughput query/s   |Median Latency ms   |95th Latency ms   |
+---------------+---------------------+--------------------+------------------+
|no compress    |1111                 |28.6                |37.2              |
+---------------+---------------------+--------------------+------------------+
|compressed     |1307                 |24.3                |32.7              |
+---------------+---------------------+--------------------+------------------+
```

## No Compress

no-fq-compress-query-lima-2

## Compressed

fq-compress-query-lima-2

Issue #74, opened 2013-03-14
/bin/bashism found in priv/ shell scripts


jj1bdx:
While building on FreeBSD 9.1-STABLE, the following scripts were unable to execute: `priv/build-jar.sh` and `priv/grab-solr.sh`.

I was hitting this issue when I was working on Riak - maybe @jaredmorrow knows this issue better than me.

The quick hack for FreeBSD is changing `#!/bin/bash` to `#!/usr/local/bin/bash`, but I don't recommend this because it includes an absolute path. Maybe `#!/usr/bin/env bash`? (Correct me if I'm wrong)

jaredmorrow:
@jj1bdx can you give the latest master a try, should work just fine now on FreeBSD.

jj1bdx:
https://github.com/basho/yokozuna/issues/77 worked. Thanks. This issue is closed.

Another dep issue found. See https://github.com/basho/riak/pull/293

PR #77, opened 2013-03-15
Use `env` to find bash so tests & build can run on *BSD

jaredmorrow:
Simply converted fixed paths on the shebang line from
'/bin/bash' to '/usr/bin/env bash'.  `yokozuna-ami.sh`
also needed a +x flag that was not already set.

Some of the bash scripts can easily be converted to pure
/bin/sh implementations, but it is probably not needed
due to them being build or test related rather than
used in runtime or operations.

Addresses #74

### DIFF
--- a/priv/build-jar.sh
+++ b/priv/build-jar.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 #
 # Temporary script to build JAR file containing customer Solr request
 # handlers.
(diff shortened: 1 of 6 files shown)


rzezeski:
I don't see any change to `yokozuna-ami.sh` related to `+x`.  That's for turning shell debugging off, right?

jaredmorrow:
Oh sorry, unclear commit message, I just `chmod +x`'d that file.  It was the only shell script not executable.

rzezeski:
+1 to merge

PR #50, opened 2013-03-27
Add Riak Test for Various Languages

coderoshi:
Add tests for other languages, maybe 3 or 4.  Include Japanese since there has been some interest there.

Mainly want to verify that:
- Character encoding is not mangled going from client to yokozuna to solr (and back).
- Languages are analyzed properly.

### DIFF
--- a/riak_test/yz_languages.erl
+++ b/riak_test/yz_languages.erl
@@ -1,0 +1,188 @@
+%% @doc Test the index adminstration API in various ways.
+-module(yz_languages).
+-compile(export_all).
+-include_lib("eunit/include/eunit.hrl").
+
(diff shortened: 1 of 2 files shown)


rzezeski:
So this will certainly test that the encoding is not getting mangled from client->solr but it isn't actually using the different language analyzers.  It is instead treating all languages as `text_general`.  I'd like to see fields added for the various languages and then use those new fields for this test.

https://github.com/basho/yokozuna/blob/master/priv/default_schema.xml#L615

https://github.com/basho/yokozuna/blob/master/priv/default_schema.xml#L60

coderoshi:
Added schema to ensure language type

rzezeski:
+1 to merge

PR #71, opened 2013-03-28
Add JVM args setting via app.config

coderoshi:
We can have reasonable defaults, however, different versions/implementations of the JVM can be optimized with different settings and args. For example:

``` erlang
{yokozuna, [
            {solr_vm_args, ["-server", "-XX:+UseStringCache"]},
]}
```

would append the string to the end of the `java` command which launches solr.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,4 @@ tests/20*
 tests/current
 .eunit
 log
+bb-*-fruit*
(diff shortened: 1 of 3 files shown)


coderoshi:
Also added app.config settings to  basho/riak@fec83d4

coderoshi:
Merging to avoid conflict with #49 work

PR #49, opened 2013-03-29
AAE Config

coderoshi:
Currently Yokozuna takes on the same config values as KV for much of AAE.  Instead, first check for a Yokozuna specific config.  If that doesn't exist then fallback to KV config.
- [x] exchange tick
- [x] expire
- [x] concurrency

### DIFF
--- a/include/yokozuna.hrl
+++ b/include/yokozuna.hrl
@@ -136,7 +136,32 @@
 -define(YZ_AE_DIR,
         application:get_env(?YZ_APP_NAME, anti_entropy_data_dir)).
 -define(YZ_ENTROPY_TICK,
-        app_helper:get_env(?YZ_APP_NAME, entropy_tick, 60000)).
+        app_helper:get_env(?YZ_APP_NAME, entropy_tick,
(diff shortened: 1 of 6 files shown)


coderoshi:
Other configs
- [x] anti_entropy_build_limit
- [x] anti_entropy (is active)
- [x] anti_entropy_timeout

include/yokozuna.hrl
```diff
@@ -134,9 +134,36 @@
...
+        app_helper:get_env(?YZ_APP_NAME, anti_entropy_expire,
+            app_helper:get_env(riak_kv, anti_entropy_expire, 604800000)
+        )).
+%% Per state transition timeout used by certain transitions
+%% Defaults to 5 minutes
+-define(YZ_ENTROPY_TIMEOUT,
+        app_helper:get_env(?YZ_APP_NAME, anti_entropy_timeout,
+            app_helper:get_env(riak_kv, anti_entropy_timeout, 300000)
+        )).
+%% Per state transition timeout used by certain transitions
```
rzezeski: Copy pasta

PR #82, opened 2013-03-29
Cleanup old rzezeski references in docs

kyleslattery:
There were a couple links to the rzezeski/yokozuna repository in the EC2 documentation. They now point to basho/yokozuna


rzezeski:
Thanks for the patch.

PR #84, opened 2013-04-02
Fix up noop spec

coderoshi:


src/yokozuna.erl
```diff
@@ -50,7 +50,7 @@ solr_port(Node, Ports) ->
 
 %% @doc get an associated flag, true if some action
 %%      (eg indexing, searching) should be supressed
--spec noop_flag(atom()) -> binary() | undefined.
+-spec noop_flag(index|search) -> boolean() | undefined.
```
rzezeski: This cannot return `undefined`.

rzezeski:
+1 to merge

PR #86, opened 2013-04-02
Delete query failed due to unformatted key

coderoshi:
It turns out that `{"delete":{"query":"_yz_rk:/some/valid/key"}}` needs formatted


rzezeski:
It would be nice to have an addition to one of the riak tests that verifies the fix and protects against regressions.

rzezeski:
+1 to merge

PR #56, opened 2013-04-12
Strip Down Default config.xml

coderoshi:
Go through **EVERY LINE** in [config.xml](https://github.com/rzezeski/yokozuna/blob/master/priv/conf/config.xml), understand what it does, and remove it if it doesn't make sense as the default for Yokozuna.

Also, rename this file to `solrconfig.xml` as that's what it is called in the Solr example and it will cause less confusion for those users who already have experience with Solr.


rzezeski:
Was reading the the solr mailing list and came across a thread about the default POST size limit of 2MB.  It can be configure in the config file via `formdataUploadLimitInKB`.  We should increase this to something ridiculous like 100MB to avoid the scenario where KV accepts the write but Solr fails.  This would cause endless failed repairs from AAE.

http://lucene.472066.n3.nabble.com/what-is-the-maximum-XML-file-size-to-import-td4058263.html

Issue #92, opened 2013-04-09
Default JVM args to empty list


rzezeski:
Change the `?YZ_DEFAULT_SOLR_VM_ARGS` to default to `[]`.  We shouldn't force a default set of JVM options if the user removes the `solr_vm_args` from `app.config`.

PR #97, opened 2013-04-25
Don't pass any JVM args as default

rzezeski:
If a user doesn't specify solr jvm args in app.config then the default
should be to pass no args.  I think this is the least surprising
behavior to have.  We can still ship pre-canned defaults in the
app.config such as larger heap but the fallback defaults should be
none.


rzezeski:
Addresses #92.

PR #59, opened 2013-04-26
yz_wm_extract making bad call to yz_extractor:is_registered

hazen:
The extractor name passed to `yz_wm_extract:check_if_registered_set_ct` is a string but `yz_extractor:is_registered` wants an atom.

### DIFF
--- a/riak_test/yz_wm_extract_test.erl
+++ b/riak_test/yz_wm_extract_test.erl
@@ -1,0 +1,50 @@
+%% @doc Test the index schema API in various ways.
+-module(yz_wm_extract_test).
+-compile(export_all).
+-import(yz_rt, [host_entries/1, select_random/1]).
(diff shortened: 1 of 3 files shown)


riak_test/yz_wm_extract_test.erl
```diff
@@ -0,0 +1,53 @@
...
+    {ok, Status, _, _} = http(put, URL, Headers, "{\"name\":\"ryan\"}"),
+    ?assertEqual("200", Status).
+
+%%%===================================================================
+%%% Helpers
+%%%===================================================================
+
+http(Method, URL, Headers, Body) ->
+    Opts = [{response_format, binary}],
+    ibrowse:trace_on(),
```
rzezeski: Remove the tracing calls.

rzezeski:
+1 to merge

PR #95, opened 2013-04-27
Create a Schema which indexes every value by default

rzezeski:
I've run into a few people who want all of the default yz index behavior, with a fallback where any field that doesn't match a dynamicField name is just indexed as a string. This could be done with the addition of a new schema that appends `<dynamicField name="*" type="string"...>`, and named perhaps `_yz_universal`.

### DIFF
--- a/priv/default_schema.xml
+++ b/priv/default_schema.xml
@@ -105,6 +105,9 @@
    <dynamicField name="*_ig" type="ignored" multiValued="true"/>
    <dynamicField name="random_*" type="random" />
 
+   <!-- catch-all field -->
+   <dynamicField name="*" type="text_general" indexed="true" stored="false" multiValued="true" />
+
    <field name="_yz_id" type="_yz_str" indexed="true" stored="true" required="true"/>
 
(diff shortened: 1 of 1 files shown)


rzezeski:
More thoughts on the catch-all field.
- Should be indexed so it can be searched on.
- Shouldn't be stored as value could be large.
- Should be multi-valued because the field could repeat.  This means sorting on these fields will not be possible.
- Should be treated as unstructured text and thus use an analyzed field type such as `text_general`.  Using `string` would store it as-is and thus reduce effectiveness of searching.
- Should potentially `omitNorms` and `omitTermFreqAndPositions` since the values are of unknown size and structure.  This would save disk space + memory BUT would reduce accuracy of scoring when querying those field.  The argument could be made that if the user wants good scoring they should explicitly define a field.  It would also prevent things like phrase queries from working since they require term positions.

The upshot of all the points above is that the default schema will be able to handle unknown fields but only basic searches will work against those fields.

I think this is a good start.

``` XML
<dynamicField name="*" type="text_general" indexed="true" stored="false" multiValued="true" />
```

PR #101, opened 2013-04-30
Issue #66: Use symbols instead of constants for put quorum values

### LINKED ISSUE #66: KV Put Hardcodes Quorum Values
https://github.com/rzezeski/yokozuna/blob/master/src/yz_kv.erl#L216

The quorum values are hardcoded.  It should use symbolic values instead.

hazen:
This closes #66 because that issue seems to be lost (to me) on Heya

### DIFF
--- a/riak_test/yokozuna_essential.erl
+++ b/riak_test/yokozuna_essential.erl
@@ -51,7 +51,7 @@
         ]).
 
 confirm() ->
-    YZBenchDir = rt:get_os_env("YZ_BENCH_DIR"),
+    YZBenchDir = rt_config:get_os_env("YZ_BENCH_DIR"),
     code:add_path(filename:join([YZBenchDir, "ebin"])),
     random:seed(now()),
     Nodes = rt:deploy_nodes(4, ?CFG),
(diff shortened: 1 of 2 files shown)


include/yokozuna.hrl
```diff
@@ -124,6 +124,10 @@
 -define(YZ_SVC_NAME, yokozuna).
 -define(YZ_VNODE_MASTER, yokozuna_vnode_master).
 -define(YZ_META_INDEXES, yokozuna_indexes).
+-define(YZ_KV_PUT_PRIMARY_WRITE, 3).
+-define(YZ_KV_PUT_WRITE, 3).
+-define(YZ_KV_PUT_DURABLE_WRITE, 3).
```
rzezeski: These are still hardcoded values.  There are atom values that can be used to tell riak that you want them to be whatever N val is.  I think it's `n_val` or `n` but can't remember off the top of my head.

hazen: Cool.  I was trying to interpret the issue.  If there are some other values global values, I'll see if I can root them out

rzezeski: Yea, so if you locally modify the essential test CFG to use a `n_val` of 2 it will fail because the quorum parameters will be greater than `n_val`.

rzezeski:
Yea,that should work, but one thing.  The commit msg isn't quite right, you're actually grabbing the n-val now, not using the symbolic values.  Both methods work but that commit msg is confusing since it's not what the code does.

hazen:
Can I change the commit message with yet another amend/force commit/push?

rzezeski:
+1 to merge

Issue #37, opened 2013-01-02
Extend a client to work for Yokozuna


coderoshi:
As we push closer to a production release, we should have a Riak client or two which can manage/query YZ. I lean toward Ruby via HTTP, but any will do.

rzezeski:
Any client that currently supports Riak Search over HTTP can start querying Yokozuna once it starts listening on the canonical Solr resource.  I.e. `/solr/<index>/select`.  I'll add an issue for that.

That is an easy start but Solr has many more features than Riak Search.  How do we expose them going forward in our _Riak_ clients?  Do we want to be very specific about what can be passed (e.g. which query params can be passed)?  That could be very tedious.  Or do we want to be general and act as a mediator between the user and Solr?  Allowing less client code but chance for invalid calls.

It's worth thinking about how we can integrate existing Solr clients as that could save us a lot of time, both now and in the future (it could also cause headaches as someone else's bugs are now ours too).  E.g. we could integrate the SolrJ client with our java client.

You are right, we need to start working on client support now, not later.

coderoshi:
After speaking with @seancribbs I think we can hold off on this until after YZ is released. #38 allows for querying YZ at parity with Search, which brings us to the goal of being a suitable replacement. Search doesn't support any indexing/administration via PB, so we'd have to create new messages anyway.

This is certainly a worthwhile goal in the future, but marking it as "must" might be a bit heavy-handed right now.

rzezeski:
This issue was a bit vague to being with.  I say we close it with #38.  But making sure the main clients work with Yokozuna is a "must" for getting it into mainline.  Both HTTP and PB should work.  If those existing clients only support the limited RS Solr-like interface that is fine.  The key is that users can upgrade and their queries still work.

I believe the following clients are the main ones to support.
- Java
- Ruby
- Erlang
- Python

It should be easy enough to run their respective unit/integration tests and make sure they work.  However, I believe some of them use the RS Solr-like update resource.  Yokozuna doesn't support that ATM, it only indexes KV data.  So those tests would break.  Other than that things should hopefully "just work", but we still need to verify.

I'll go ahead and create an issue for each client to verify that it has parity with RS on the query side of things.

Issue #20, opened 2012-11-02
Riak Search Migration [JIRA: RIAK-1713]


rzezeski:
Provide methods and documentation for migrating from Riak Search.
## Checklist
- [x] Add function to flip query service from RS to Yokozuna (basho/riak_api#31)
- [x] Add riak-admin cmd to flip query service?
- [x] Add YZ AAE status reporting (#154)
- [x] Add YZ AAE status to `raik-admin aae-status`
- [ ] Add command for immediate re-index
- [x] Test that AAE migration works
- [ ] Test that command migration works (i.e. some command to immediately re-index under Yokozuna)
- [x] Document migration steps
- [x] Document migration of schema
- [ ] Create example Solr schema that represents approximation of default RS schema
## Migration by AAE

A nice benefit of active-anti-entropy (AAE) is that it provides a
"free" upgrade path from Riak Search to Yokozuna.  After upgrading a
node queries would still go to Riak Search but new data would
temporarily be indexed by both Riak Search and Yokozuna.  Over time
AAE would repair the missing data from Yokozuna, completing it's
index.  After AAE has repaired everything the Riak Search hook could
be removed and queries could be redirected to Yokozuna.  This simple
description hand-waves over a few things.
- How to determine when AAE has repaired all the data?
- What does it mean to "redirect the query"?  Would it require the
  client switching the HTTP resource it points to?  Would it be a
  dynamic re-binding in Riak so that the Riak Search Solr interface
  now points to Yokozuna?
- Riak Search tries it's best to look and act like Solr but at the end
  of the day it isn't Solr.  The results, in terms of ranking and
  format, will probably change.  How will this affect clients?  Will
  the Riak Search client work against Yokozuna?  Should there be a
  translation?  Should we just force the client to use a real Solr
  client?  If so, how do you properly integrate existing Solr clients
  with Riak clients?  E.g. a client runs a Solr query for the keys but
  then runs a Riak query for the object data.  Are there easy ways to make
  this one operation?

rzezeski:
## Migration By Fold
- Migration by AAE has overhead of Riak Read per object. (Should we be using read-repair for AAE?)
- We could use vnode fold directly for more efficient migration.
1. get list of indexed buckets
2. fold over objects in each bucket
3. send function to fold which re-indexes those objects
- Would need some function or command to invoke this migration method.

rzezeski:
Other things to think about:
[...]

PR #38, opened 2013-04-30
Add Protobuf support to Yokozuna

coderoshi:
Protocol buffers are a lower bandwidth alternative to the HTTP interface, and would be nice to implement any future clients (#37) against.

Currently this issue's aim is to match the functionality of RiakSearch. RS only supports a limited number of query params via PB, so this will mainly be of interest for existing RS users wishing to transition to Yokozuna. Also, RS only supports querying, and does not have PB support for creating/managing indexes.

### DIFF
--- a/riak_test/yz_languages.erl
+++ b/riak_test/yz_languages.erl
@@ -38,15 +38,6 @@ wait_for_joins(Cluster) ->
     rt:wait_until_nodes_ready(Cluster),
     rt:wait_until_no_pending_changes(Cluster).
 
-%% @doc Confirm a custom schema may be added.
-confirm_create_schema(Cluster, Name, RawSchema) ->
-    HP = select_random(host_entries(rt:connection_info(Cluster))),
-    lager:info("confirm_create_schema ~s [~p]", [Name, HP]),
-    URL = schema_url(HP, Name),
-    Headers = [{"content-type", "application/xml"}],
-    {ok, Status, _, _} = http(put, URL, Headers, RawSchema),
-    ?assertEqual("204", Status).
-
 select_random(List) ->
     Length = length(List),
     Idx = random:uniform(Length),
--- a/riak_test/yz_pb.erl
+++ b/riak_test/yz_pb.erl
@@ -82,8 +82,11 @@ store_and_search(Cluster, Bucket, Key, Body, Search, Params) ->
     lager:info("search_results"),
     {ok,{search_results,[{Bucket,Results}],Score,Found}} =
             riakc_pb_socket:search(Pid, Bucket, Search, Params),
-    Bucket == binary_to_list(proplists:get_value(<<"_yz_rk">>, Results)),
-    Found == 1,
+    ?assertEqual(
+        Key,
+        binary_to_list(proplists:get_value(<<"_yz_rk">>, Results))),
+    ?assertEqual(Found, 1),
+    ?assertNot(Score == 0.0),
     ok.
 
 confirm_basic_search(Cluster) ->
--- a/riak_test/yz_pb.erl
+++ b/riak_test/yz_pb.erl
@@ -14,6 +14,7 @@ confirm() ->
     Cluster = prepare_cluster(4),
     confirm_basic_search(Cluster),
     confirm_encoded_search(Cluster),
+    confirm_multivalued_field(Cluster),
     pass.
 
 
@@ -60,7 +61,8 @@ http(Method, URL, Headers, Body) ->
     Opts = [],
     ibrowse:send_req(URL, Headers, Method, Body, Opts).
 
-create_index(HP, Index) ->
+create_index(Cluster, Index) ->
+    HP = select_random(host_entries(rt:connection_info(Cluster))),
     lager:info("create_index ~s [~p]", [Index, HP]),
     URL = index_url(HP, Index),
     Headers = [{"content-type", "application/json"}],
@@ -69,13 +71,15 @@ create_index(HP, Index) ->
     ?assertEqual("204", Status).
 
 store_and_search(Cluster, Bucket, Key, Body, Search, Params) ->
+    store_and_search(Cluster, Bucket, Key, Body, "text/plain", Search, Params).
+
+store_and_search(Cluster, Bucket, Key, Body, CT, Search, Params) ->
     HP = select_random(host_entries(rt:connection_info(Cluster))),
-    create_index(HP, Bucket),
     URL = bucket_url(HP, Bucket, Key),
     lager:info("Storing to bucket ~s", [URL]),
     {Host, Port} = HP,
     %% populate a value
-    {ok, "204", _, _} = ibrowse:send_req(URL, [{"Content-Type", "text/plain"}], put, Body),
+    {ok, "204", _, _} = ibrowse:send_req(URL, [{"Content-Type", CT}], put, Body),
     %% Sleep for soft commit
     timer:sleep(1100),
     {ok, Pid} = riakc_pb_socket:start_link(Host, (Port-1)),
@@ -91,6 +95,7 @@ store_and_search(Cluster, Bucket, Key, Body, Search, Params) ->
 
 confirm_basic_search(Cluster) ->
     Bucket = "basic",
+    create_index(Cluster, Bucket),
     lager:info("confirm_basic_search ~s", [Bucket]),
     Body = "herp derp",
     Params = [{sort, <<"score desc">>}, {fl, ["*","score"]}],
@@ -98,7 +103,28 @@ confirm_basic_search(Cluster) ->
 
 confirm_encoded_search(Cluster) ->
     Bucket = "encoded",
+    create_index(Cluster, Bucket),
     lager:info("confirm_encoded_search ~s", [Bucket]),
     Body = "א בְּרֵאשִׁית, בָּרָא אֱלֹהִים, אֵת הַשָּׁמַיִם, וְאֵת הָאָרֶץ",
     Params = [{sort, <<"score desc">>}, {fl, ["_yz_rk"]}],
     store_and_search(Cluster, Bucket, "וְאֵת", Body, <<"text:בָּרָא">>, Params).
+
+confirm_multivalued_field(Cluster) ->
+    Bucket = "basic",
+    lager:info("cofirm multiValued=true fields decode properly"),
+    Body = <<"{\"name_ss\":\"turner\", \"name_ss\":\"hooch\"}">>,
+    Params = [],
+    HP = select_random(host_entries(rt:connection_info(Cluster))),
+    URL = bucket_url(HP, Bucket, "multivalued"),
+    lager:info("Storing to bucket ~s", [URL]),
+    {Host, Port} = HP,
+    %% populate a value
+    {ok, "204", _, _} = ibrowse:send_req(URL, [{"Content-Type", "application/json"}], put, Body),
+    %% Sleep for soft commit
+    timer:sleep(1100),
+    Search = <<"name_ss:turner">>,
+    {ok, Pid} = riakc_pb_socket:start_link(Host, (Port-1)),
+    {ok,{search_results,[{Bucket,Fields}],Score,Found}} =
+            riakc_pb_socket:search(Pid, Bucket, Search, Params),
+    ?assert(lists:member({<<"name_ss">>,<<"turner">>}, Fields)),
+    ?assert(lists:member({<<"name_ss">>,<<"hooch">>}, Fields)).
--- a/riak_test/yz_pb.erl
+++ b/riak_test/yz_pb.erl
@@ -1,0 +1,114 @@
+%% Running this test requires the `riak-erlang-client' project
+%% in the yokozuna/deps directory.
+
+%% @doc Test the index adminstration API in various ways.
+-module(yz_pb).
+-compile(export_all).
+-include_lib("eunit/include/eunit.hrl").
+
+-define(FMT(S, Args), lists:flatten(io_lib:format(S, Args))).
+-define(NO_HEADERS, []).
+-define(NO_BODY, <<>>).
+
+confirm() ->
+    case riak_client_exists() of
+        true ->
+            YZBenchDir = rt:get_os_env("YZ_BENCH_DIR"),
+            code:add_path(filename:join([YZBenchDir, "ebin"])),
+            random:seed(now()),
+            Cluster = prepare_cluster(4),
+            confirm_basic_search(Cluster),
+            confirm_encoded_search(Cluster),
+            pass;
+        _ ->
+            pass
+    end.
+
+
+prepare_cluster(NumNodes) ->
+    %% Note: may need to use below call b/c of diff between
+    %% deploy_nodes/1 & /2
+    %%
+    % Nodes = rt:deploy_nodes(NumNodes, ?CFG),
+    Nodes = rt:deploy_nodes(NumNodes),
+    Cluster = join(Nodes),
+    wait_for_joins(Cluster),
+    rt:wait_for_cluster_service(Cluster, yokozuna),
+    Cluster.
+
+join(Nodes) ->
+    [NodeA|Others] = Nodes,
+    [rt:join(Node, NodeA) || Node <- Others],
+    Nodes.
+
+%% TODO: replace with rt:wait_for_joins
+wait_for_joins(Cluster) ->
+    lager:info("Waiting for ownership handoff to finish"),
+    rt:wait_until_nodes_ready(Cluster),
+    rt:wait_until_no_pending_changes(Cluster).
+
+select_random(List) ->
+    Length = length(List),
+    Idx = random:uniform(Length),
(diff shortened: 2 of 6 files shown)


riak_test/yz_languages.erl
```diff
@@ -38,15 +38,6 @@ wait_for_joins(Cluster) ->
     rt:wait_until_nodes_ready(Cluster),
     rt:wait_until_no_pending_changes(Cluster).
 
-%% @doc Confirm a custom schema may be added.
```
rzezeski, line 41: I assume this test was removed since `yz_schema_admin` tests schema stuff now.  But it would have been nice to have this change in a dedicated commit for a cleaner history. Given the commit msg I have no clue why this was pulled.  And it really has nothing to do with the rest of the commit.   No need to change anything in this PR, just wanted to mention it.

src/yz_pb_search.erl
```diff
@@ -0,0 +1,127 @@
...
+    {error, missing_query};
+extract_params(#rpbsearchqueryreq{q=Query,
+                                rows=Rows, start=Start,
+                                filter=Filter,
+                                df=DefaultField, op=_DefaultOp}) ->
+    {ok, [{q, Query},
+          {wt, "json"},
+          {omitHeader, true},
+          {fq, default(Filter, "")},
+          {fl, <<"*,score">>},
```
rzezeski: I need to try testing this locally to verify (and we should add a test to riak-test), but we should be passing in the user supplied `fl`.  If no `fl` is supplied then we use `*,score`.  Would be good to run a few queries manually against riak search and yoko and compare results.

src/yz_pb_search.erl
```diff
@@ -0,0 +1,127 @@
...
+    Value.
+
+encode_doc(Doc) ->
+    riak_pb_search_codec:encode_search_doc([{Prop, to_binary(Value)} || {Prop, Value} <- Doc]).
+
+to_binary(A) when is_atom(A) -> to_binary(atom_to_list(A));
+to_binary(B) when is_binary(B) -> B;
+to_binary(I) when is_integer(I) -> to_binary(integer_to_list(I));
+to_binary(F) when is_float(F) -> to_binary(float_to_list(F));
+to_binary(L) when is_list(L) -> list_to_binary(L).
```
rzezeski: This function is suspect in regards to non ascii scripts.  It will not work for a non-ascii string.  The unicode module must be used.

```
1> S = "א בְּרֵאשִׁית, בָּרָא אֱלֹהִים, אֵת הַשָּׁמַיִם, וְאֵת הָאָרֶץ".
[1488,32,1489,1456,1468,1512,1461,1488,1513,1460,1473,1497,
 1514,44,32,1489,1464,1468,1512,1464,1488,32,1488,1457,1500,
 1465,1492,1460,1497|...]
2> 
2> B = list_to_binary(S).
** exception error: bad argument
     in function  list_to_binary/1
        called as list_to_binary([1488,32,1489,1456,1468,1512,1461,1488,1513,1460,1473,1497,
                                  1514,44,32,1489,1464,1468,1512,1464,1488,32,1488,1457,1500,
                                  1465,1492,1460|...])
3> B = unicode:characters_to_binary(S).
<<215,144,32,215,145,214,176,214,188,215,168,214,181,215,
  144,215,169,214,180,215,129,215,153,215,170,44,32,215,
  145,...>>
```

The riak test verifies that non-ascii may be stored and search, but it doesn't verify that it is retuned properly.  It would be good to add a stored field of non-ascii and verify that the same exact value is returned.

src/yz_app.erl
```diff
@@ -24,6 +24,9 @@
 -compile(export_all).
 -include("yokozuna.hrl").
 
+%% 27 is message type rpbsearchqueryreq
+%% 28 is message type rpbsearchqueryresp
+-define(SERVICES, [{yz_pb_search, 27, 28}]).
```
rzezeski, line 29: What happens if both RS and YZ are enabled and running?  That will be the case when users upgrade.

riak_test/yz_pb.erl
```diff
@@ -0,0 +1,114 @@
...
+-include_lib("eunit/include/eunit.hrl").
+
+-define(FMT(S, Args), lists:flatten(io_lib:format(S, Args))).
+-define(NO_HEADERS, []).
+-define(NO_BODY, <<>>).
+
+confirm() ->
+    case riak_client_exists() of
+        true ->
+            YZBenchDir = rt:get_os_env("YZ_BENCH_DIR"),
```
rzezeski: You're gonna want to rebase against master and change `rt:get_os_env` to `rt_config:get_os_env`.  Riak Test recently changed.  You didn't hit this problem because the verify script doesn't run a git pull.

src/yz_app.erl
```diff
@@ -35,6 +38,7 @@ start(_StartType, _StartArgs) ->
         {ok, Pid} ->
             Routes = yz_wm_search:routes() ++ yz_wm_extract:routes() ++ yz_wm_index:routes() ++ yz_wm_schema:routes(),
             yz_misc:add_routes(Routes),
+            ok = riak_api_pb_service:register(?SERVICES),
```
rzezeski, line 41: Should this be after the service is marked up?  If a pb query lands before the service is marked up do bad things happen?

src/yz_pb_search.erl
```diff
@@ -0,0 +1,127 @@
...
+    {ok, riak_pb_codec:decode(Code, Bin)}.
+
+%% @doc encode/1 callback. Encodes an outgoing response message.
+encode(Message) ->
+    {ok, riak_pb_codec:encode(Message)}.
+
+%% @doc process/2 callback. Handles an incoming request message.
+process(Msg, #state{client=_Client}=State) ->
+    #rpbsearchqueryreq{index=IndexBin, sort=_Sort0,
+                       fl=_FL0, presort=_Presort0}=Msg,
```
rzezeski: No need to match against all the fields in record if they are just ignored.

src/yz_pb_search.erl
```diff
@@ -0,0 +1,127 @@
...
+    case extract_params(Msg) of
+        {ok, Params} ->
+            Mapping = yz_events:get_mapping(),
+            Index = binary_to_list(IndexBin),
+            try
+                {_Headers, Body} = yz_solr:dist_search(Index,
+                                                       [{content_type, "application/json"}],
```
rzezeski: The headers here are unnecessary, this is a search so no body is sent.

riak_test/yz_pb.erl
```diff
@@ -0,0 +1,114 @@
+%% Running this test requires the `riak-erlang-client' project
```
rzezeski: I'm not sure if it is a recent change to riak test but I didn't need to bring in riakc.  It's under deps in riak test.

riak_test/yz_pb.erl
```diff
@@ -0,0 +1,114 @@
...
+-module(yz_pb).
+-compile(export_all).
+-include_lib("eunit/include/eunit.hrl").
+
+-define(FMT(S, Args), lists:flatten(io_lib:format(S, Args))).
+-define(NO_HEADERS, []).
+-define(NO_BODY, <<>>).
+
+confirm() ->
+    case riak_client_exists() of
```
rzezeski: I say get rid of the check and bomb out hard if client isn't available.  This test should always be run.

rzezeski:
So I was able to verify that unicode encoding does not get messed up in process of pb search.  The user just has to be aware to use `unicode:characters_to_binary` to properly convert the byte sequence.

So the `to_binary` function is probably safe BUT that's only because it is always binary type being returned for text data.

```
curl -X PUT -H 'content-type: application/json' 'http://localhost:10018/riak/encoded/stored' -d '{"name_s":"בָרָא", "text":"א בְּרֵאשִׁית, בָּרָא אֱלֹהִים, אֵת הַשָּׁמַיִם, וְאֵת הָאָרֶץ"}'

22> test_pb:go().
R: {ok,{search_results,[{"encoded",
                         [{<<"_yz_id">>,<<"test_28">>},
                          {<<"_yz_ed">>,
                           <<"20130503T033738 28 encoded test g2IHj38+">>},
                          {<<"_yz_fpn">>,<<"26">>},
                          {<<"_yz_node">>,<<"dev4@127.0.0.1">>},
                          {<<"_yz_pn">>,<<"28">>},
                          {<<"_yz_rk">>,<<"test">>},
                          {<<"_yz_rb">>,<<"encoded">>},
                          {<<"score">>,<<"1.85792129999999999779e-01">>}]},
                        {"encoded",
                         [{<<"name_s">>,
                           <<215,145,214,184,215,168,214,184,215,144>>},
                          {<<"_yz_id">>,<<"stored_49">>},
                          {<<"_yz_ed">>,
                           <<"20130503T044453 49 encoded stored g2IFGaM9">>},
                          {<<"_yz_fpn">>,<<"47">>},
                          {<<"_yz_node">>,<<"dev1@127.0.0.1">>},
                          {<<"_yz_pn">>,<<"49">>},
                          {<<"_yz_rk">>,<<"stored">>},
                          {<<"_yz_rb">>,<<"encoded">>},
                          {<<"score">>,<<"9.58915050000000018793e-02">>}]}],
                       0.18579213321208954,2}}
R2: {ok,{search_results,[],0.0,0}}
ok

23> unicode:characters_to_list(<<215,145,214,184,215,168,214,184,215,144>>).
[1489,1464,1512,1464,1488]
24> io:format("~ts~n", [unicode:characters_to_list(<<215,145,214,184,215,168,214,184,215,144>>)]).
בָרָא
```

Can't just `binary_to_list`

```
25> io:format("~ts~n", [binary_to_list(<<215,145,214,184,215,168,214,184,215,144>>)]).                        
×Ö¸×¨Ö¸×
ok
```

riak_test/yz_languages.erl
coderoshi, line 41: Good point. I actually removed it because it was never called... I actually forgot I removed it.

src/yz_app.erl
coderoshi, line 29: It can't be. We have the same problem with the /search WM endpoint. The client API will only allow one range of messages at a time, and to use the searchreq/resp messages, this range is it.

src/yz_app.erl
coderoshi, line 41: Why would it be any different than adding wm routes before the service is up?

src/yz_pb_search.erl
coderoshi: true, this should be accept

src/yz_app.erl
rzezeski, line 29: Yea, the `solr/select` end-point issue is a bug and will need to be dealt with before putting yoko in master.  This will also have to be dealt with.  But that can be a different PR, probably under #20.  Migrating users will have both systems running temporarily and we need a nice story there.

src/yz_pb_search.erl
coderoshi: I was attempting to match the current functionality of search, which doesn't actually use `fl`, though I suppose there's no danger in adding it.

src/yz_app.erl
rzezeski, line 41: Ah right, the routes need to be avail before considering the yoko service up.  Same for the PB server.  I do wonder if HTTP/PB requests could come in before service is marked up and thus cause errors.  E.g. creating a coverage plan requires that the yokozuna service be marked as up, but running a HTTP/PB request requires that a coverage plan be calculated.  So a chicken and egg scenario.  I wonder if @seancribbs knows of any startup races between PB and service?

src/yz_pb_search.erl
rzezeski: Solr ignores the `accept` header.  It bases the content body solely on `wt` query value.  To add insult to injury, even when you specify `wt=json` it returns with `content-type: text/plain;charset=UTF-8`.  You can't explain that. :)

src/yz_pb_search.erl
rzezeski: Gotcha.  I think that is a slippery slope.  There are other oddities we would need to enforce:
- Can only apply `sort` when `fl` is set to unique id, otherwise must return an error (basho/riak_search#86).
- The `sort` parameter must be ignored (basho/riak_search#136).
- Possibly other weird things...

I think Yokozuna should fix these issues rather than reproduce them.  Having these issues is not a feature on which anyone should be relying on.  E.g. I don't see why someone would complain because something that used to give them an error now actually works.

src/yz_app.erl
seancribbs, line 41: This problem occurs with other apps that provide PB services, so there is not really a way to address it yet. Generally, it is good enough to put your riak_api registrations after the riak_core one.

src/yz_pb_search.erl
```diff
@@ -0,0 +1,137 @@
...
+                                rows=Rows, start=Start,
+                                filter=Filter, fl=FilterList,
+                                df=DefaultField, op=DefaultOp}) ->
+    {ok, [{q, Query},
+          {wt, "json"},
+          {omitHeader, true},
+          {'q.op', default(DefaultOp, "AND")},
+          {sort, default(Sort, "")},
+          {fq, default(Filter, "")},
+          {fl, default(FilterList, <<"*,score">>)},
```
rzezeski: It's a field list, not filter :)

I also just realized that PB search always wants a max score.  That will only be available if `score` is passed to Solr.  So it seems we have to make sure `score` is always part of `fl`.  E.g. pretty sure a PB search with `fl=_yz_rk` would bomb out.

src/yz_pb_search.erl
```diff
@@ -0,0 +1,137 @@
...
+            catch
+                throw:not_found ->
+                    ErrMsg = io_lib:format(?YZ_ERR_INDEX_NOT_FOUND, [Index]),
+                    {error, ErrMsg, State};
+                throw:insufficient_vnodes_available ->
+                    {error, ?YZ_ERR_NOT_ENOUGH_NODES, State};
+                throw:{Message, URL, Err} ->
+                    ?INFO("~p ~p ~p~n", [Message, URL, Err]),
+                    {error, Message, State};
+                _:_ ->
```
rzezeski: This msg tells the user to look in the logs but this handler doesn't log anything.  It should log the error type, reason and the stracktrace.

However, I wonder if this catch-all is the proper thing to do.  This will catch anything: system error, process exists, user throws, etc.  Do we really want to catch all those possible errors and surpress them?  I feel like we should only deal with user throws that we expect and let everything else bubble up.

(As a side note I made bad use of throw in early versions of Yokozuna.  A lot of places that throw should instead return `{error, Reason}` tuples.  But that's outside the scope of this change).

src/yz_pb_search.erl
```diff
@@ -0,0 +1,137 @@
...
+    Value.
+
+encode_doc(Doc) ->
+    riak_pb_search_codec:encode_search_doc([{Prop, to_binary(Value)} || {Prop, Value} <- Doc]).
+
+to_binary(A) when is_atom(A) -> to_binary(atom_to_list(A));
+to_binary(B) when is_binary(B) -> B;
+to_binary(I) when is_integer(I) -> to_binary(integer_to_list(I));
+to_binary(F) when is_float(F) -> to_binary(float_to_list(F));
+to_binary(L) when is_list(L) -> unicode:characters_to_binary(L).
```
rzezeski: This still isn't right.  The `mochijson2:decode` function will convert all JSON string to binary.  It will convert JSON arrays to a list.  So `L` will always be a list of arbitrary erlang terms.  Which means we should encode it with `term_to_binary`.  The problem is I don't think we have any easy way to make PB search smart enough to know to decode that back with `binary_to_term` on the client side.

This is important because Solr returns a multiValued store field as a JSON array.  I'm not sure how to work around this problem.

riak_test/yz_pb.erl
```diff
@@ -0,0 +1,101 @@
...
+    {Host, Port} = HP,
+    %% populate a value
+    {ok, "204", _, _} = ibrowse:send_req(URL, [{"Content-Type", "text/plain"}], put, Body),
+    %% Sleep for soft commit
+    timer:sleep(1100),
+    {ok, Pid} = riakc_pb_socket:start_link(Host, (Port-1)),
+    lager:info("search_results"),
+    {ok,{search_results,[{Bucket,Results}],Score,Found}} =
+            riakc_pb_socket:search(Pid, Bucket, Search, Params),
+    Bucket == binary_to_list(proplists:get_value(<<"_yz_rk">>, Results)),
```
rzezeski: This test is not actually verifying anything.  It's just doing term comparison.  I.e. '=' vs. '=='.  It makes more sense to use `?assertEqual` so you can see both expected and actual value.

src/yz_pb_search.erl
coderoshi: I added the log message, but I agree, sending an error rather than throw will be addressed in #57

src/yz_pb_search.erl
coderoshi: In my tests you'll get a `maxScore` whether you add it to the `fl` or not. The only difference with adding score to the `fl` is whether or not its pair is in the doc response.

PR #103, opened 2013-05-02
Use mochiglobal to cache + get coverage plans

rzezeski:
Creating a coverage plan consumes a lot of CPU and creates potentially
large lists (garbage) on the process potentially leading to GC.  This
gets worse as the ring size increases.

Currently Yokozuna creates a coverage plan _per_ request.  I.e. every
search request must first pay the expensive cost of calculating a plan
before running the search.  But it doesn't have to be this way.  The
plans could be periodically calculated in an out-of-band process and
cached in a good read-heavy solution like mochiglobal.

There is no reason to calc a plan per request.  The argument could be
made that this is the only way to avoid sending requests to downed
nodes but that is false.  Coverage relies on nodewatcher which uses
net-kernel tick which runs on a delay of ~15s.

The other argument could be ring change (join/remove a node).  In this
case the cached plan might be in disagreement with the ring.  The
divergence would last for a maximum upper bound of T where T is the
period of time before a new plan is written to the index.  This is
already a problem in Riak where the ring changes between the time a
coordinator fires of a request and the vnodes see it.  I'm just going
to punt and assume vnode forwarding should take care of this.  Even if
this isn't the case this scenario is made more rare by the fact that
the partition that has moved must be part of the plan.  This can also
be made even more rare by invalidating the cache on ring change events.

This patch makes the following changes:
- A new plan is generated for each index every 2 seconds.
- During a search request the cache is first checked.  The cache uses
  mochiglobal which generates BEAM code dynamically so that the cost
  of a fetch is that of a function call.
- If the cache doesn't have a plan stored for the index then a plan
  will be generated as part of the request.
- The `yz_cover` module is now a `gen_sever`.  It's responsibility is
  to periodically re-generate plans for each index and cache them.

NOTE: Another thing that might worry someone about this patch is that
using the same plan means not load-balancing well.  But this matters
less re Yokozuna for two reasons.
1. Yokozuna doesn't have vnodes, so balancing across vnodes doesn't
   matter.
2. Coverage currently seems to always select all nodes.  Once coverage
   gets better at optimizing for nodes then it would be wise to cache
   multiple plans and round-robin them.

### DIFF
--- a/src/yz_cover.erl
+++ b/src/yz_cover.erl
@@ -1,6 +1,6 @@
 %% -------------------------------------------------------------------
 %%
-%% Copyright (c) 2012 Basho Technologies, Inc.  All Rights Reserved.
+%% Copyright (c) 2012-2013 Basho Technologies, Inc.  All Rights Reserved.
 %%
 %% This file is provided to you under the Apache License,
 %% Version 2.0 (the "License"); you may not use this file
@@ -20,6 +20,13 @@
 
 -module(yz_cover).
 -compile(export_all).
+-behavior(gen_server).
+-export([code_change/3,
+         handle_call/3,
+         handle_cast/2,
+         handle_info/2,
+         init/1,
+         terminate/2]).
 -include("yokozuna.hrl").
 
 %% @doc This module contains functionality related to creating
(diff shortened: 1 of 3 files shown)


src/yz_cover.erl
```diff
@@ -1,6 +1,6 @@
 %% -------------------------------------------------------------------
 %%
-%% Copyright (c) 2012 Basho Technologies, Inc.  All Rights Reserved.
+%% Copyright (c) 2012-2013 Basho Technologies, Inc.  All Rights Reserved.
```
coderoshi: fwiw, copyright ranges have no legal meaning.

rzezeski:
Below are query bench results on my imac and linux desktop.  This bench runs the same query over and over so it's always hitting the solr query cache.  This way time in solr is minimized.  This is the appropriate way to bench this change because this patch affects the yokozuna-side of things.  It does, hoever, mean that in cases where Solr dominates the time taken that it will have less impact.  Also, the slower the CPU the more this cache should help.

## iMac Results
- 3.4GHz i7 (4 cores w/ HT)
- 32 GB 1600MHz DDR3
- Partitioned in such a way that all riak I/O is against HDD of fusion drive - i.e. SSD is not involved at all.

```
+----------+-------------------+--------------------+--------------------+--------------------+
|          |Throughput (ops/s) |Median Latency (ms) |95th Latency        |99th Latency        |
+----------+-------------------+--------------------+--------------------+--------------------+
|no cache  |843                |39.97               |47.30               |51.74               |
+----------+-------------------+--------------------+--------------------+--------------------+
|cache     |1206 (143%)        |26.96 (67%)         |32.76 (69%)         |38.48 (74%)         |
+----------+-------------------+--------------------+--------------------+--------------------+
```

### No Cache

query-lima-no-cache

### Cache

query-lima-mochi

## Linux Box Results
- Intel i7 3.5GHz - 6 core w/ HT
- 32GB 1333MHz DDR3
- SSD - Intel 520 SATA-3

```
+----------+-------------------+--------------------+--------------------+--------------------+
|          |Throughput (ops/s) |Median Latency (ms) |95th Latency        |99th Latency        |
+----------+-------------------+--------------------+--------------------+--------------------+
|no cache  |1314               |24.29               |31.37               |35.55               |
+----------+-------------------+--------------------+--------------------+--------------------+
|cache     |1718 (131%)        |18.41 (76%)         |25.58 (82%)         |30.85 (87%)         |
+----------+-------------------+--------------------+--------------------+--------------------+
```

### No Cache

query-lima-no-cache

## Cache

query-lima-mochi

rzezeski:
Also, the coverage query runtime is proportional to the ring size (I'm not sure what the function plot looks like, linear at best).  So this caching helps even more with larger rings.  In the bench runs above a default ring size of 64 was used.

src/yz_cover.erl
```diff
@@ -157,3 +220,22 @@ make_logical(LogicalIndex, Cover) ->
 partition(LogicalIndex, LP) ->
     {_, P} = lists:keyfind(LP, 1, LogicalIndex),
     P.
+
+%% @private
+%%
+%% @doc Schedule next tick to be sent to this server.
+-spec schedule_tick() -> ok.
+schedule_tick() ->
+    erlang:send_after(2000, ?MODULE, tick),
```
coderoshi: Magic number! A constant would be nice.

coderoshi:
This PR does two things, and those things it does well. +1

PR #112, opened 2013-05-10
Handle Errors in a more reasonable manner.

coderoshi:
This is an alternate pr for #57 (github won't allow multiple PRs per issue).

The goal here is to allow writes to occur in KV, even if YZ has a problem indexing. This can happen for many reasons, such as a failing extractor (eg. json that is unparseable), or errors other than 200 returned from solr. Rather than crashing, we instead allow the write to complete on the KV side, attempt to index, and any failing indexes are flagged with a `_yz_err` field. This allows a user to search for values that failed to index for whatever reason (the true reason will be logged).


src/yz_doc.erl
```diff
@@ -94,12 +94,17 @@ extract_fields({MD, V}) ->
...
-            case yz_extractor:run(V, ExtractorDef) of
-                {error, Reason} ->
-                    ?ERROR("failed to index with reason ~s~nValue: ~s", [Reason, V]),
-                    {error, Reason};
-                Fields ->
-                    Fields
+            try
+                case yz_extractor:run(V, ExtractorDef) of
+                    {error, Reason} ->
+                        ?WARN("failed to index fields from value with reason ~s~nValue: ~s", [Reason, V]),
```
rzezeski: This should be error level.

src/yz_doc.erl
```diff
@@ -94,12 +94,17 @@ extract_fields({MD, V}) ->
...
+                        [{?YZ_ERR_FIELD_S, 1}];
+                    Fields ->
+                        Fields
+                end
+            catch _:Err ->
+                ?ERROR("failed to index fields from value with reason ~s~nValue: ~s", [Err, V]),
```
rzezeski: Add the stacktrace to the message.

src/yz_kv.erl
```diff
@@ -169,9 +169,9 @@ index(Obj, Reason, Ring, P, BKey, IdxN, Index, IndexContent) ->
         ok = update_hashtree({insert, yz_kv:hash_object(Obj)},
                              P, IdxN, BKey)
     catch _:Err ->
-            ?ERROR("failed to index object ~p with error ~p", [BKey, Err])
-    end,
-    ok.
+        ?ERROR("failed to index object ~p with error ~p", [BKey, Err]),
```
rzezeski: I know it wasn't here before, but add the stacktrace to the message.

riak_test/yz_errors.erl
```diff
@@ -0,0 +1,116 @@
...
+    {ok, "204", _, _} = ibrowse:send_req(URL, Headers, put, Body, Opts),
+    %% Sleep for soft commit
+    timer:sleep(1100),
+    %% still store the value in riak
+    {ok, "200", _, Body} = ibrowse:send_req(URL, [{"accept", CT}], get, []),
+    ?assert(search_expect(HP, "bad_json", ?YZ_ERR_FIELD_S, "1", 1)),
+    ok.
+
+expect_bad_xml(HP) ->
+    ok = create_index(HP, "bad_xml"),
```
rzezeski, line 61: Rather than create an index for each test just use one for all of them and save some time.

coderoshi, line 61: We're counting the number of ?YZ_ERR_FIELD_S matches, which in this case will always be 1 since it's a new index, rather than keeping track of the count.

src/yz_doc.erl
```diff
@@ -94,12 +94,18 @@ extract_fields({MD, V}) ->
...
+                case yz_extractor:run(V, ExtractorDef) of
+                    {error, Reason} ->
+                        ?ERROR("failed to index fields from value with reason ~s~nValue: ~s", [Reason, V]),
+                        [{?YZ_ERR_FIELD_S, 1}];
+                    Fields ->
+                        Fields
+                end
+            catch _:Err ->
+                Trace = erlang:get_stacktrace(),
+                ?ERROR("failed to index fields from value with reason ~s~nValue: ~sTrace: ~p~n", [Err, V, Trace]),
```
rzezeski: This will be hard to read putting the trace right up against the value.  I say do `reason ~s ~s~nValue...[Err, Trace,...]`.

src/yz_kv.erl
```diff
@@ -169,9 +169,10 @@ index(Obj, Reason, Ring, P, BKey, IdxN, Index, IndexContent) ->
         ok = update_hashtree({insert, yz_kv:hash_object(Obj)},
                              P, IdxN, BKey)
     catch _:Err ->
-            ?ERROR("failed to index object ~p with error ~p", [BKey, Err])
-    end,
-    ok.
+        Trace = erlang:get_stacktrace(),
+        ?ERROR("failed to index object ~p with error ~p because ~p~n", [BKey, Err, Trace]),
+        {error, Err}
```
rzezeski: For the time being `index` should be returning `ok`.  In the future we might want to return `ok | {error, term()}` when we do the vnode hooks, but for now stick with the spec.

include/yokozuna.hrl
```diff
@@ -320,6 +320,14 @@
 -define(YZ_RB_FIELD_XML, ?YZ_FIELD_XML(?YZ_RB_FIELD_S)).
 -define(YZ_RB_FIELD_XPATH, "/schema/fields/field[@name=\"_yz_rb\" and @type=\"_yz_str\" and @indexed=\"true\" and @stored=\"true\"]").
 
+
+%% Riak extraction error
+-define(YZ_ERR_FIELD, '_yz_err').
+-define(YZ_ERR_FIELD_S, "_yz_err").
+-define(YZ_ERR_FIELD_B, <<"_yz_err">>).
+-define(YZ_ERR_FIELD_XML, ?YZ_FIELD_XML(?YZ_ERR_FIELD_S)).
+-define(YZ_ERR_FIELD_XPATH, "/schema/fields/field[@name=\"_yz_err\" and @type=\"_yz_str\" and @indexed=\"true\" and @stored=\"true\"]").
```
rzezeski: Don't store the field.  We shouldn't be querying it enough to take advantage of a filter query cache and there is no reason to include the value with the doc since it's a binary attribute.

rzezeski:
+1 to merge

PR #111, opened 2013-05-10
Remove redundant yz_json_extractor_with_suffix

coderoshi:
Now that the default index config indexes all non-suffixed values as a string, this extra extractor should probably be retired.


rzezeski:
Looks like you forgot to pull the unit tests.

PR #114, opened 2013-05-11
add instruction about enabling yokozuna to README.

ksauzz:
I tried to add the instruction how to enable yokozuna at 'Getting started'. This one-liner attempts to toggle `enabled` option to `true` in yokozuna section only. It may not seem to be a good way but It works for now. 

thanks.

refs: https://github.com/basho/riak/pull/324


rzezeski:
+1

Issue #109, opened 2013-05-06
Don't crash cover server on `insufficient vnodes`


rzezeski:
While testing issue #104 I discovered that starting Yokozuna on-the-fly can cause the `yz_cover` server to crash because the other nodes aren't up and thus the coverage plan returns `insufficient_vnodes_available`.  The cover server should be modified to deal with this return value and store `undefined` for the plan.  That way a plan will be created during the query request and the query requests will fail (not the cover server) until a plan can be made and cache.

rzezeski:
Fixed by #115.

PR #115, opened 2013-05-12
Don't crash cover on insufficient_vnodes_available

rzezeski:
- Change `calc_plan` to return error-tuple rather than match it and
  throw it.
- Check for error-tuple when generating cached coverage plans and
  store undefined.  That will prevent cover server from crashing.
- Check for error-tuple during search request and return proper error
  msg to user.


rzezeski:
This addresses #109.

coderoshi:
This change also needs handled on #38.

coderoshi:
+1

rzezeski:
+1, works for me.

PR #118, opened 2013-05-21
Close hashtree on application stop

rzezeski:
Make sure to close the Yokozuna hashtrees properly during application
stop.
- Need to trap exits on `yz_index_hashtree` servers so that
  `terminate` callback will be invoked.
- Add `close_trees` call to `terminate` callback.
- Add `closed` flag to state to make sure trees are only closed if
  they are currenty open.


coderoshi:
+1

Issue #121, opened 2013-05-25
Upgrade to Solr 4.3.0


rzezeski:
Update to Solr 4.3.0.

http://lucene.apache.org/solr/4_3_0/changes/Changes.html
- Lookout for logging changes - make sure logging works.
- Update codec/version in config.
- Take close look at `solr.xml` changes.  Don't think anything has to change for this upgrade but there will be changes for Solr 4.4 and 5.x.

coderoshi:
done?

PR #122, opened 2013-05-29
Upgrade to Solr 4.3.0

rzezeski:
- Upgrade to latest Solr.
- Remove JSON query response writer (so that Solr will actually return `application/json` content type)
- Remove unused lib directives in `solrconfig.xml`.


rzezeski:
This addresses #121.

coderoshi:
+1, works as advertised. My issue was a a borked Java install.

Issue #87, opened 2013-04-08
Support searching across buckets that share a schema [JIRA: RIAK-1710]


eliaslevy:
It should be possible to perform a distributed search across buckets (indexes) that share a common schema by Solr's distributed search shards parameter.  This would be useful when we wish to primarily store and search a subset of data within a bucket (e.g. a single customer's data) but at times we wish to search across a larger set of data (e.g. all customers).

Solr's distributed search shards parameter could be overloaded, so that when a client makes a query against the Riak cluster it passes a shards parameter that contains a list of buckets/indexes to query.  Yokozuna could then replace this shards parameter with a list of the actual internal shards to use for the distributed search query.

rzezeski:
I'll have to think about this some.  I've looked through Solr's dist search code a few times now.  I don't claim to know all the details but off the top of my head I feel like this will create an HTTP request per shard.  Not node, but shard, so C customers and N nodes means N*C requests, and that is just for the 1st phase.  I'm worried what affect that might have on latencies.

Overloading the shards param doesn't make sense because shards in Solr represents a list of disjoint host/port pairs that make up _one_ index [1].  I actually wonder if Solr's dist search will puke on a shards URL with multiple cores included.  If yokozuna were to support this I think a comma delimited list in the URL or a header like 'yz-indexes' would make more sense.  But first would have to see if this is feasible.

Another way to solve this problem is index multiple riak buckets under one yokozuna index.  But that's still a vaporware feature ATM.

[1]: In fact, Yokozuna should make sure to strip out any 'shards' param passed to it or else a user could unknowingly break the request.

rzezeski:
Another thought.  Each solr shard will have to return R results (default of 10, based on `rows` param).  So the total number of docs returned back on 1st phase of query is N_I_R.  Where N is number of nodes, I is number of indexes queried, R is number of rows to return in final results.

So 8 nodes, 3 customers, and 10 rows is 240 docs returned by 1st phase of dist search.  Bump rows up to 100 and now it's 2400.

eliaslevy:
[...]

PR #90, opened 2013-06-11
1:M index:bucket mapping

rzezeski:
Allow a one-to-many index-to-bucket mapping.  This allows splitting data into logical buckets without requiring the use of multiple Solr Cores.
- Change the index content flag to instead be the name of the index for which the bucket indexes under.  I.e. change the bucket property name from `yz_index_content` to `yz_index`.
- The bucket property is already being pulled during index time so just change the index hook to use the value as the index to write to.

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,16 +25,18 @@ development cluster.
 
 ### Creating an Index ###
 
-An _index_ must be created in order for Yokozuna to index data.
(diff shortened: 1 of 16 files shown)


coderoshi:
+1 fwiw

rzezeski:
Add reference to #87 which mentioned this functionality.

Issue #124, opened 2013-06-04
JVM Self Destruct


coderoshi:
There are many reasons why a JVM may continue running after it's parent erlang process has died, from core dumps to running older Erlang prior to R15B02.

The solr JVM should run a thread which occasionally checks if its PPID=1. If so, the erlang parent process is crashed, and the JVM process should `System.exit()` itself.

Note that there are two parts to this task:
1. get the solr web service to launch a low priority background thread that runs every few seconds.
2. having that thread check for a ppid (which will require some external ps call, or better yet, just writing a script which checks).

Note that the JVM does not have any built-in method for getting a PPID, and many of the "get the PID this way" hacks found out on the interwebs are unreliable (eg. ManagementFactory.getRuntimeMXBean().getName() is not guaranteed).

dmitrizagidulin:
Quick question - does that mean that the Solr instance that YZ spins up should exit once the erlang parent crashes?

coderoshi:
@dmitrizagidulin absolutely yes

dmitrizagidulin:
Ok, cool. Yeah, I've been having to constantly kill -9 the Solr instance, manually.

hazen:
Saving @coderoshi a step by adding his comments:
"there's two parts to this: getting the solr web service to launch a low priority background thread that checks every few seconds, and then having that thread check for a ppid (which will require some external ps call, or better yet, just writing a script which checks). the jvm does not have any built-in method for getting a ppid, and many of the "get the pid this way" hacks are unreliable (eg. ManagementFactory.getRuntimeMXBean().getName() is not guaranteed)"

coderoshi:
@dmitrizagidulin if you've been having to kill -9 the java procs every time, you're probably running R15B01. Only R15B02 is capable of stopping the process when you kill erlang.

dmitrizagidulin:
@coderoshi ahh, I think that's it, yes. Thanks.

coderoshi:
Here's one way to call any custom class through jetty: http://wiki.eclipse.org/Jetty/Reference/jetty.xml_syntax. Example:

```
<Call name="takePill">
  <Arg>
    <New class="com.basho.riak.yokozuna.Cyanide">
    </New>
  </Arg>
</Call>
```

That'll create that object. you can optionally add a new function that it calls, or just spawn the thread via the constructor

PR #127, opened 2013-06-13
JVM Self Destruct

hazen:
This PR closes Issue #124

### DIFF
--- a/priv/build-jar.sh
+++ b/priv/build-jar.sh
@@ -26,9 +26,12 @@ if [ ! -e $SOLR_JAR_DIR ]; then
 fi
 
 echo "Compile..."
-javac -cp "$SOLR_JAR_DIR/*" java/com/basho/yokozuna/handler/*.java java/com/basho/yokozuna/query/*.java java/com/basho/yokozuna/monitor/*.java
+javac -cp "$SOLR_JAR_DIR/*" java/com/basho/yokozuna/handler/*.java java/com/basho/yokozuna/query/*.java java_monitor/com/basho/yokozuna/monitor/*.java
 echo "Create yokozuna.jar..."
 mkdir java_lib
 jar cvf java_lib/yokozuna.jar -C java/ .
-jar cvf solr/lib/ext/yz_monitor.jar -C java_monitor/ .
 echo "Finished building yokozuna.jar..."
+
+echo "Create yz_monitor.jar..."
+jar cvf solr/lib/ext/yz_monitor.jar -C java_monitor/ .
+echo "Finished building yz_monitor.jar..."
--- a/priv/build-jar.sh
+++ b/priv/build-jar.sh
@@ -26,8 +26,9 @@ if [ ! -e $SOLR_JAR_DIR ]; then
 fi
 
 echo "Compile..."
-javac -cp "$SOLR_JAR_DIR/*" java/com/basho/yokozuna/handler/*.java java/com/basho/yokozuna/query/*.java
+javac -cp "$SOLR_JAR_DIR/*" java/com/basho/yokozuna/handler/*.java java/com/basho/yokozuna/query/*.java java/com/basho/yokozuna/monitor/*.java
 echo "Create yokozuna.jar..."
 mkdir java_lib
 jar cvf java_lib/yokozuna.jar -C java/ .
+jar cvf solr/lib/ext/yz_monitor.jar -C java_monitor/ .
 echo "Finished building yokozuna.jar..."
--- a/priv/java_monitor/com/basho/yokozuna/monitor/Monitor.java
+++ b/priv/java_monitor/com/basho/yokozuna/monitor/Monitor.java
@@ -1,0 +1,93 @@
+/*
+ * Copyright (c) 2013 Basho Technologies, Inc.  All Rights Reserved.
+ *
+ * This file is provided to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package com.basho.yokozuna.monitor;
+
+import java.io.BufferedReader;
+import java.io.IOException;
+import java.io.InputStreamReader;
+import java.util.Timer;
+import java.util.TimerTask;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Fire off a shell script to kill Solr once spawning Erlang has died
+ * 
+ * @author Brett Hazen
+ */
+public class Monitor {
+    protected static Logger log = LoggerFactory.getLogger(Monitor.class);
+    
+    /**
+     * Launches monitoring bash script
+     * @param monitorScript name of script to run
+     * @param sleepSeconds interval at which to run PPID script
+     */
+    public static void monitor(final String monitorScript, final String sleepSeconds) {
+        int sleepInterval = Integer.parseInt(sleepSeconds);
+        TimerTask readyForSuicide = new MonitorTimerTask(monitorScript);
+        Timer timer = new Timer();
+
+       // scheduling the task at fixed rate
+       timer.schedule(readyForSuicide, sleepInterval*1000);      
+    }
+   
+    /**
+     * Main for testing
+     * @param monitorScript name of script to run
+     * @param sleepSeconds interval at which to run PPID script
+     */
+    public static void main(String[] args) {
+        monitor(args[0], args[1]);
+    }
+    
+    public class MonitorTimerTask extends TimerTask {
+        protected String _monitorScript;
+        
+        MonitorTimerTask(String script) {
+            _monitorScript = script;
+        }
+        
+        @Override
+        public void run() {
+            try {
+                String[] cmd = {_monitorScript};
+                Process p = Runtime.getRuntime().exec(cmd);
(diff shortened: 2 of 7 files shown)


coderoshi:
YES! It works as advertised, at least on my OSX. We need to figure out how to write a test for this, and try it on other machines. But it certainly killed what it needed to kill for me.

rzezeski:
Nice, going to test this today.  But my first review comment is about the commit msg.  Mind doing it in subject + body style (http://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html).  I know it's a bit of a contrived thing but various git tools seem to play better with commit msgs that follow that style.

rzezeski:
Why is the Monitor class under a different path than the other custom Java code?  I.e. `priv/java_monitor`.  I would like to see it all under `priv/java` and in the `yokozuna.jar`.

priv/build-jar.sh
```diff
@@ -26,8 +26,10 @@ if [ ! -e $SOLR_JAR_DIR ]; then
...
 
 echo "Compile..."
-javac -cp "$SOLR_JAR_DIR/*" java/com/basho/yokozuna/handler/*.java java/com/basho/yokozuna/query/*.java
+javac -cp "$SOLR_JAR_DIR/*" java/com/basho/yokozuna/handler/*.java java/com/basho/yokozuna/query/*.java java_monitor/com/basho/yokozuna/monitor/*.java
 echo "Create yokozuna.jar..."
 mkdir java_lib
 jar cvf java_lib/yokozuna.jar -C java/ .
+jar cvf solr/lib/ext/yz_monitor.jar -C java_monitor/ .
 echo "Finished building yokozuna.jar..."
+cp monitor.sh solr
```
rzezeski: The copying of this script should be in the `grab-solr.sh` script.  This script is solely for building the `yokozuna.jar`.  The `grab-solr.sh` is what builds the `priv/solr` directory.

priv/java_monitor/com/basho/yokozuna/monitor/Monitor.java
```diff
@@ -0,0 +1,88 @@
...
+               try {
+                   String[] cmd = {monitorScript};
+                   Process p = Runtime.getRuntime().exec(cmd);
+                   BufferedReader buf = new BufferedReader(new InputStreamReader(p.getInputStream()));
+                   p.waitFor();
+                   // Read single output line to determine PPID
+                   String line = buf.readLine();
+                   int exitValue = p.exitValue();
+                   if (exitValue != 0) {
+                       System.err.println("Problem running Solr monitoring script");
```
rzezeski: This should be `log.error`.

priv/java_monitor/com/basho/yokozuna/monitor/Monitor.java
```diff
@@ -0,0 +1,88 @@
...
+                   }
+                   // Die if our parent is no longer Erlang
+                   int ppid = Integer.parseInt(line);
+                   if (ppid == 1) {
+                       System.err.println("Shutting down Solr after Yokozuna has finished");
```
rzezeski: Get rid of the println.  Use error level.  The message is a bit misleading.  Yokozuna crashed in this case.  It didn't shutdown gracefully.

priv/java_monitor/com/basho/yokozuna/monitor/Monitor.java
```diff
@@ -0,0 +1,88 @@
...
+               } catch(IOException e) {
+                   System.err.println(e.getMessage());
+                   log.error(e.getMessage());
+               } catch(InterruptedException e) {
+                   System.err.println(e.getMessage());
+                   log.error(e.getMessage());
+               }
+           }
+       };
+       Timer timer = new Timer();
```
rzezeski: Every invocation of this method is creating a new `Timer` and scheduling calls.  This is a leak.  The entry point into this class should be a different method which simply starts the timer.

Use `Timer.schedule` for fixed-delay execution rather than fixed-rate and use a 2nd argument of delay rather than date.

hazen:
Nice to see some git commit caremadness.  My "favorite" is when people write things like "Checked in code changes."  I'll look at that website, @rzezeski .

So I initially put all the Java in the same jar, but it needs to be in a different place for Jetty, per my discussions with @coderoshi .  It would be cleaner to have a single, duplicated jar, I guess, if that does not confuse Java.  One is for Jetty and one for Solr (as I understand it.)

hazen:
We seem to have 2 methods of distribution: git build and binary download.  We would not need the copy if I could have updated the Amazon tarball.  I'm happy with killing the patch file, if you still think it's a good idea.

priv/java_monitor/com/basho/yokozuna/monitor/Monitor.java
hazen: I think com.basho.yokozuna.monitor.Monitor.monitor() is only called once, right?  The anonymous run TimerTask "readyForSuicide" is called every _n_ seconds.

coderoshi: `monitor` will only ever be called once, so this won't leak. However, it might be cleaner Java code to implement TimerTask as its own class, rather than make an anonymous class, for clarity. No need for a separate file or anything, within `class Monitor` is fine.

rzezeski: Wow, yea, I totally missed that when reading earlier.  Sorry.

rzezeski:
> We seem to have 2 methods of distribution: git build and binary download. We would not need the copy if I could have updated the Amazon tarball. I'm happy with killing the patch file, if you still think it's a good idea.

Not sure I completely follow.  Everything is built on top of the custom binary build.  The `grab-solr.sh` grabs that custom binary and from that builds the solr dir which is used in the Yokozuna release and everything based on it (e.g. the AMI).  The patch file is redundant.

riak_test/yz_monitor_solr.erl
```diff
@@ -0,0 +1,59 @@
...
+    wait_for_joins(Cluster),
+    rt:wait_for_cluster_service(Cluster, yokozuna),
+    Cluster.
+
+join(Nodes) ->
+    [NodeA|Others] = Nodes,
+    [rt:join(Node, NodeA) || Node <- Others],
+    Nodes.
+
+% Kill the spawning Erlang process and verify the JVM is killed, too
```
rzezeski: Line comments start with double %.

riak_test/yz_monitor_solr.erl
```diff
@@ -0,0 +1,59 @@
...
+-spec test_solr_monitor(list(node())) -> ok
+    ; (list(node())) -> false.
+test_solr_monitor(Cluster) ->
+    Node = hd(Cluster),
+    ErlPid = rpc:call(Node, os, getpid, []),
+    JvmPid = get_jvm_pid(Node),
+    os:cmd("kill -9 " ++  ErlPid),
+    is_jvm_dead_yet(JvmPid, 10).
```
rzezeski: Change `is_jvm_deat_yet` to a pure predicate and use `wait_until` for the repeating.

riak_test/yz_monitor_solr.erl
```diff
@@ -0,0 +1,59 @@
...
+    Cluster.
+
+join(Nodes) ->
+    [NodeA|Others] = Nodes,
+    [rt:join(Node, NodeA) || Node <- Others],
+    Nodes.
+
+% Kill the spawning Erlang process and verify the JVM is killed, too
+-spec test_solr_monitor(list(node())) -> ok
+    ; (list(node())) -> false.
```
rzezeski: `lists(node())` can be expressed as `[node()]`.  Also, I think just using `ok | false` for the return type is clearer than using spec overloading here.

riak_test/yz_monitor_solr.erl
```diff
@@ -0,0 +1,59 @@
...
+-compile(export_all).
+-import(yz_rt, [host_entries/1,
+                wait_for_joins/1]).
+-include_lib("eunit/include/eunit.hrl").
+
+-define(CFG, [{yokozuna, [{enabled, true}]}]).
+
+confirm() ->
+    YZBenchDir = rt_config:get_os_env("YZ_BENCH_DIR"),
+    code:add_path(filename:join([YZBenchDir, "ebin"])),
```
rzezeski: The custom bench code isn't needed for this test.

rzezeski:
Would like to have @jaredmorrow give a look at `monitor.sh` to make sure he doesn't see anything that would cause issues on certain platforms.

jaredmorrow:
`monitor.sh` is fine I looked at it and just checked out on my most picky platform.  The other .sh files though are bash scripts so you'd obviously need bash installed for those to work.

rzezeski:
@jaredmorrow thx for the quick turn around.  That's a good point about the other scripts.  They are used during the actual building of Yokozuna so once we start migrating into master I suppose it will be important to update those if we want them to build w/o requiring installing bash?  Not sure how much I rely on bash-only behavior or non-standard arguments to various commands.  These build-time issues can be dealt with once that integration starts happening (hopefully in July).

jaredmorrow:
All of my builders have bash installed.  So if they are only build time, I think we are fine.

rzezeski:
Hey Brett, I think you re-ran the verify script without blowing away your verify directory first.  You need to do this when you change files involving the `solr` dir because that stuff is not symlinked like the beam files.  Thus you were just testing your old code.

Had to make a few changes to get it to work but it passes for me now.

hazen:
Bummer. I was hoping to have tested the right thing.  Guess I was a bit too hasty.  I need to nuke _solr_ to get that part to rebuild properly, right? I wonder if we should include that in the "make clean" target?

rzezeski:
Not just the solr dir.  The `rt` and `riak_yz` in the verify dir also need be nuked because the devrels need to be rebuilt.

Not sure we want to delete solr during clean.  That will cause the tar to be downloaded (although perhaps that should be changed, only download if the md5 doesn't match).

Perhaps a dist-clean or something.  Or add a solr-clean and clean-all?

PR #48, opened 2013-06-16
Send Delete & Update in Same Request

rzezeski:
Currently the `yz_kv:index` code uses two separate requests to a) cleanup siblings and b) index latest data.  This can be done in one request.
- Verify that the Solr XML/JSON update code performs operations in the order given.  If this doesn't hold true then 2 requests must be used to enforce correct ordering.
- Run a benchmark to compare 2 requests vs. 1 request and verify no regression (ideally, it should help performance).

I already have a branch for this code but it was written against Yokozuna when it still used XML update.  It needs to be updated for the latest.

https://github.com/rzezeski/yokozuna/tree/combine-index-and-del

### DIFF
--- a/include/yokozuna.hrl
+++ b/include/yokozuna.hrl
@@ -19,29 +19,6 @@
 %% -------------------------------------------------------------------
 
 %%%===================================================================
-%%% Records
-%%%===================================================================
-
--record(yz_index_cmd, {
-          doc :: doc(),
-          index :: string(),
-          req_id :: non_neg_integer()
-         }).
-
--record(yz_search_cmd, {
-          qry :: term(),
-          req_id :: non_neg_integer()
-         }).
-
-%% A reference to a merkle tree.
--record(tree_ref, {
-          index :: string(),
-          name :: tree_name(),
-          pid :: pid(),
-          ref :: reference()
-         }).
-
-%%%===================================================================
 %%% Types
 %%%===================================================================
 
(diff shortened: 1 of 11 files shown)


rzezeski:
On my MacBook Pro this patch achieved:
- 15% increase in write throughput.
- A 17% drop in write median latency (11ms).
- A 15% drop in write 95th latency (15ms).

### Before (Separate Requests)

diff-reqs

### After (Same Request)

same-req

hazen:
Very nice upgrade!

include/yokozuna.hrl
```diff
@@ -316,7 +296,7 @@
 %% Riak bucket
 -define(YZ_RB_FIELD, '_yz_rb').
 -define(YZ_RB_FIELD_S, "_yz_rb").
--define(YZ_RB_FIELD_B, <<"_yz_rk">>).
+-define(YZ_RB_FIELD_B, <<"_yz_rb">>).
```
coderoshi, line 299: lol

rzezeski, line 299: Yea I know.  Luckily this macro was not in use anywhere.

src/yz_kv.erl
```diff
@@ -269,20 +268,17 @@ check_flag(Flag) ->
     true == erlang:get(Flag).
 
 %% @private
-cleanup(1, {Index, _Obj, Key, _LP}) ->
+cleanup(1, {_Index, _Obj, Key, _LP}) ->
```
coderoshi: Index isn't used anywhere anymore

coderoshi:
Looks good, tests pass, +1

PR #128, opened 2013-06-19
Serialize index creation through claimant

rzezeski:
If index creation events are executed on disjoint nodes in a tight enough time span then the operations can create data races on the ring and false removal events can be generated by the `yz_events` delta logic.  Rather than try to get clever and allow concurrent operations instead serialize all index operations through the claimant.  This means that index creation/removal cannot be performed if the claimant is down, but avoids the data race.

### DIFF
--- a/src/yz_index.erl
+++ b/src/yz_index.erl
@@ -47,7 +47,7 @@ create(Name) ->
 %%
 %% NOTE: All create requests are serialized through the claimant node
 %%       to avoid races between disjoint nodes.  If the claimant is
-%%       down not indexes may be created.
+%%       down no indexes may be created.
 -spec create(string(), schema_name()) -> ok |
                                          {error, schema_not_found} |
                                          {error, {rpc_fail, node(), term()}}.
--- a/src/yz_index.erl
+++ b/src/yz_index.erl
@@ -37,15 +37,38 @@ create(Name) ->
 
 %% @doc Create the index `Name' across the entire cluster using
 %%      `SchemaName' as the schema.
-%%      Returns 'ok' if created, or 'notfound' if the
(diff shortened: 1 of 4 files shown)


coderoshi:
If you want to cowboy merge this, I'll punt. Otherwise, it'll have to wait until tomorrow.

rzezeski:
I'll  wait for review.

src/yz_index.erl
```diff
@@ -37,15 +37,39 @@ create(Name) ->
...
+%% `ok' - The schema was found and the create request was successfully
+%%        run on the claimant.
+%%
+%% `schema_not_found' - The `SchemaName' could not be found in Riak.
+%%
+%% `rpc_fail' - The claimant could not be contacted.
+%%
+%% NOTE: All create requests are serialized through the claimant node
+%%       to avoid races between disjoint nodes.  If the claimant is
+%%       down not indexes may be created.
```
coderoshi: not = no

src/yz_wm_index.erl
```diff
@@ -141,13 +141,18 @@ delete_resource(RD, S) ->
 %% Responds to a PUT request by creating an index and setting the
 %% index flag for the "index" name given in the route. Returns 204 if
 %% created, or 409 if an index of that name already exists. Returns a
-%% 500 error if the schema does not exist.
+%% 400 error if the schema does not exist.
```
coderoshi, line 144: It's set to 500 below

rzezeski, line 144: That's if the claimant is down.  The schema exists check is done in `malformed_request`.

coderoshi:
Looks good, tests all pass, just trivial comments +1

PR #131, opened 2013-06-20
Avoid HTTP and PB conflicts with Riak Search

rzezeski:
When both Riak Search and Yokozuna are enabled Yokozuna should yield the original Riak Search service endpoints to Riak Search.  This way during rolling upgrades Riak Search can continue handling requests.  Eventually a switch will be added that will tell Yokozuna to take over all query handling so a user can switch from one to the other after they are convinced a successful migration has been performed.

### DIFF
--- a/riak_test/yz_pb.erl
+++ b/riak_test/yz_pb.erl
@@ -78,17 +78,22 @@ store_and_search(Cluster, Bucket, Key, Body, CT, Search, Params) ->
     {Host, Port} = HP,
     %% populate a value
     {ok, "204", _, _} = ibrowse:send_req(URL, [{"Content-Type", CT}], put, Body),
-    %% Sleep for soft commit
-    timer:sleep(1100),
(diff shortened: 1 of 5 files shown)


riak_test/yz_pb.erl
```diff
@@ -78,17 +78,22 @@ store_and_search(Cluster, Bucket, Key, Body, CT, Search, Params) ->
...
+                    1 ->
+                        [{Bucket,Results}] = R,
+                        KeyCheck = (Key == binary_to_list(proplists:get_value(<<"_yz_rk">>, Results))),
+                        ScoreCheck = (Score =/= 0.0),
+                        KeyCheck and ScoreCheck;
+                    0 ->
+                        false
+                end
+        end,
+    yz_rt:wait_until(Cluster, F),
```
coderoshi, line 96: nice. lolbacks

coderoshi:
+1

Issue #19, opened 2012-11-02
Map-Reduce Integration


rzezeski:
Allow Yokozuna query results to be used as input to map-reduce
operations.

rzezeski:
This is what the results fed back from RS into map/reduce look like.  Notice it passes positions and score for each match.  I'm not going to do this by default for Yokozuna as it's expensive to do that and users won't always need that info.  In the future an option can be added to include that information (although in Solr positions are optional per field).  So for this to work Yokozuna needs to execute a query and for each page of results create a `[{Bucket,Key,[]}]` list.

```
(dev1@127.0.0.1)1> {ok,RC} = riak_search:local_client().
{ok,{riak_search_client,{riak_client,'dev1@127.0.0.1',
                                     undefined}}}
(dev1@127.0.0.1)2> {ok,QOps} = C:parse_query(<<"rs1">>, <<"zezeski">>).
* 1: variable 'C' is unbound
(dev1@127.0.0.1)3> {ok,QOps} = RC:parse_query(<<"rs1">>, <<"zezeski">>).
{ok,{scope,#Ref<0.0.5.225935>,"rs1","value",
           {term,#Ref<0.0.5.225937>,<<"zezeski">>,
                 [{'dev1@127.0.0.1',1},
                  {'dev1@127.0.0.1',1},
                  {'dev1@127.0.0.1',1}],
                 1.0,1.0}}}
(dev1@127.0.0.1)4> {ok,FOps} = C:parse_filter(<<"rs1">>, <<"">>).
* 1: variable 'C' is unbound
(dev1@127.0.0.1)5> {ok,FOps} = RC:parse_filter(<<"rs1">>, <<"">>).
{ok,[]}
(dev1@127.0.0.1)6> F = fun(Results, Acc) -> io:format("results: ~p~n", [Results]), end.
* 1: syntax error before: 'end'
(dev1@127.0.0.1)6> F = fun(Results, Acc) -> io:format("results: ~p~n", [Results]) end.
#Fun<erl_eval.12.82930912>
(dev1@127.0.0.1)7> RC:search_fold(<<"rs1">>, QOps, FOps, F, ok, 60000).
results: [{<<"rs1">>,<<"ryan">>,[{p,[1]},{score,[1.0]}]}]
results: []
ok
```

rzezeski:
The java client, and any other clients relying on search/map-reduce protocol buffer support need this issue resolved.  See #106.

rzezeski:
[...]

PR #133, opened 2013-06-25
Add map-reduce support

rzezeski:
Add ability to use Yokozuna query as input for map-reduce job.
- Add a `mapred_search/3` API used to run map-reduce jobs with query
  results generating bucket/key pairs as input.
- Add a `search_fold/5` API used to fold over all results matching a
  query.  Thus, if a query has 143 matches all 143 will be iterated
  and fed to the fold function `F` passed as argument.
- Add a test to verify map-reduce support works.  It writes 100
  objects and runs a map-reduce job whose query matches all 100
  objects.  The job sums the number of objects matched and verifies
  that the value is 100.
- Added the KVC repo as a dependency.  It allows convenient extraction
  of proplists and mochijson structures.
- The `Timeout` parameter to `mapred_search` is ignored.  According to
  Bryan Fink this timeout is for the duration of the entire job, not
  per-batch (which is what Riak Search is using it for).  He also
  mentioned the existing endpoints such as PB and HTTP already enforce
  this timeout and it probably doesn't make sense to also track it in
  Yokozuna.  I think this makes sense because if Yokozuna also tracked
  it then what happens when both systems decide to deal with the
  timeout and kill various processes?  I imagine that could cause
  other complications.  In the future this parameter should be looked
  at carefully and potentially removed from this part of the
  map-reduce system completely.
- Yokozuna feeds results into pipe/map-reduce one at a time.  It would
  be nice to have a batch insert to save a bit of work.
- Yokozuna doesn't return positions and score in the `KeyData` but
  instead passes a hard-coded empty list (`[]`).  Passing in positions
  is expensive and not guaranteed to be available for every field in
  Yokozuna.  I'm not sure if anyone was actually using this key data
  so I opted to hard-code it to an empty list for now.
- The `search_fold` iteration is hard-coded to batches of 10.  In the
  future this should be configurable and benchmarked.

### DIFF
--- a/include/yokozuna.hrl
+++ b/include/yokozuna.hrl
@@ -30,6 +30,7 @@
 -type proplist() :: proplists:proplist().
 -type ring() :: riak_core_ring:riak_core_ring().
 -type timestamp() :: erlang:timestamp().
+-type pipe() :: riak_pipe:pipe().
 
 -type index_set() :: ordset(index_name()).
 -type base64() :: binary().
--- a/rebar.config
+++ b/rebar.config
@@ -8,6 +8,8 @@
    {git, "git://github.com/cmullaparthi/ibrowse.git", {tag, "v3.0.4"}}},
   {riak_kv, ".*",
    {git, "git://github.com/basho/riak_kv", {branch, "yz-merge-1.3.0"}}},
+  {kvc, ".*",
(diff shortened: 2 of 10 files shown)


rzezeski:
This addresses issue #19.

PR #135, opened 2013-07-03
Add a KV Counter Extractor

lenary:
This adds a simple extractor that returns a pair of `{<<"riak_kv_counter">>, CounterValue}` to the yz backend.

It'll have to wait until work is rebased onto 1.4.0, as counters don't exist before that release.


src/yz_doc.erl
```diff
@@ -49,8 +49,9 @@ has_siblings(O) -> riak_object:value_count(O) > 1.
 %% @doc Given an object generate the doc to be indexed by Solr.
 -spec make_docs(obj(), binary(), binary(), boolean()) -> [doc()].
 make_docs(O, FPN, Partition, IndexContent) ->
-    [make_doc(O, Content, FPN, Partition, IndexContent)
-     || Content <- riak_object:get_contents(O)].
+    O2 = riak_kv_counter:merge(O),
```
lenary, line 52: A quick note: this function doesn't do anything if none of the siblings are counters, ie, it's safe. However, we require it if counters do exist because we want to get the most up to date version of those counters, which requires a merge.

lenary, line 52: In basho/riak_kv#594 I have updated this function so that it should be quicker if no siblings are counters. I guess it needs testing more, but this should help.

coderoshi, line 52: Merged basho/riak_kv#594 into rz-yz-merge-1.4.0 branch. Should be good to go in the near term

rzezeski:
I'm a bit cautious about brining in this merge function.  It adds a lists:any (versus list comprehension, which is faster) + try/catch + dict:fetch for every single write.  I'm going to tentatively mark this for 0.9.0 so we have time to discuss potentially better ways.  Or at least benchmark it to see what type of overhead this might have.

lenary:
@rzezeski a) this function _needs_ to be called so you can converge as many siblings as possible. b) this function _should_ be faster than the previous version, but [citation needed] as i am yet to microbenchmark it. I will do in the near future.

coderoshi:
**This PR is at a standstill, but be careful to avoid merging into _master_. Instead merge this to _develop_.**

rzezeski:
I moved this issue to the 1.1.0 release (end of October) as it isn't a MUST.  I had some discussions with @russelldb  in Boston about this and it sounded like he had a way to do this that only impacts the write path if necessary but involves making the extractor system smarter.  In any case, there are more fundamental things to get done.  I very much want counter support but it can wait.

lenary:
That's _totally_ cool with me. I think @russelldb and I are on the same page about it anyway.

lenary:
As the codepaths around CRDTs are changing for 2.0, @russelldb and I want to hold off on this extractor until after the original riak counter implementation is deprecated.

We will work on extractors for the full CRDTs implemented in 2.0, but not in time for 2.0

rzezeski:
I think that makes sense.  Going to close this.  Thanks for the update.

Issue #134, opened 2013-06-25
Don't throw on Solr start failure


rzezeski:
Currently the `yz_solr_proc` server will throw an exception if Solr fails to start in the allotted amount of time.  This will cause max restart frequency to be reached, rippling through the system and eventually bringing the node down.

This can be reproduced by binding to the port that the Solr process wants to use.

NOTE: Looking at this log more closely it actually might be a bad supervision tree in `riak_core`.  The `yz_solr_proc` server still shouldn't throw but might also want to investigate the core sup tree as well.

```
2013-06-25 18:13:40 =CRASH REPORT====
  crasher:
    initial call: yz_solr_proc:init/1
    pid: <0.1705.0>
    registered_name: yz_solr_proc
    exception exit: {{bad_return_value,{error,"Solr didn't start in alloted time"}},[{gen_server,init_it,6,[{file,"gen_server.erl"},{line,332}]},{proc_lib,init_p_do_apply,3,[{file,"proc_lib.erl"},{line,227}]}]}
    ancestors: [yz_solr_proc_sup,yz_sup,<0.669.0>]
    messages: [{#Port<0.10041>,{data,"Error: Exception thrown by the agent : java.rmi.server.ExportException: Port already in use: 10015; nested exception is: \n\tjava.net.BindException: Address already in use\n"}},{#Port<0.10041>,{exit_status,1}},{'EXIT',#Port<0.10041>,normal}]
    links: [<0.1704.0>]
    dictionary: []
    trap_exit: true
    status: running
    heap_size: 1597
    stack_size: 24
    reductions: 7949
  neighbours:
2013-06-25 18:13:40 =SUPERVISOR REPORT====
     Supervisor: {local,yz_solr_proc_sup}
     Context:    start_error
     Reason:     {bad_return_value,{error,"Solr didn't start in alloted time"}}
     Offender:   [{pid,undefined},{name,yz_solr_proc},{mfargs,{yz_solr_proc,start_link,["data/yz","10016","10015"]}},{restart_type,permanent},{shutdown,5000},{child_type,worker}]

2013-06-25 18:13:40 =SUPERVISOR REPORT====
     Supervisor: {local,yz_sup}
     Context:    start_error
     Reason:     shutdown
     Offender:   [{pid,undefined},{name,yz_solr_proc_sup},{mfargs,{yz_solr_proc_sup,start_link,[]}},{restart_type,permanent},{shutdown,infinity},{child_type,supervisor}]

[...]

PR #136, opened 2013-07-09
Fix various things in Solr startup

coderoshi:
This was originally written to address #134.  Mainly, it was to change `wait_for_solr` from throwing, causing the crash of the `yz_solr_proc:init` function and thus causing the whole node to stop.  It turns out that even without the throw, anything but `{ok, State}` will cause the supervisor tree to shutdown completely.  It's worse than that, the init code has a default of 5s to start according to @beerriot, so the `wait_for_solr` timeout isn't actually even honored.

The new plan is to cause the Riak node to have a full shutdown if the Solr process cannot stay up.  @beerriot will be making a PR to move the start out of the `init` function.  Until then various fixes have been made around the Solr startup process.  The main changes were to avoid throwing/exiting inside the process, instead use the `stop` directive for `gen_server`, and to move the solr proc directly under `yz_sup` in order to make sure max restart frequency can actually be reached and cause the Riak node to shutdown.


rzezeski:
This PR is meant to address #134.

src/yz_sup.erl
```diff
@@ -30,7 +30,15 @@
 %%%===================================================================
 
 start_link(Enabled) ->
-    supervisor:start_link({local, ?MODULE}, ?MODULE, [Enabled]).
+    process_flag(trap_exit, true),
```
rzezeski: This gives me an uneasy feeling.  This process flag is being applied to the caller of `start_link` which in this case happens to be Yokozuna's application master.  I'm not so sure we should be playing process flag games with a process we don't own.

Furthermore, I noticed this seems to cause issues when stoping.  I don't think it's caused by this PR.  Rather when calling stop the `yz_events` server gets a gen_event exit from riak_core_ring_events which it doesn't seem to have a handle_info for and thus causes a function clause error to get thrown.  This in turn trips the trap exit and the "Yokozuna had a problem starting..." message.

OH...wow, yea this is bad.  This call is never returning because of the receive call below.  It's blocking because this receive call is executing on application master process.  That process, in turn, doesn't wait for a message from the application it's starting, the app start is synchronous.  So the only thing that is going to trip this receive is a) a crash or b) some other message that might be sent to application master but that is OTP land and nothing we should touch.  Thus just dropping the message like we do below could have other unintended consequences.

rzezeski:
I'm -1 on this PR because of the process flag change in `yz_sup`.  This seems against the grain of OTP and I've already found it causes `start_link` to not return until a crash.  Perhaps a failing init is meant to ripple through the system and stop all applications?  The application startup process in Riak is still something that is a bit magic to me.  I don't know if it's a reltool thing, a supervisor thing, or what.  I'm starting to wonder if doing this work in the `init` function is the wrong thing to do in the first place.  I notice that if a Solr instance is already up you don't actually fail in init but instead get past init and get a `handle_info` call with an exit status of 1 from the port because it couldn't bind to the port.  It gets past init because there is already a Solr instance up that will give a 200 response to the `yz_solr:cores` call.  When this type of failure happens the worker is just restarted over and over and over like a typical OTP supervision tree.  This is "the erlang way" but it seems like keeping track of Solr restarts and shutting down only the Yokozuna application might make more sense.  I'm not sure.  One problem right now is the restart frequencies aren't set so well so in this case you get infinite restart attempts rather than bubbling up the supervisor chain and stopping whole node.

coderoshi:
According to the OTP supervisor init process, if any children fail on init, the whole supervisor is meant to fail. When the supervisor fails, the yokozuna application fails, and when that fails, the failure chains up to the whole node.

I chose to do the work in init, because the alternative is to startup the supervisor, then separately attach the children supervisors one by one, which was kind of a weird workflow I've never seen replicated in Riak.

I'd love to stop the launch of the actual yokozuna application, however, application:start/2 only specs two return values: {ok, Pid} and {error, Reason}. Sending error propagates failure to take down the whole node. But claiming the process started ok is also not exactly true. This was the crux of my unease with the whole idea of catching a failure on startup.

src/yz_solr_proc.erl
```diff
@@ -57,6 +57,11 @@
 start_link(Dir, SolrPort, SolrJMXPort) ->
     gen_server:start_link({local, ?MODULE}, ?MODULE, [Dir, SolrPort, SolrJMXPort], []).
 
+%% @doc Get the operating system's PID of the Solr/JVM process.  May
+%%      return `undefined' if Solr failed to start.
+-spec getpid() -> undefined | pos_integer().
+getpid() ->
```
coderoshi, line 63: Where is this used?

coderoshi, line 63: Nevermind, yz_monitor_solr

coderoshi:
+1

Issue #40, opened 2013-01-03
Make EC2 Deployment Easier


rzezeski:
A few of the steps for deploying EC2 nodes could be replaced with a shell script.
- Setting node name
- Setting cookie (prompt user)
- Listen on external interface for HTTP/PB (prompt user to verify this is okay)
- Create data dirs

rzezeski:
Given that Yokozuna will soon be part of Riak proper, which has its own EC2 support, there is no need to continue the custom EC2 support in the Yokozuna repo itself.

rzezeski:
I've added a PR to remove the EC2 scripts and doc: #140.

PR #140, opened 2013-07-22
Remove EC2 scripts and docs

rzezeski:
Yokozuna will soon be part of Riak proper which has its own EC2
support.  Not only that but the base image for this EC2 script is well
out of date and the AMI hasn't been updated for several releases.

This also means issue #40 is obsolete.


coderoshi:
+1

PR #147, opened 2013-07-24
Add ability to enabled/disable components

rzezeski:
This is actually a rename of the `noop_flag` API.
- Add a `component()` type that enumerates the components that may be
  enabled or disabled.
- Add functions to the `yokozuna` module to enabled/disable components
  and check if a component is enabled.
- By default a component is considered enabled.
- Honor the search enabled flag in the protobuff search path.
- Add better code documentation and specs.


include/yokozuna.hrl
```diff
@@ -77,6 +77,23 @@
                    | {siblings, binary()}
                    | {'query', binary()}.
 
+%% @doc The `component()' type reprsents components that may be
```
coderoshi: represents

src/yz_kv.erl
```diff
@@ -110,7 +110,7 @@ get_md_entry(MD, Key) ->
     yz_misc:dict_get(Key, MD, none).
 
 index(Obj, Reason, VNodeState) ->
-    case yokozuna:noop_flag(index) orelse not ?YZ_ENABLED of
+    case yokozuna:is_enabled(index) orelse not ?YZ_ENABLED of
```
coderoshi: this should check if "not enabled" (=noop), or better yet:

case yokozuna:is_enabled(index) andalso ?YZ_ENABLED of
  false -> ok;
  true -> %% do work...

rzezeski:
How about now @coderoshi?

coderoshi:
+1, though https://github.com/basho/yokozuna/pull/147/files#L2R113 could be clarified

PR #148, opened 2013-07-24
set java.awt.headless to prevent Mac dock icons

beerriot:
Java 1.6 on Mac will add an icon to the dock and app switcher for each
JVM running. When the VM starts, this "app" comes to the foreground and
steals focus.

Setting "-Djava.awt.headless=true" runs the VM in headless mode, so it
does not connect to the windowing server, and thus does not create these
artifacts.


rzezeski:
+1 to merge

PR #149, opened 2013-07-25
return ordset from flagged_buckets, because yz_misc:delta expects it

beerriot:
The list produced by riak_core_bucket:get_buckets/1 is not guaranteed to be
in order by bucket name, so we need to turn it into an ordset before sending
it to yz_misc:delta, which expects an ordset.


rzezeski:
+1 to merge.

rzezeski:
+1 to merge.

PR #151, opened 2013-07-25
fix index/extractor ring modification timestamp conflicts

beerriot:
Before this change, if two index or extractor-map changes happened within
one second, one of the modifications may be lost. This is due to the
1-second resolution in the ring's meta_entry records.

This commit solves the issue in two ways:
1. Modifications are made inside the ring_trans call, to the ring passed
   into the function. The had been made in an always-clobber manner before.
2. The function sleeps for one second to ensure that the update will have
   a newer timestamp.

The second point is not the greatest because it blocks other calls to
the riak_core_ring_manager process for that second. Reads are not blocked,
since they look at an ETS table instead. An intermediate serialization+wait
process could be used instead if the temporary block causes major problems.

### DIFF
--- a/src/yz_extractor.erl
+++ b/src/yz_extractor.erl
@@ -120,20 +120,26 @@ register(MimeType, Def) ->
 -spec register(mime_type(), extractor_def(), register_opts()) ->
                       extractor_map() | already_registered.
 register(MimeType, Def, Opts) ->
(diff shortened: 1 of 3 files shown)


rzezeski:
Reviewed the code and ran `yokozuna_essential` 3 times.  All passed.  +1 to merge.

PR #153, opened 2013-08-06
Fix field name creation in XML extractor

rzezeski:
Fix the field name creation when there are an even number of elements
in the name stack.

e.g.

```
<one>
  <two>
    <three>
      <four>
      </four>
    </three>
  </two>
</one>
```

Should generate `one_two_three_four` not `one_twothree_four` as it did
up until this patch.


coderoshi:
+1 short and sweet

PR #144, opened 2013-08-07
Add Administration Support to PB

coderoshi:
Add schema and index administration to the protobuff interface so that the official clients will work.

This relies on basho/riak_pb#51

After basho/riak-erlang-client#112 is merged, I'll update this PR to use the new client.

### DIFF
--- a/src/yz_pb_admin.erl
+++ b/src/yz_pb_admin.erl
@@ -1,0 +1,142 @@
+%% -------------------------------------------------------------------
+%%
+%% yz_pb_search: PB Service for Yokozuna queries
+%%
+%% Copyright (c) 2013 Basho Technologies, Inc.  All Rights Reserved.
+%%
+%% This file is provided to you under the Apache License,
+%% Version 2.0 (the "License"); you may not use this file
+%% except in compliance with the License.  You may obtain
+%% a copy of the License at
+%%
+%%   http://www.apache.org/licenses/LICENSE-2.0
+%%
+%% Unless required by applicable law or agreed to in writing,
+%% software distributed under the License is distributed on an
+%% "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+%% KIND, either express or implied.  See the License for the
+%% specific language governing permissions and limitations
+%% under the License.
+%%
+%% -------------------------------------------------------------------
+
+%% @doc Implements a `riak_api_pb_service' for performing
+%% administrative functions in Yokozuna, like managing
+%% indexes or schema
+-module(yz_pb_admin).
+
+-include_lib("riak_pb/include/riak_kv_pb.hrl").
+-include_lib("riak_pb/include/riak_yokozuna_pb.hrl").
+-include("yokozuna.hrl").
+
+-behaviour(riak_api_pb_service).
+
+-export([init/0,
+         decode/2,
+         encode/1,
+         process/2,
+         process_stream/3]).
+-compile(export_all).
+
+%% @doc init/0 callback. Returns the service internal start state.
+-spec init() -> any().
+init() ->
+    no_state.
+
+%% @doc decode/2 callback. Decodes an incoming message.
+decode(Code, Bin) ->
+    {ok, riak_pb_codec:decode(Code, Bin)}.
+
+%% @doc encode/1 callback. Encodes an outgoing response message.
+encode(Message) ->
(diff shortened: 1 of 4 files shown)


src/yz_pb_admin.erl
```diff
@@ -0,0 +1,142 @@
+%% -------------------------------------------------------------------
+%%
+%% yz_pb_search: PB Service for Yokozuna queries
```
rzezeski: Copy-pasta.  Also, let's not bother repeating module names.  It's already in the filename and -module.

src/yz_pb_admin.erl
```diff
@@ -0,0 +1,142 @@
...
+
+%% @doc encode/1 callback. Encodes an outgoing response message.
+encode(Message) ->
+    {ok, riak_pb_codec:encode(Message)}.
+
+%% @doc process/2 callback. Handles an incoming request message.
+process(#rpbyokozunaindexdeletereq{name = IndexNameBin}, State) ->
+    Ring = yz_misc:get_ring(transformed),
+    IndexName = binary_to_list(IndexNameBin),
+    case yz_index:exists(IndexName) of
```
rzezeski, line 74: Don't need to change for this PR but just want to note that this will return `false` if the index is simply down temporarily.  Thus a cilent could call delete and get `notfound` when in reality the index exists, it just isn't up ATM.

src/yz_pb_admin.erl
```diff
@@ -0,0 +1,142 @@
...
+                Buckets ->
+                    Msg = "Can't delete index with associate buckets ~p",
+                    Msg2 = lists:flatten(io_lib:fwrite(Msg, [Buckets])),
+                    {error, Msg2, State}
+            end;
+        false ->
+            {error, "notfound", State}
+    end;
+
+process(#rpbyokozunaindexputreq{
```
rzezeski, line 89: Nitpick...what about calling it create?  Put seems like a holdover from HTTP.

src/yz_pb_admin.erl
```diff
@@ -0,0 +1,142 @@
...
+        {error, {rpc_fail, Claimant, _}} ->
+            Msg = "Cannot create index while claimant node "++Claimant++" is down~n",
+            {error, Msg, State};
+        {error, schema_not_found} ->
+            {error, "Schema not found", State}
+    end;
+
+process(#rpbyokozunaindexgetreq{name = IndexName}, State) ->
+    Ring = yz_misc:get_ring(transformed),
+    lager:warning("GET REQ IDX ~p~n", [IndexName]),
```
rzezeski: That little guy?  I wouldn't worry about that little guy.

src/yz_pb_admin.erl
```diff
@@ -0,0 +1,142 @@
...
+        ok ->
+            {reply, #rpbputresp{}, State};
+        {error, {rpc_fail, Claimant, _}} ->
+            Msg = "Cannot create index while claimant node "++Claimant++" is down~n",
+            {error, Msg, State};
+        {error, schema_not_found} ->
+            {error, "Schema not found", State}
+    end;
+
+process(#rpbyokozunaindexgetreq{name = IndexName}, State) ->
```
rzezeski: Nitpick...should we make get and list separate in PB?  What will it look like to clients to use get to list the indexes?  In Erlang we could use get/0 which passes sentinel value to get/1, but will this play well with all clients?  Perhaps an actual `list` API makes sense here.

rzezeski: This makes me wonder what the client API will look like.  For example, getting an index.  Should that be `riakc_pb_socket:get_index/1`?  But Yokozuna isn't the only thing with the notion of an index.  Will users become confused and think this is related to 2i?  Should the client APIs be namespaced?

coderoshi: This is just the message passing portion. There's no reason that client implementation can't wrap this message into a list() client function.

src/yz_pb_admin.erl
coderoshi: lol I've deleted it locally, I just need to get this in

src/yz_pb_admin.erl
coderoshi, line 89: I thought about changing it, but really this creates or updates. I could make two identical messages, but all of the existing PBs use "put".

coderoshi:
Closing this PR to point at _develop_ branch

PR #161, opened 2013-08-27
Add Administration Support to PB

coderoshi:
Pointing the former PR #144 to _develop_ branch.

In order to test you'll need basho_bench to find the newest riak-erlang-client. Update the `misc/bench/deps/basho_bench/rebar.config` to use riak-erlang-client's {branch, "master"}


coderoshi:
There's a new branch in basho_bench that points to the newer riakc. This branch should now work out of the box.

riak_test/yz_pb.erl
```diff
@@ -58,47 +75,86 @@ http(Method, URL, Headers, Body) ->
...
+    end ),
+    riakc_pb_socket:stop(Pid),
+    ok.
+
+confirm_admin_index(Cluster) ->
+    Index = <<"index">>,
+    create_index(Cluster, Index, Index),
+    Node = select_random(Cluster),
+    [{Host, Port}] = host_entries(rt:connection_info([Node])),
+    {ok, Pid} = riakc_pb_socket:start_link(Host, (Port-1)),
```
rzezeski, line 141: There are functions to get the PB port from the connection info but we can use those next time around.

src/yz_app.erl
```diff
@@ -26,7 +26,8 @@
 
 %% 27 is message type rpbsearchqueryreq
 %% 28 is message type rpbsearchqueryresp
--define(SERVICES, [{yz_pb_search, 27, 28}]).
+-define(QUERY_SERVICES, [{yz_pb_search, 27, 28}]).
+-define(ADMIN_SERVICES, [{yz_pb_admin, 54, 60}]).
```
rzezeski, line 30: We will want to add the `ADMIN_SERVICES` range to the `yokozuna:swtich_to_yokozuna/0` function.

src/yz_pb_admin.erl
```diff
@@ -0,0 +1,153 @@
...
+    case yz_schema:store(SchemaName, Content) of
+        ok  ->
+            {reply, #rpbputresp{}, State};
+        {error, Reason} ->
+            Msg = io_lib:format("Error storing schema ~p~n", [Reason]),
+            {error, Msg, State}
+    end;
+
+process(#rpbyokozunaschemagetreq{name = SchemaName}, State) ->
+    {ok, Content} = yz_schema:get(SchemaName),
```
rzezeski, line 67: We don't have to change this now but probably don't want to assume `{ok,Content}` as this will crash the PB pid.

src/yz_pb_admin.erl
```diff
@@ -0,0 +1,153 @@
...
+                [] ->
+                    ok = yz_index:remove(IndexName),
+                    {reply, rpbdelresp, State};
+                Buckets ->
+                    Msg = "Can't delete index with associate buckets ~p",
+                    Msg2 = lists:flatten(io_lib:fwrite(Msg, [Buckets])),
+                    {error, Msg2, State}
+            end;
+        false ->
+            {error, "notfound", State}
```
rzezeski, line 86: This isn't consistent with the HTTP delete which returns success whether existing or not.  Don't want to do anything about it now just making a note.

rzezeski:
+1 to merge

PR #162, opened 2013-08-29
Simlify grab-solr.sh

rzezeski:
Yokozuna always pulls a pre-built Solr package now.  Pull out the
build-from-source stuff.


coderoshi:
+1

PR #164, opened 2013-09-04
Use application env for component API

rzezeski:
It turns out that mochiglobal is a really bad idea if you often read
keys that are undefined.  This is because `mochiglobal:get` uses
try/catch to check for undefined and that becomes really expensive
when every call causes an `undef` error.

In the cases of Yokozuna, even when disabled, the KV vnode is making a
call to the `yz_kv:index` function which is doing a get for an
undefined key.  On my development machine I saw an order of magnitude
drop from ~1600 ops/s to ~150 ops/s.

This patch uses application env instead which brings the write
throughput back up to speed.

### DIFF
--- a/src/yokozuna.erl
+++ b/src/yokozuna.erl
@@ -34,21 +34,21 @@
 %%      investigating latency issues.
 -spec disable(component()) -> ok.
 disable(Component) ->
-    mochiglobal:put(Component, false).
+    application:set_env(?YZ_APP_NAME, {component, Component}, false).
 
 %% @doc Enabled the given `Component'.  This only needs to be called
 %%      if a component was previously disabled as all components are
 %%      considered enabled by default.
 -spec enable(component()) -> ok.
 enable(Component) ->
-    mochiglobal:put(Component, true).
+    application:set_env(?YZ_APP_NAME, {component, Component}, true).
 
 %% @doc Determine if the given `Component' is enabled or not.  If a
 %%      component is not explicitly disabled then it is considered
 %%      enabled.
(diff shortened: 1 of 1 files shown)


rzezeski:
Following are a set of benchmarks proving the existence of the throughput loss and validation that the fix works.  All runs below used my "fruit benchmark" of loading 1 million small 'text/plain' objects.

### Riak 1.4.0 Base (i.e. NO Yokozuna Code)

riak-base-140

### Riak/Yokozuna 0.8.0 - Yokozuna Disabled

riak-yz-080-disabled

### Riak/Yokozuna + This Patch - Yokozuna Disabled

riak-yz-latest-disabled

coderoshi:
+1

rzezeski:
@javajolt `application:get_env` uses ETS under the covers.  There can very well be contention issues with ETS tables.  I've seen it in the wild on 16-core machines, but ETS concurrency gets better every release.  In fact if you look closely the graph for this patch is just ever so slightly slower than baseline.  This may be in part thanks to the ETS usage.  It could also be the case statement + `andalso` in `yz_kv:index` since `yokozuna:is_enabled(index)` defaults to true.  Either way this should almost never be the dominating factor during writes.

That said, this API is also used during **every** query and I am slightly concerned it could cause contention in the case where the query results are cached by Solr and thus they can complete fast enough so that Yokozuna becomes the bottleneck.  I plan to benchmark this and have an idea for a solution that involves `sidejob_config` which is like `mochiglobal` but without the try/catch.  Anyways, for the purposes of this PR I'm going to punt on the query benchmark since this is already an obvious improvement.

PR #167, opened 2013-09-07
Use periodic sync for hashtree updates

rzezeski:
While running the yokozuna_essential test I noticed vnodes crashing
because of unexpected `{ref(),ok}` messages being handled by
`riak_kv_vnode:handle_info`.  The yokozuna index function,
`yz_kv:index` runs on the KV vnode process.  Part of that function is
to update the Yokozuna hashtree.  It does that by making a
gen_server:call with a timeout of 1 second.  If this call times out
then the reply ends up being handled by the KV vnode `handle_info`
callback which has no function head for that reply.  Thus crashing the
vnode and causing pain.

This patch updates Yokozuna's hashtree update to be like KV.  It uses
mostly async updates followed by the occasional blocking call with
`infinity` timeout.  This a) reduces overhead of most updates and b)
prevents chance of an unexpected msg making it to the KV vnode
callback.

I also removed the try/catch around the hashtree update since there is
already one in the `index` function covering this call.

### DIFF
--- a/src/yz_kv.erl
+++ b/src/yz_kv.erl
@@ -194,25 +194,69 @@ update_aae_tree_stats(Index, BuildTime) ->
     riak_kv_entropy_info:tree_built(yz, Index, BuildTime),
     ok.
 
+%% @private
+%%
+%% @doc Update the hashtree for `Parition'/`IdxN'.
+%%
+%% `Action' - Either delete the `BKey' by passing `delete' or insert
+%% hash by passing `{insert, ObjHash}'.
+%%
+%% `Partition' - The partition number of the tree to update.
+%%
+%% `IdxN' - The preflist the `BKey' belongs to.
+%%
+%% `BKey' - The bucket/key encoded as binary.
 -spec update_hashtree(delete | {insert, binary()}, p(), {p(),n()},
                       {binary(), binary()}) -> ok.
 update_hashtree(Action, Partition, IdxN, BKey) ->
-    try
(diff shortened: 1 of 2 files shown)


rzezeski:
5 runs of `yokozuna_essential` without vnode failures.

src/yz_kv.erl
```diff
@@ -194,26 +194,70 @@ update_aae_tree_stats(Index, BuildTime) ->
...
     end.
 
+%% @private
+%%
+%% @doc Return the max number of async hashtree calls that may be
+%% performed before requiring a blocking call.
+-spec max_hashtree_tokens() -> pos_integer().
+max_hashtree_tokens() ->
+    %% Use same max as riak_kv
+    app_helper:get_env(riak_kv, anti_entropy_max_async, 90).
```
coderoshi, line 259: Doesn't yz using the aae max calls effectively cut the available kv aae calls in half? Could this cause other problems, or should we suggest a higher default for anti_entropy_max_async if you're using yz?

rzezeski, line 259: No, this is just getting a value from the application env.  The number is then inserted into the process dictionary.  KV uses `hashtree_tokens` and Yokozuna uses `yz_hashtree_tokens` to prevent decrementing the same pdict key.  See the `get_method` function.

coderoshi:
+1 works for me as well. Migrations timed out, but I'm not convinced that's related.

PR #169, opened 2013-09-08
SpatialSearch fix in the default-schema.xml

vincenzovitale:
set multiValued for the dynamic field *_p
added the dynamic field *_coordinate

### DIFF
--- a/priv/default_schema.xml
+++ b/priv/default_schema.xml
@@ -93,7 +93,8 @@
    <dynamicField name="*_bs" type="boolean" indexed="true" stored="true"  multiValued="true"/>
    <dynamicField name="*_dt"  type="date"    indexed="true"  stored="true"/>
    <dynamicField name="*_dts" type="date"    indexed="true"  stored="true" multiValued="true"/>
-   <dynamicField name="*_p"  type="location" indexed="true" stored="true"/>
+   <dynamicField name="*_coordinate"  type="tdouble" indexed="true"  stored="false" multiValued="false"/>
+   <dynamicField name="*_p" type="location" indexed="true" stored="true" multiValued="false"/>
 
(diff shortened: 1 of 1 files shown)


rzezeski:
Vincenzo,

Thanks for the report.

I was unable to reproduce.  Did you by chance use a different Solr package than the one that Yokozuna downloads?  It looks like Lucene added multi-valued field cache support in 4.0.

https://issues.apache.org/jira/browse/LUCENE-3354

In any case the *_coordinate field should be added.  Otherwise the field is being indexed under the catch-all field which is of type text.  The query returns the correct answer for me but should make sure that it uses the double type.

FWIW, Solr 4 also added support for the new Lucene spatial support.

http://wiki.apache.org/solr/SolrAdaptersForLuceneSpatial4

vincenzovitale:
Hi Ryan,

I built the riak develop branch since I saw your last commit adding yokozuna there. Looking at what I got in my deps, the yokozuna included is the develop branch at fc8f497c52e234d57bd7f3f45452089d4e5af698 (Update verify.sh to use develop branches). So it seems I'm missing a few commits from the yz develop branch, but indeed the solr version used is https://s3.amazonaws.com/yzami/pkgs/solr-4.3.0-yz.tgz so it should be fixed.

You didn't get any problem without adding the *_coordinate dynamic field (even without changing the multiValue attr)?

rzezeski:
That's correct, I didn't run into any trouble.  The `*_coordinate` field will get picked up by the `*` dynamic field.  And as of Lucene/Solr4 there is multi-valued support for spatial.  That said, I think this change makes sense for the default schema.  I'm going to merge it.

Thanks for taking the time to report this and submit a pull request.

vincenzovitale:
Thanks Ryan

Issue #143, opened 2013-07-23
Remove the Default Index


rzezeski:
Yokozuna currently indexes all data in all buckets even if not marked for indexing.  This is done so that AAE will work.  Find another way to keep AAE from doing superfluous repairs without requiring the default index.  Some ideas:
- Log file of non-indexed bucket hashes.
- Do both KV and Solr fold when building Yokozuna trees.

beerriot:
Closing this, because default index is gone. We took the second idea option.

PR #171, opened 2013-09-12
Remove the default index

rzezeski:
The default index, `_yz_default`, was added to prevent AAE from doing
unnecessary repairs after tree re-build.  After a Yokozuna hashtree is
cleared it is rebuilt from data in Solr.  If Solr only contains data
for indexed buckets then its hashtree will be missing hashes for
non-indexed buckets.  Since AAE exchanges based on partition the
system will think keys are missing and potentially invoke lots of
repair.  The default index avoids this by indexing the needed data for
AAE for buckets which are not indexed.  AAE works, but at the cost of
affecting write throughput on non-indexed buckets and increasing the
index size.  On my local workstation the default index reduced write
throughput to 56% of baseline.

This tradeoff is unnecessary.  Write throughput on a non-indexed
bucket can remain much closer to baseline without causing trouble for
AAE.
- Determine if the object needs to be indexed.
- If it does then write to Solr and update the hashtree.  If it
  doesn't then just update the hashtree.  This ensures that the
  hashtrees are kept up to date with significantly less cost than
  sending an update to Solr
- After a Yokozuna hashtree is cleared it will be missing keys stored
  in non-indexed buckets.
- During exchange each missing key will cause the repair function to
  be invoked.
- This repair function will now check if the object needs to be
  indexed.  I.e. an actual missing index entry versus an object that
  doesn't need to be indexed.  If the object is not to be indexed then
  the Yokozuna hashtree is simply sent an update.
- So the first exchange of a tree after re-build will invoke the
  repair function potentially many times.  But these invocations will
  consist of a simple check + a hashtree update.  This is far less
  work than minimally indexing all data.

### DIFF
--- a/riak_test/yz_flag_transitions.erl
+++ b/riak_test/yz_flag_transitions.erl
@@ -48,9 +48,9 @@ confirm() ->
     verify_many_to_one_index_remove(Cluster),
     pass.
 
-%% @doc When an index is associated the indexes for the bucket should
-%%      be removed from the default index and AAE should re-index
-%%      objects under the bucket's index.
+%% @doc When an index is associated with a bucket the Yokozuna AAE
+%% trees should be cleared.  Thus on next exchange missing indexes
+%% will be discovered and repaired.
 verify_index_add(Cluster, YZBenchDir) ->
     lager:info("Verify adding index"),
     yz_rt:load_data(Cluster, "fruit", YZBenchDir, ?NUM_KEYS),
@@ -60,7 +60,6 @@ verify_index_add(Cluster, YZBenchDir) ->
     HP = yz_rt:select_random(Hosts),
     lager:info("Verify fruit index doesn't exist"),
     {ok, "404", _, _} = yz_rt:search(yokozuna, HP, "fruit", "*", "*"),
-    lager:info("Verify objects are indexed under default index"),
     lager:info("Create fruit index + set flag"),
     yz_rt:create_index(yz_rt:select_random(Cluster), "fruit"),
     yz_rt:set_index(yz_rt:select_random(Cluster), <<"fruit">>),
@@ -70,20 +69,18 @@ verify_index_add(Cluster, YZBenchDir) ->
     F = fun(Node) ->
                 lager:info("Verify that AAE re-indexes objects under fruit index [~p]", [Node]),
                 HP2 = hd(yz_rt:host_entries(rt:connection_info([Node]))),
-                %% Also verify that the data was deleted under default index
                 yz_rt:search_expect(HP2, "fruit", "*", "*", ?NUM_KEYS)
         end,
     yz_rt:wait_until(Cluster, F).
 
 %% @doc When an index is dissociated the indexes for that bucket's
-%%      index should be deleted and AAE should re-index objects under
-%%      the default index.
+%%      index should be deleted.
 verify_index_remove(Cluster) ->
     lager:info("Verify removing index"),
     Node = yz_rt:select_random(Cluster),
     yz_rt:remove_index(Node, <<"fruit">>),
     F = fun(Node2) ->
-                lager:info("Verify fruit indexes are deleted + objects re-indexed under default index [~p]", [Node2]),
+                lager:info("Verify fruit indexes are deleted [~p]", [Node2]),
                 HP = hd(yz_rt:host_entries(rt:connection_info([Node2]))),
                 yz_rt:search_expect(HP, "fruit", "*", "*", 0)
         end,
--- a/riak_test/yz_flag_transitions.erl
+++ b/riak_test/yz_flag_transitions.erl
@@ -1,16 +1,14 @@
 %% @doc Test the addition or removal of the index entry from a bucket
 %%      that has data.  If a bucket already has data then associating
-%%      it with an index should cause the deletion of all entries for
-%%      this bucket in the default index and re-indexing under the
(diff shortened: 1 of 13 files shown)


rzezeski:
```
+----------+-------------------------+--------------------+
|setup     |mean throughput          |median latency      |
+----------+-------------------------+--------------------+
|0.9.0 off |1577 writes/s            |20.6 ms             |
+----------+-------------------------+--------------------+
|0.9.0 on  |902 writes/s (57%)       |35.5 ms (1.7x)      |
+----------+-------------------------+--------------------+
|dev off   |1576 writes/s            |20.6 ms             |
+----------+-------------------------+--------------------+
|dev on    |1461 writes/s (93%)      |22.3 ms (1.08x)     |
+----------+-------------------------+--------------------+
```

The table above gives a summary of the benchmark results on my workstation.  The `setup` column described the Yokozuna version and whether or not Yokozuna is enabled.  E.g. `0.9.0 off` means Yokozuna 0.9.0 with Yokozuna disabled, `on` means Yokozuna is enabled but **NO** index is assocaited with the bucket.  The throughput are the writes per second as determined by basho bench.  The percentage represents the relative percentage of the previous run.  E.g. for 0.9.0 enabling Yokozuna causes the throughput to drop to 56% of baseline.  I ran a base for each version, 0.9.0 and development, to verify that the baselines are the same for each version and that some other effect hasn't helped the throughput.  Following are the basho bench graphs of each run.

### 0.9.0 off

load-fruit-090-yz-off

### 0.9.0 on

load-fruit-090-yz-on-non-indexed

### development off

load-fruit-dev-yz-off

### development on

load-fruit-dev-yz-on

rzezeski:
This addresses issue #143.

rzezeski:
Just realized that the repairs after tree rebuild will count as actual index repairs.  Need to change `repair` to return atom indicating if real repair and only count those towards total.

riak_test/yz_flag_transitions.erl
```diff
@@ -63,7 +61,6 @@ verify_index_add(Cluster, YZBenchDir) ->
     lager:info("Verify fruit index doesn't exist"),
     {ok, "404", _, _} = yz_rt:search(yokozuna, HP, "fruit", "*", "*"),
     lager:info("Verify objects are indexed under default index"),
```
beerriot: This log line should probably be deleted as well.

src/yz_exchange_fsm.erl
```diff
@@ -197,26 +197,49 @@ exchange_segment_kv(Tree, IndexN, Segment) ->
 -spec repair(p(), keydiff()) -> ok.
 repair(Partition, {remote_missing, KeyBin}) ->
     %% Yokozuna has it but KV doesn't
+    Ring = yz_misc:get_ring(transformed),
     BKey = binary_to_term(KeyBin),
+    Index = yz_kv:get_index(BKey, Ring),
+    ShortPL = yz_kv:get_short_preflist(BKey, Ring),
     FakeObj = fake_kv_object(BKey),
-    yz_kv:index(FakeObj, delete, Partition),
+    %% Repeat sopme logic in `yz_kv:index/3' to avoid extra work.  Can
```
beerriot: "sopme" == "some"?

beerriot:
verify.sh results:

```
yz_wm_extract_test-bitcask : pass
yz_siblings-bitcask        : fail
yz_schema_admin-bitcask    : pass
yz_rs_migration-bitcask    : fail
yz_pb-bitcask              : pass
yz_monitor_solr-bitcask    : pass
yz_mapreduce-bitcask       : pass
yz_languages-bitcask       : pass
yz_index_admin-bitcask     : pass
yz_flag_transitions-bitcask: pass
yz_fallback-bitcask        : pass
yz_errors-bitcask          : pass
yokozuna_essential-bitcask : pass
```

Full log (except for build stage) here: https://gist.github.com/beerriot/3924dcbddb94bf4d111d

The migration test failed because I don't have an earlier version of riak set up for riak_test to upgrade from. I have not yet dug into the siblings test failure.

beerriot:
This PR should probably also link to an associated PR in basho/riak for removing the `data/_yz_default` creation in `reltool.conf`.

riak_test/yz_flag_transitions.erl
rzezeski: Good call, turns out I missed several references to the default index.

src/yz_exchange_fsm.erl
rzezeski: Fixed

beerriot:
I've been re-running the `yz_siblings` test all afternoon, and I haven't seen it fail. Given that I also don't see any way that this patch set would have changed the way siblings are handled, I vote that we ignore the failure I saw in my initial run, and open a separate ticket to track this inquiry.

beerriot:
I've just verified that I see the same performance improvement that Ryan saw. That is, before this PR, write throughput for non-indexed data drops by 50% when you turn yokozuna on. After this PR, the drop is less than 10% (noisy local nodes prevent me from being more precise).

rzezeski:
I finally pushed my fix for indefinite repair along with a test for it.  Unless you find other problems that should be all of the commits I plan to make.  I still need to make a Riak patch.

beerriot:
All tests pass after the small correction I added (though I did have a timeout in one run of `yokozuna_essential`). We've discussed some other improvements (performance and otherwise), but I think they can all be done as separate PRs.

+1 to merge this

PR #172, opened 2013-09-18
Small dialyzer fixes

rzezeski:
NOTE: This requires 16B01 or else dialyzer will fail.

I figured I'd could make some small fixes while incapacitated.  Just some small fixes to make dialyzer happier.

```
yokozuna.erl:37: The call application:set_env('yokozuna',{'component',_},'false') breaks the contract (Application,Par,Val) -> 'ok' when is_subtype(Application,atom()), is_subtype(Par,atom()), is_subtype(Val,term())
yokozuna.erl:44: The call application:set_env('yokozuna',{'component',_},'true') breaks the contract (Application,Par,Val) -> 'ok' when is_subtype(Application,atom()), is_subtype(Par,atom()), is_subtype(Val,term())
yokozuna.erl:51: The call app_helper:get_env('yokozuna',{'component',_},'true') will never return since it differs in the 2nd argument from the success typing arguments: (atom(),atom(),any())
yz_kv.erl:71: The call lists:sort(Vclock::vclock:vclock()) contains an opaque term as 1st argument when a structured term of type [any()] is expected
yz_kv.erl:97: Invalid type specification for function yz_kv:get_tree_build_time/1. The success typing is (hashtree:hashtree()) -> any()
yz_pb_admin.erl:32: Callback info about the riak_api_pb_service behaviour is not available
yz_pb_search.erl:30: Callback info about the riak_api_pb_service behaviour is not available
```

Note that these errors still exist but not sure much can be done about them.  I was worried about the first 3.  They are correct according to type specs but I reviewed the application code.  It technically can accept `term()` as key so I'm going to let Yokozuna exploit that for a little while longer.

### DIFF
--- a/include/yokozuna.hrl
+++ b/include/yokozuna.hrl
@@ -69,7 +69,7 @@
 -type p_node() :: {p(), node()}.
 -type lp_node() :: {lp(), node()}.
 -type cover_set() :: [p_node()].
--type logical_cover_set() :: [lp_node()].
+-type logical_cover_set() :: [{lp_node(), filter()}].
 -type filter_cover_set() :: [{p_node(), filter()}].
 
 -type ring_event() :: {ring_event, riak_core_ring:riak_core_ring()}.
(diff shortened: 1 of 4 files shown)


coderoshi:
+1 works as advertised

Issue #165, opened 2013-09-06
Setup Data Directory At Startup, Not At Install


beerriot:
Yokozuna currently expects `solr.xml` and several files in a `_yz_conf` directory to be present in the data directory at startup. It expects that the files will have been put in place at install (because reltool copies them during release generation).

This causes maintenance hassle and fragility for two reasons:
1. Changing where yokozuna data is stored requires _both_ editing app.config, and also moving these files.
2. Cleaning all data out of a Riak node is not as simple as deleting everything in the data directory. (Using riak_test's `rt:clean_cluster/1` will cause Yokozuna to fail at restart.)

Instead of expecting the files to exist, Yokozuna should check if they exist, and create them if they do not.

beerriot:
The removal of the default index in #143 will address a portion of this ticket.

PR #173, opened 2013-09-19
ensure data dir and solr.xml exist at startup

beerriot:
Instead of relying on installation to setup the data directory, set it up at startup. This addresses #165.

This PR moves the template `solr.xml` into `priv/template_solr.xml`, so that it will be packaged with the build. I named it `template_…` because I find that less confusing than seeing multiple `solr.xml` files scattered around.

I've also removed the default index definition from the solr.xml template.


coderoshi:
+1 the kinder, gentler solr config

PR #174, opened 2013-09-19
Translate bucket, or expect index name, in mapred_search

beerriot:
This PR changes mapred_search to lookup the index name associated with a bucket, instead of passing the bucket name to solr. The bucket name is also passed in a filter query, such that only keys from the bucket originally specified are returned from the index.

This PR also allows the specification of an index, instead of a bucket, which will produce all keys matching the query, regardless of bucket. To trigger this behavior, pass `{index, Index}` or `{struct, [{<<"index">>}, Index}]}` (mochijson2 format) instead of `Bucket` as the `Group` parameter to `mapred_search`.

### DIFF
--- a/src/yokozuna.erl
+++ b/src/yokozuna.erl
@@ -55,7 +55,11 @@ is_enabled(Component) ->
 %% `Pipe' - Reference to the Riak Pipe job responsible for running the
 %%   map/reduce job.
 %%
-%% `Index' - The name of the index to run the search against.
+%% `Group' - The name of the bucket or index to run the search
+%%   against. The name of a bucket should be given as a string (list
+%%   or binary). The name of an index should be given as one of two
+%%   tuples: `{index, Index}' or `{struct,[{<<"index">>, Index}]}'
+%%   (erlang-y or mochijson2-ish).
 %%
 %% `Query' - The query to run.
 %%
@@ -77,16 +81,13 @@ is_enabled(Component) ->
 %%   timeout value should be revisited in the future and see if it
 %%   can't be removed from the API completely.
 -spec mapred_search(pipe(), [term()], pos_integer()) -> ok.
-mapred_search(Pipe, [Index, Query], _Timeout) ->
-    mapred_search(Pipe, [Index, Query, <<"">>], _Timeout);
-mapred_search(Pipe, [Index, Query, Filter], _Timeout) ->
-    Index2 = case is_binary(Index) of
-                 false -> Index;
-                 true -> binary_to_list(Index)
-             end,
+mapred_search(Pipe, [Group, Query], _Timeout) ->
+    mapred_search(Pipe, [Group, Query, <<"">>], _Timeout);
+mapred_search(Pipe, [Group, Query, Filter], _Timeout) ->
+    Index = index_for_group(Group),
     %% Function to convert `search_fold' results into pipe format.
-    Q = fun({Bucket,Key,Props}) ->
-                riak_pipe:queue_work(Pipe, {{Bucket, Key}, {struct, Props}})
+    Q = fun({Index,Key,Props}) ->
+                riak_pipe:queue_work(Pipe, {{Index, Key}, {struct, Props}})
         end,
     F = fun(Results, Acc) ->
                 %% TODO: is there a way to send a batch of results?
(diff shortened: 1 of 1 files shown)


src/yokozuna.erl
```diff
@@ -94,9 +96,33 @@ mapred_search(Pipe, [Index, Query, Filter], _Timeout) ->
...
         end,
-    ok = search_fold(Index2, Query, Filter, F, ok),
+    ok = search_fold(Index, Query, WholeFilter, F, ok),
     riak_pipe:eoi(Pipe).
 
+index_for_group({index, Index}) ->
+    if is_binary(Index) -> {binary_to_list(Index), <<"">>};
+       is_list(Index)   -> {Index, <<"">>}
+    end;
+index_for_group({struct,[{<<"index">>, Index}]}) ->
```
rzezeski: Would it be possible to move this normalization code into riak_kv?  Seems odd that Yokozuna should care about the structure of how it came in.

src/yokozuna.erl
```diff
@@ -77,13 +81,11 @@ is_enabled(Component) ->
...
-mapred_search(Pipe, [Index, Query, Filter], _Timeout) ->
-    Index2 = case is_binary(Index) of
-                 false -> Index;
-                 true -> binary_to_list(Index)
-             end,
+mapred_search(Pipe, [Group, Query], _Timeout) ->
+    mapred_search(Pipe, [Group, Query, <<"">>], _Timeout);
+mapred_search(Pipe, [Group, Query, Filter], _Timeout) ->
+    {Index, BucketFilter} = index_for_group(Group),
+    WholeFilter = join_filters(BucketFilter, Filter),
```
rzezeski: We should change search fold to take a list of `Filters`.  Each filter should become its own `fq` param.  You can repeat `fq` param as many times as your want and Solr will do an implicit conjunction.  The reason to make them separate is that filter queries may be cached in solr.  And the cache key is determined by the query string.  By concatenating the user filter + bucket filter you are making unique keys and thus the bucket filter query cached result cannot be shared across query requests.  It also means more entries in the cache table and thus more on-heap memory in Solr land.

Although, I do wonder if we should not bother adding this implicit filter at all.  If the user decides, during migration, to migrate multiple Riak Search buckets to the same index then he or she should understand that a query on that index can return data under all buckets.  If they don't like this they can change their query code to pass an explicit bucket filter.

rzezeski:
The riak test passes.  So that's great.  I think the only two changes I would like to see are:
1. Default to Riak Search if both are enabled and no env value set.
2. Don't bother with the implicit filter.  Explain in migration doc (which I'm on the hook for) that if you migrate to many-to-one setup then you need to add filter if you don't want to include data from other buckets.

src/yokozuna.erl
beerriot: We'll have to normalize to the mochijson2 format, in that case.

The issue is that the user might decide to send either

```
"inputs":{"index":"foo","query":"bar:baz"}
```

or

```
"inputs":{"module":yokozuna,"function":"mapred_search","arg":[{"index":"foo"},"bar:baz"]}
```

The first, version is what is currently translated to the simple tuple. There's no way to translate the second, modfun, version into the nice `{index, Index}` form without special-case code in the modfun path of riak_kv_mrc_pipe. (I suppose there is the option of just not supporting the modfun version.)

I don't particularly like the mochijson2 format leakage here either, so I'll probably use the `kvc` module to hide it.

Done in 29199de.

src/yokozuna.erl
beerriot: If we don't have to add the implicit filter, we don't have to worry about mis-encoding it. I'm +1 on tearing it out and noting the behavior in the docs. I think it's a separate issue from riak_search migration, though, since it will equally effect new YZ users that have never touched RS.

rzezeski: Yes, +1 to tearing it out and documenting.

beerriot: Done in 6419789

rzezeski:
After much discussion with @beerriot we have decided to back-out almost all of these changes and simply treat the bucket name as the index name.  For example if someone uses the following input for HTTP map-reudce against Yokozuna then `foo` represents the index name.

`"inputs":{"bucket":"foo", "query":"bar"}`

This should only matter for users migrating from Riak Search.  They will just have to keep in mind that when they migrate they should manually keep a 1:1 bucket:index mapping like is the default in Riak Search.  If they do that then their map-reduce/search queries will run as they did before.  For new users they should instead use the more obvious form.

`"inputs":{"index":"foo", "query":"bar"}`

If the user has many buckets under index foo but wants to filter to only include results from one bucket then they should use the following form.

`"inputs":{"index":"foo", "query":"bar", "filter":"_yz_rb:my_bucket"}`

The reason to back out these changes is that it greatly simplifies the code and makes it less brittle.  Considering that the original idea of using 'bucket' to lookup the corresponding index is still something that could be confusing to users I'd rather choose the thing that makes the code less brittle.

rzezeski:
None of the code on this branch is needed since we decided to ditch the bucket vs. index idea.  The "bucket" is the index name.  It's as simple as that.

PR #175, opened 2013-09-20
Fix associated_buckets

rzezeski:
Fix the associated_buckets function to actually return bucket names.
Reduce the copy & paste, move to common location.  Add check to index
admin test.


beerriot:
Makes sense to me & test passes: +1

Issue #146, opened 2013-07-23
Add Visibility and Stats


rzezeski:
Add visibility and stats.  Visibility could be something like a function to check for common problems and report back status.  I.e. like `riak-admin diag`.  Perhaps integrate Solr stats into the Riak stats, or also track Yokozuna specific stats as well.
- Slow query log?

coderoshi:
@rzezeski wdyt about this? Is this covered by #318?

rzezeski:
No, this is something I'm still working on.

rzezeski:
I marked this as 2.0-beta because I promised the CSEs months ago that there would be a diagnostic tool that would look for obvious issues. You can see my partial work here:

https://github.com/basho/yokozuna/compare/develop...feature;rz;diag

Should this be a blocker for beta? Should it even be in 2.0 at all?

/cc @jonmeredith @gburd @michellep @coderoshi

jonmeredith:
I would mark as 2.0-rc - ready before we can cut a real release candidate
as we need something in there for early adopters to have a chance.

On Mon, Mar 24, 2014 at 11:46 AM, Ryan Zezeski notifications@github.comwrote:

> I marked this as 2.0-beta because I promised the CSEs months ago that
> there would be a diagnostic tool that would look for obvious issues. You
> can see my partial work here:
> 
> develop...feature;rz;diaghttps://github.com/basho/yokozuna/compare/develop...feature;rz;diag
> 
> Should this be a blocker for beta? Should it even be in 2.0 at all?
> 
> /cc @jonmeredith https://github.com/jonmeredith @gburdhttps://github.com/gburd
> @michellep https://github.com/michellep @coderoshihttps://github.com/coderoshi
> 
> ## 
> 
> Reply to this email directly or view it on GitHubhttps://github.com/basho/yokozuna/issues/146#issuecomment-38476746
> .

## 

Jon Meredith
VP, Engineering
Basho Technologies, Inc.
jmeredith@basho.com

rzezeski:
Initial version of the diagnostic module is in (`yz_diag`). Finally closing this little guy.

PR #179, opened 2013-09-20
Setup Index and Search Stats

coderoshi:
Adding a yokozuna stats service. This overloads many of the terms used in RS because 1) they were pretty good and 2) we have the same problems to solve, and 3) people will already be used to them. This is missing search fold metrics. I know. I don't want to miss pre2 because of that, I just need to write tests for it.

This PR is for #146

### DIFF
--- a/riak_test/yz_stat_test.erl
+++ b/riak_test/yz_stat_test.erl
@@ -49,15 +49,10 @@ populate_data_and_wait(Pid, Cluster, IndexBucket, Count) ->
     Values = populate_data(Pid, IndexBucket, Count, []),
     Search = <<"text:*">>,
     F = fun(_) ->
-        {ok,{search_results,R,Score,Found}} =
-            riakc_pb_socket:search(Pid, IndexBucket, Search, []),
-        case Found of
-            Count ->
-                Score =/= 0.0;
-            0 ->
-                false
-        end
-    end,
+                {ok,{search_results,R,Score,Found}} =
+                    riakc_pb_socket:search(Pid, IndexBucket, Search, []),
+                (Count == Found) and (Score =/= 0.0)
+        end,
     yz_rt:wait_until(Cluster, F),
     Values.
 
@@ -86,7 +81,7 @@ confirm_stats(Cluster) ->
 
     Node = select_random(Cluster),
     Stats = rpc:call(Node, yz_stat, get_stats, []),
-    
+
     ILatency = proplists:get_value(index_latency, Stats, []),
     ?assert(proplists:get_value(min, ILatency, 0) > 0),
     ?assert(proplists:get_value(max, ILatency, 0) > 0),
--- a/riak_test/yz_stat_test.erl
+++ b/riak_test/yz_stat_test.erl
@@ -90,22 +90,25 @@ confirm_stats(Cluster) ->
 
     yz_rt:wait_until(Cluster, fun check_stat_values/1).
 
+-define(STAT_NAME(Path), yz_stat:stat_name(Path)).
+
 check_stat_values(Node) ->
     Stats = rpc:call(Node, yz_stat, get_stats, []),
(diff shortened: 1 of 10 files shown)


riak_test/yz_stat_test.erl
```diff
@@ -0,0 +1,102 @@
...
+    {ok, _Obj} = riakc_pb_socket:put(Pid, PO, [return_head]),
+    populate_data(Pid, IB, Count - 1, [KV|Acc]).
+
+populate_data_and_wait(Pid, Cluster, IndexBucket, Count) ->
+    Values = populate_data(Pid, IndexBucket, Count, []),
+    Search = <<"text:*">>,
+    F = fun(_) ->
+        {ok,{search_results,R,Score,Found}} =
+            riakc_pb_socket:search(Pid, IndexBucket, Search, []),
+        case Found of
```
rzezeski: Using this case here will cause the function to throw a `case_clause` exception instead of returning false.  As shown below.  Instead you want to do something like.

``` erlang
(Count == Found) and (Score =/= 0.0)
```

Here's the error I hit.

```
================ yz_stat_test failure stack trace =====================
{{case_clause,8},
 [{yz_stat_test,'-populate_data_and_wait/4-fun-0-',4,
                [{file,"riak_test/yz_stat_test.erl"},{line,54}]},
  {rt,wait_until,3,[{file,"src/rt.erl"},{line,478}]},
  {yz_rt,'-wait_until/2-fun-0-',3,[{file,"riak_test/yz_rt.erl"},{line,220}]},
  {yz_rt,'-wait_until/2-lc$^0/1-0-',2,
         [{file,"riak_test/yz_rt.erl"},{line,220}]},
  {yz_rt,wait_until,2,[{file,"riak_test/yz_rt.erl"},{line,220}]},
  {yz_stat_test,populate_data_and_wait,4,
                [{file,"riak_test/yz_stat_test.erl"},{line,61}]},
  {yz_stat_test,confirm_stats,1,
                [{file,"riak_test/yz_stat_test.erl"},{line,83}]},
  {yz_stat_test,confirm,0,[{file,"riak_test/yz_stat_test.erl"},{line,21}]}]}
=======================================================================
```

src/yz_kv.erl
```diff
@@ -153,7 +155,9 @@ index(Obj, Reason, P) ->
                             index(Obj, Reason, Ring, P, BKey, ShortPL, Index);
                         false ->
                             dont_index(Obj, Reason, P, ShortPL, BKey)
-                    end;
+                    end,
+                    TD = timer:now_diff(os:timestamp(), T1),
+                    yz_stat:update({index_end, TD});
```
rzezeski: We need to wrap all this in try/catch/after to prevent missed decrements.  Move the try/catch up from `index/7` and into here.

rzezeski:
Below are graphs for a simple benchmark I ran.  Write 1 million entries then run a query for 5 minutes that matches one of them.  This benchmark focuses on the overhead of Riak rather than Solr.

There appears to be about a 3% hit on indexing and query by adding stats.

### Indexing

| Measurement | No Stats | Stats |
| --- | --- | --- |
| Mean Thruput | 849 ops/s | 826 ops/s (97%) |
| Median Latency | 37.9ms | 39.2ms (1.3ms +) |
| 95th Latency | 59.6ms | 61.6ms (2ms +) |
| 99th Latency | 77.7ms | 78.0ms (0.3ms +) |

### Query

| Measurement | No Stats | Stats |
| --- | --- | --- |
| Mean Thruput | 1615 ops/s | 1559 ops/s (97%) |
| Median Latency | 20.4ms | 21.1ms (0.7ms +) |
| 95th Latency | 27.9ms | 29.1ms (1.2ms +) |
| 99th Latency | 33.1ms | 34.5ms (1.4ms +) |

coderoshi:
+1

PR #182, opened 2013-09-23
Require a JDK during build process.

bkerley:
These changes add a couple checks to `tools/build-solr.sh` and `tools/build-jar.sh`, checking for the presence of `javac`, `jar`, and `ant` and failing out when they're not present.

### DIFF
--- a/tools/build-jar.sh
+++ b/tools/build-jar.sh
@@ -3,6 +3,11 @@
 # Temporary script to build JAR file containing customer Solr request
 # handlers.
 
+if [ ! -x "`which javac`" ] || [ ! -x "`which jar`" ]; then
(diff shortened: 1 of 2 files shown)


hazen:
:+1: Builds for me on OSX. Just hope `which` is available on all of our supported platforms.

rzezeski:
Pinging @jaredmorrow since he has dealt with all of our major cross-platform scripts.

jaredmorrow:
Two things, first we don't actually build solr on a lot of platforms, we are simply fetching it after it is already built.  Second, even if we did, this is at build time and not riak runtime.  At build time I actually have a functioning shell even on old machines.

FWIW many build machines have maven, but not ant, since we have not required ant for anything.  This should not be affected since we only run `grab-solr.sh` and not `build-solr.sh`.

+1 from me.

PR #184, opened 2013-09-25
Convert `index_name()` to binary

rzezeski:
Use a binary index name to stay consistent with binary bucket name,
key name, etc.


coderoshi:
code: +1. tests: +1.
final score: +1

hazen:
@coderoshi I question your math. Clearly it would be +2

PR #187, opened 2013-09-25
SOLR_PKG_DIR skips download and uses a local copy

bookshelfdave:
we often build different version of riak_ee from scratch, this skips the lengthy download and uses a local copy. Feel free to tell me gtfo.


rzezeski:
I think this is a good idea.  But isn't `[[` a "bashism"?

rzezeski:
Never mind about that last comment.  It's already using bash.  +1

PR #188, opened 2013-09-26
Move wait_for_solr out of yz_solr_proc:init

beerriot:
Solr startup may take several seconds. All `gen_server:init` functions must finish within five seconds, though. So, waiting for Solr to become available in `yz_solr_proc:init` is not practical.

This PR splits the startup into two stages.
1. In `init`, we still make sure the data directory is setup correctly, and also spawn the JVM process.
2. After returning from init, `yz_solr_proc` checks once per second whether Solr has finished starting. If it hasn't finished by the set timeout, `yz_solr_proc` exits with a message about the timeout.

In order to support this new startup timeout, the application supervisors have also been rearranged. The `yz_solr_proc` process now lives under its own `yz_solr_sup` supervisor, while all other servers now live under a `yz_general_sup` supervisor. This is to allow for different restart policies for the children of each. Both of these new supervisors now live under the top-level `yz_sup`, where the restart strategy sets max restarts to zero. The idea is that if any child of the `solr` or `general` supervisor misbehaves enough that one of those supervisors exits, the whole app should tear down.

The restart strategy of `yz_solr_sup` attempts to match Solr's startup time. This PR sets things so that `yz_solr_sup` is not allowed to restart more than once within a span three times the length of the configured startup timeout. The idea is that if Solr falls over, the supervisor will attempt to restart it once, but if it fails to restart fast enough, the restart is not tried again.

A too-slow initial startup looks a little odd: `yz_solr_sup` exits with a message about the startup timeout, and then the supervisor tries to restart it. Typically this restart happens too quickly for the JVM to die and release network resources, though, so the restart will fail immediately when Jetty gets an error about ports already being in use. The biggest problem I foresee with this is that users are likely to send us log excerpts of the wrong error message. I'm not sure I see a clean fix yet, though.

The final commit of this PR also moves the registration with the `riak_core_node_watcher` service to `yz_solr_proc` so that it doesn't happen until Solr has become available. This also ties the service's status to the `yz_solr_proc` process instead of the yz supervisor. Since Erlang application startup no longer waits for Solr, this seemed the right thing to do.

~~The test for this change is `yz_solr_start_timeout`. It's a little bit of a tricky test. What it needs to do is prevent Solr from starting, but allow the JVM to start. This will keep `yz_solr_proc` in its wait state until it reaches its timeout. The best way I've found to make this happen so far is to delete `solr.war` from the node under test. An error is printed in the Solr log, but Jetty still hangs around anyway. This makes testing a little sticky for two reasons:~~
1. ~~If yz_solr_start_timeout fails, the dev node will be left in a state that will never correctly start solr, because solr.war is missing.~~
2. ~~I had to change `verify.sh` to use `devrel` instead of `stagedevrel`. With `stagedevrel`, the `solr.war` is contained in a symlinked yokozuna directory. The git-reset done by riak_test between tests does not reach into symlinked directories, so `solr.war` would never be restored.~~

~~A possible better test would be to use intercepts to override `yz_solr_proc:build_cmd/3` to make it return any command other than Solr that would stick around but not respond to Solr pings. Using the "manual" test-specific method of installing intercepts is not fast enough to get one installed before the process starts, though. The global config version may be, but I didn't try because the correct location of the intercept module was also uncertain.~~

The test has been changed to pass a well-known string for the data directory. This well-known string causes `yz_solr_proc` to start `grep` instead of jetty/solr. Since grep will never respond to a solr ping, `yz_solr_proc` eventually reaches its timeout. Hopefully our test machines have grep installed. This change means that `verify.sh` goes back to using `stagedevrel` safely.

### DIFF
--- a/src/yz_general_sup.erl
+++ b/src/yz_general_sup.erl
@@ -1,0 +1,63 @@
+%% -------------------------------------------------------------------
+%%
+%% Copyright (c) 2013 Basho Technologies, Inc.  All Rights Reserved.
+%%
+%% This file is provided to you under the Apache License,
+%% Version 2.0 (the "License"); you may not use this file
+%% except in compliance with the License.  You may obtain
+%% a copy of the License at
+%%
+%%   http://www.apache.org/licenses/LICENSE-2.0
+%%
+%% Unless required by applicable law or agreed to in writing,
+%% software distributed under the License is distributed on an
+%% "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+%% KIND, either express or implied.  See the License for the
+%% specific language governing permissions and limitations
+%% under the License.
+%%
+%% -------------------------------------------------------------------
+
+%% @doc Supervisor for long-lived processes *other* than yz_solr_proc,
+%% because they don't want to live under yz_solr_sup's specific
+%% restart strategy.
+
+-module(yz_general_sup).
+-behaviour(supervisor).
+-include("yokozuna.hrl").
+-export([start_link/0]).
+-export([init/1]).
+
+
+%%%===================================================================
+%%% API
+%%%===================================================================
+
+start_link() ->
+    supervisor:start_link({local, ?MODULE}, ?MODULE, []).
+
+
+%%%===================================================================
+%%% Callbacks
+%%%===================================================================
+
+init([]) ->
+    Events = {yz_events,
+              {yz_events, start_link, []},
+              permanent, 5000, worker, [yz_events]},
+
+    HashtreeSup = {yz_index_hashtree_sup,
+                   {yz_index_hashtree_sup, start_link, []},
+                   permanent, infinity, supervisor, [yz_index_hashtree_sup]},
+
+    EntropyMgr = {yz_entropy_mgr,
+                  {yz_entropy_mgr, start_link, []},
+                  permanent, 5000, worker, [yz_entropy_mgr]},
+
+    Cover = {yz_cover,
+             {yz_cover, start_link, []},
+             permanent, 5000, worker, [yz_cover]},
+
+    Children = [Events, HashtreeSup, EntropyMgr, Cover],
+
+    {ok, {{one_for_one, 5, 10}, Children}}.
--- a/src/yz_solr_proc.erl
+++ b/src/yz_solr_proc.erl
@@ -143,7 +143,7 @@ terminate(_, S) ->
             ok;
         Pid ->
             os:cmd("kill -TERM " ++ integer_to_list(Pid)),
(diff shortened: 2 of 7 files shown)


src/yz_general_sup.erl
```diff
@@ -0,0 +1,63 @@
...
+%% "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+%% KIND, either express or implied.  See the License for the
+%% specific language governing permissions and limitations
+%% under the License.
+%%
+%% -------------------------------------------------------------------
+
+%% @doc Supervisor for long-lived processes *other* than yz_solr_proc,
+%% because they don't want to live under yz_solr_sup's specific
+%% restart strategy.
```
coderoshi, line 23: they don't, do they? :)

src/yz_solr_proc.erl
```diff
@@ -193,6 +219,13 @@ get_pid(Port) ->
 
 %% @private
 %%
+%% @doc send a message to this server to remind it to check if solr
+%% finished starting
+schedule_solr_check(WaitTimeSecs) ->
+    erlang:send_after(1000, self(), {check_solr, WaitTimeSecs-1}).
```
coderoshi, line 225: For my own edification, how is this better than the `wait_for_solr` loop with timer?

src/yz_sup.erl
```diff
@@ -42,30 +53,18 @@ init([false]) ->
...
+    SolrSup = {yz_solr_sup,
+               {yz_solr_sup, start_link, []},
+               permanent, 5000, supervisor, [yz_solr_sup]},
 
-    Cover = {yz_cover,
-             {yz_cover, start_link, []},
-             permanent, 5000, worker, [yz_cover]},
+    GeneralSup = {yz_general_sup,
+                  {yz_general_sup, start_link, []},
+                  permanent, 5000, supervisor, [yz_general_sup]},
```
coderoshi: The HashtreeSup has a startup time of infinity, while this is 5000. Shouldn't this be infinity as well then?

coderoshi:
Tests pass. Assuming sane answers to my questions, +1

src/yz_solr_proc.erl
beerriot, line 225: It doesn't block the server until the check completes. Since no other process requests anything from this server, there's not much practical difference. But not blocking might aid debugging (one could ask for the pid of the JVM before solr is up, for instance).

src/yz_sup.erl
beerriot: That's shutdown time, not startup, but yes, that makes sense. Will fix.

coderoshi:
+1 for realsies

PR #190, opened 2013-09-27
Remove entropy data timestamp

rzezeski:
The entropy data timestamp is a holdover left from the very early days
of Yokozuna.  The 3rd commit to be exact.  The idea was to allow only
building a tree from data up to a certain time.  This was to prevent
spurious repair in a non-quiescent cluster.

However, the AAE system in KV doesn't have this feature and thus it is
never used.  It is simply dead code taking up time and space.  Time to
say goodbye.

### DIFF
--- a/java_src/com/basho/yokozuna/handler/EntropyData.java
+++ b/java_src/com/basho/yokozuna/handler/EntropyData.java
@@ -69,11 +69,6 @@ public class EntropyData
         BytesRef cont = contParam != null ?
             decodeCont(contParam) : DEFAULT_CONT;
 
-        // TODO: Make before required in handler config
-        String before = req.getParams().get("before");
-        if (before == null) {
-            throw new Exception("Parameter 'before' is required");
-        }
         int n = req.getParams().getInt("n", DEFAULT_N);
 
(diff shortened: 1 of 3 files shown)


rzezeski:
I ran the fruit load benchmark on the before and after.  Comparing both index throughput & latency as well as index size.

| Measurement | Before | After (No TS) |
| --- | --- | --- |
| Mean Thruput | 840 ops/s | 856 ops/s (+2%) |
| Median Latency | 38.2ms | 37.5ms (-1.9%) |
| 95th Latency | 61.4ms | 58.6ms (-4.6%) |
| 99th Latency | 78.5ms | 78.9ms (+0.5%) |
| Index Size | 23.1M | 21.4M (-7.4%) |

Removing the timestamp improves all latency except the 99th.  I have
no idea why this occurred.  Perhaps, whatever dominates the long tail
has nothing to do with generating this timestamp.

Notice that the index size, even for this small benchmark of storing 1
million small text blobs, has shrunk 7.4%.

coderoshi:
+1. good code, integration tests still work.

Minor, but it might be worth incrementing EntropyData's getVersion.

PR #193, opened 2013-09-30
Various riak test cleanup

rzezeski:
An attempt to make more use of built-in riak test functions and move common actions into `yz_rt`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,12 +2,12 @@ REBAR = $(shell pwd)/rebar
 
 .PHONY: deps rel stagedevrel test
 
-all: deps compile
+all: deps compile-riak-test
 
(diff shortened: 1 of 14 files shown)


rzezeski:
Some of the tests will fail until basho/riak_api#37 is resolved.

rzezeski:
If you notice any failures related to `rt:build_cluster` let me know.  I see it fail all the time but haven't made a PR yet.  Want to see if others hit first.

coderoshi:
This may be a good time to rename the tests to a yz_*_test standard.

yokozuna_essential fails with the "Unrecognized message gen_cast" error.

Otherwise, +1

Issue #192, opened 2013-09-27
Handle Schema Not Found


drewkerrigan:
Encountering the following:

```
{error,{badmatch,{error,<<"not_here">>,notfound}},[{yz_pb_admin,process,2,[{file,"src/yz_pb_admin.erl"},{line,67}]},{riak_api_pb_server,process_message,4,[{file,"src/riak_api_pb_server.erl"},{line,383}]},{riak_api_pb_server,connected,2,[{file,"src/riak_api_pb_server.erl"},{line,221}]},{riak_api_pb_server,decode_buffer,2,[{file,"src/riak_api_pb_server.erl"},{line,359}]},{gen_fsm,handle_msg,7,[{file,"gen_fsm.erl"},{line,505}]},{proc_lib,init_p_do_apply,3,[{file,"proc_lib.erl"},{line,239}]}]}
```

When calling:

```
riakc_pb_socket:get_search_schema(Pid,<<"not_here">>).
```

Because of:

```
process(#rpbyokozunaschemagetreq{name = SchemaName}, State) ->
    {ok, Content} = yz_schema:get(SchemaName),
    Schema = #rpbyokozunaschema{name = SchemaName, content = Content},
    {reply, #rpbyokozunaschemagetresp{schema = Schema}, State};
```

PR #194, opened 2013-10-01
Send correct message when schema is not found

coderoshi:
Fixes the failure case in #192


rzezeski:
+1 to merge, tests pass.

PR #195, opened 2013-10-01
Cache Solr archive

rzezeski:
Cache the Solr archive in /tmp/yokozuna to avoid superfluous network
activity.

The very first time you build Yokozuna (or if the Solr version changes) it will pull Solr archive from S3.  All subsequent builds will check `/tmp/yokozuna` for a local cache and use a symlink.  Thus reducing network usage, disk usage, and build time.

Example of build output when cached copy is used.

```
Using cached copy of Solr /tmp/yokozuna/solr-4.3.0-yz.tgz
```

### DIFF
--- a/tools/grab-solr.sh
+++ b/tools/grab-solr.sh
@@ -4,6 +4,7 @@
 #
 # Usage:
 #     ./grab-solr.sh
+#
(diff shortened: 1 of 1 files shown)


Vagabond:
+1 works as advertised.

coderoshi:
+1

PR #145, opened 2013-10-02
Security

coderoshi:
Added two new permissions:
- yokozuna.search
- yokozuna.admin

The first permission is read access for an index. Currently it does not support types, only bucket name which represent an index. After speaking with @Vagabond I'm going to expand this to support a default type `index`, so you could grant a user search access like this:
- `riak-admin security grant yokozuna.search ON index indexName TO userName`

We may wish to make this a reserved word for bucket types.

The admin permissions are universal per type index or schema. Meaning, you have access to index, you can read, list, create and delete them. For example:
- `riak-admin security grant yokozuna.admin ON index TO userName`
- `riak-admin security grant yokozuna.admin ON schema TO userName`

We may find a need to break these up later into more fine-grained permissions, but that seems a bit overzealous at this point in time, because it would potentially turn yokozuna's 3 permissions into 7.


rzezeski:
> After speaking with @Vagabond I'm going to expand this to support a default type index, so you could grant a user search access like this:

What is a "default type index"?  Is that a bucket type or security type?

Also, if I'm reading this correctly.  Where we currently use the name "bucket" in security it will actually stand for "index" when granted one of the `yokozuna.*` permissions?  Just want to make sure I'm following correctly.  Examples might help.

> We may wish to make this a reserved word for bucket types.

I'm not sure I understand this fully.  Are you saying that, if we add an "index" type to security that we should prevent a user from creating a bucket type with the name "index" because that would just confused things?

coderoshi:
> What is a "default type index"?  Is that a bucket type or security type?

Bucket type. There's no such thing as a security type.

> Also, if I'm reading this correctly. Where we currently use the name "bucket" in security it will actually stand for "index" when granted one of the yokozuna.\* permissions? Just want to make sure I'm following correctly. Examples might help.

I'm not sure what other example you're asking for. This is standard command line:

`riak-admin security grant <permissions> ON ANY|<type> [bucket] TO <users>`

This is an example for adding search permission to the `<type>` _index_ and `[bucket]` _indexName_:

`riak-admin security grant yokozuna.search ON index indexName TO userName`

Or for all indexes

`riak-admin security grant yokozuna.search ON index TO userName`

> I'm not sure I understand this fully. Are you saying that, if we add an "index" type to security that we should prevent a user from creating a bucket type with the name "index" because that would just confused things?

Yep.

rzezeski:
So "index" is a bucket type?  I think I still fundamentally don't understand bucket types.  To me a bucket type is a way for many buckets to share the same properties in an efficient manner.  I don't see the usefulness or sense in having an index bucket type because that assumes all buckets want to use the same index.  I.e. if you had the `yz_index` property defined on the bucket type "index" then all buckets that are indexed would all go to the same index or they would all have to overwrite this value at the bucket property/ring level.  Am I misunderstanding what "type" means here?

In the security example, `<type>` is a bucket type?

Vagabond:
You're way overthinking this. Security operates on 2-tuples, of
{ThingOne, ThingTwo}, If you grant a user permission on ThingOne, they
get implicit access for any tuple with ThingOne as the first element. If
you Grant a user a permission on a {ThingOne, ThingTwo} they only have
access to the specific element unde ThingOne. This is wildcard access vs
specific access.

riak_core_security doesn't know about bucket types, or anything, it is
just fundamentally providing access control based on a user's grants and
the permissions requested. The API can if the user has a particular
permission globally, matching just the first element of the tuple, or a
complete match.

So if you used something like "yz_index" as 'ThingOne', and made the
specific index as 'ThingTwo' you could do things like this.

GRANT riak_search.get ON yz_index TO andrew
GRANT riak_search.edit_schema ON yz_index myindex TO andrew

So I can read ANY index, but I can only edit the schema on the 'myindex'
index. The first grant has an implicit wildcard in it.

Point is, this actually has nothing to do with bucket types.

Andrew

coderoshi:
Sorry @Vagabond, as you were typing this, we had taken this conversation into mumble. I think everything is getting cleared up.

src/yz_pb_admin.erl
```diff
@@ -44,9 +44,25 @@ init() ->
...
+%%      also checks that this request has permission
 decode(Code, Bin) ->
-    {ok, riak_pb_codec:decode(Code, Bin)}.
-
+    Msg = riak_pb_codec:decode(Code, Bin),
+    case Msg of
+        #rpbyokozunaschemaputreq{} ->
+            {ok, Msg, {"yokozuna.schema", ?YZ_SECURITY_THING_ONE}};
+        #rpbyokozunaschemagetreq{} ->
+            {ok, Msg, {"yokozuna.schema", {?YZ_SECURITY_THING_ONE}}};
```
Vagabond: These should not be wrapped in a tuple. it should be like #rpbyokozunaschemaputreq{} and #rpbyokozunaindexdeletereq{}

See an example:

https://github.com/basho/riak_kv/blob/develop/src/riak_kv_pb_bucket.erl#L71

src/yz_wm_index.erl
```diff
@@ -123,6 +124,41 @@ content_types_accepted(RD, S) ->
...
+%% as opposed to a special yokozuna-specific config
+forbidden(RD, Ctx=#ctx{security=undefined}) ->
+    {riak_kv_wm_utils:is_forbidden(RD), RD, Ctx};
+forbidden(RD, Ctx=#ctx{security=Security}) ->
+    case riak_kv_wm_utils:is_forbidden(RD) of
+        true ->
+            {true, RD, Ctx};
+        false ->
+            Res = riak_core_security:check_permission({"yokozuna.index",
+                                                       ?YZ_SECURITY_THING_ONE},
```
Vagabond: Depending on the action, aren't these operations tied to a specific index? At least in some cases, it seems like this resource should  require {?YZ_SECURITY_THING_ONE, Index} as a permission.

src/yz_app.erl
```diff
@@ -57,6 +57,7 @@ maybe_setup(true) ->
     maybe_register_pb(?QUERY_SERVICES),
     maybe_register_pb(?ADMIN_SERVICES),
     setup_stats(),
+    ok = riak_core:register(yokozuna, [{permissions, [search,index,schema]}]),
```
Vagabond: Should there be more granular permissions for working with the schema?

src/yz_wm_schema.erl
```diff
@@ -87,6 +90,40 @@ malformed_request(RD, S) ->
...
+%% as opposed to a special yokozuna-specific config
+forbidden(RD, Ctx=#ctx{security=undefined}) ->
+    {riak_kv_wm_utils:is_forbidden(RD), RD, Ctx};
+forbidden(RD, Ctx=#ctx{security=Security}) ->
+    case riak_kv_wm_utils:is_forbidden(RD) of
+        true ->
+            {true, RD, Ctx};
+        false ->
+            Res = riak_core_security:check_permission({"yokozuna.schema",
+                                                       ?YZ_SECURITY_THING_ONE},
```
Vagabond: I don't know how schemas and indexes relate, but again this permission check feels over-broad?

src/yz_pb_admin.erl
coderoshi: Thanks, typo

src/yz_wm_index.erl
coderoshi: As you could probably tell from the vestigial tuples from above, I originally set per index permissions. But realistically, it seemed too narrow. The only thing you can do with an index is get its information, which you could get from a list. Same goes with schema. You can't actually update anything. I can add per-index grants back, if anyone has strong feelings about it.

riak_test/yz_security.erl
```diff
@@ -0,0 +1,253 @@
...
+    Pid0 = get_secure_pid(Host, Port),
+    lager:info("verifying user cannot create index without grants"),
+    ?assertMatch({error, <<"Permission", _/binary>>},
+        riakc_pb_socket:create_search_index(Pid0, ?INDEX_B)),
+
+    lager:info("verifying user cannot list indexes without grants"),
+    ?assertMatch({error, <<"Permission", _/binary>>},
+        riakc_pb_socket:list_search_indexes(Pid0)),
+
+    riakc_pb_socket:stop(Pid0),
```
Vagabond, line 139: You shouldn't need to close the connection, GRANT and then re-open the connection. At worst, a sleep() should suffice.

src/yz_wm_schema.erl
coderoshi: You can only create or read a schema. The same thought as above went into removing per-schema

riak_test/yz_security.erl
coderoshi, line 139: yes I was closing then opening because I didn't want to sleep.

riak_test/yz_security.erl
```diff
@@ -0,0 +1,253 @@
...
+
+    lager:info("Grant index permission to user"),
+    ok = ?GRANT(Node, [["yokozuna.index","ON","index","TO",?USER]]),
```
Vagabond: Feels like also trying GRANT .. ON index ?INDEX TO USER might be more interesting, so both the wildcard and specific index grant can be tested.

src/yz_app.erl
coderoshi: I put a lot of consideration into this. I wanted to keep it smaller, since it's always easier to add than remove permissions. Unlike KV, yokozuna isn't mature enough yet to know how people are going to want to use it. Practically speaking, I'm sure there will be users who can search, and administrators who can manage indexes and schemas. I'm open to adding more, but for 2.0 I just didn't see the value in it.

src/yz_wm_index.erl
Vagabond: Well, you can _create_ indexes with a PUT, no?

src/yz_wm_schema.erl
Vagabond: I'd argue that _creating_ vs _reading_ are different enough that you'd want to be able to more granularly control them, but I'm not a expert in this field.

src/yz_wm_index.erl
coderoshi: sure, but it's sort of weird giving someone permission to create an index that doesn't exist yet.

Vagabond:
If you want to just fix the typo (which maybe the test should have caught?) I can live with the permission structure you've chosen.

rzezeski:
The tests pass (I tested against this branch with a fresh merge of develop), that's a good sign. Still need to run benchmarks and run through the code. This will not make the tech preview.

```
yz_rs_migration-bitcask      : fail
yz_flag_transitions-bitcask  : pass
```

rzezeski:
Is this output from `print-user` confusing to anyone else in the context of Yokozuna? The `type` is `index`. I think that's probably okay since we don't call it `bucket type` but the second column is called `bucket` which has a very specific meaning in Riak.  So what is a user to think about a `yokozuna.index` grant for all buckets? That doesn't even make sense. I think I'm rehashing our old discussion about the fact that the term "bucket" is being overloaded. This is when @Vagabond explained the "thing1 vs thing2" to me. This grant is saying that the user can create indexes. But is that clear from the table below?

```
Applied permissions

+----------+----------+----------------------------------------+
|   type   |  bucket  |                 grants                 |
+----------+----------+----------------------------------------+
|  index   |    *     |             yokozuna.index             |
| default  |  fruit   |        riak_kv.put, riak_kv.get        |
+----------+----------+----------------------------------------+
```

Here is after I add the search grant. It seems like maybe that second column should have a more generic title. But I'm not sure what that is.

```
Applied permissions

+----------+----------+----------------------------------------+
|   type   |  bucket  |                 grants                 |
+----------+----------+----------------------------------------+
|  index   |  fruit   |            yokozuna.search             |
|  index   |    *     |             yokozuna.index             |
| default  |  fruit   |        riak_kv.put, riak_kv.get        |
+----------+----------+----------------------------------------+
```

Vagabond:
I plan to have something like letting applications register permissions with an associated 'granularity', so kv could use 'buckets' and yz could use 'indexes' and we could print seperate tables for seperate granularities.

rzezeski:
Benchmark results are in. In general indexing performance drops by about 3% with security enabled. Query by about 6%. And the patch itself doesn't seem to add any overhead. I only benchmarked protocol buffers but I would hope HTTP isn't much different. Most of the cost should be in connection establishment. Which, by the way, takes much longer now. Long lived connections will be vital with security enabled.

### Indexing Results

| Measurement | Security OFF | Security ON | Delta |
| --- | --- | --- | --- |
| Mean Throughput | 1986 ops/s | 1917 ops/s | \- ~3.5% |
| Median Latency | 12.1 ms | 13.1ms | \+ ~8.2% |
| 95th Latency | 38.3 ms | 40.1 ms | \+ ~4.7% |
| 99th Latency | 77.1 ms | 72.0 ms | \- ~6.6% |

### Query Results

| Measurement | Security OFF | Security ON | Delta |
| --- | --- | --- | --- |
| Mean Throughput | 2191 ops/s | 2065 ops/s | \- ~5.8% |
| Median Latency | 14.8 ms | 15.6 ms | \+ ~5.4% |
| 95th Latency | 22.6 ms | 24.3 ms | \+ ~7.5% |
| 99th Latency | 27.4 ms | 29.1 ms | \+ ~6.2% |

### Before/After Patch

I wanted to make sure that the patch itself didn't incur any overhead. These results show that before the patch and with the patch with security off the results are the same. This is good.

| Measurement | No Patch | Security OFF | Delta |
| --- | --- | --- | --- |
| Index Thru | 1999 ops/s | 1986 ops/s | \- 0.7% |
| Query Thru | 2184 ops/s | 2191 ops/s | \+ 0.3% |

rzezeski:
So I've been looking at the grants for Yokozuna and I'm not sure they really make sense to me. E.g. granting the permission `yokozuna.index` on `index`. What does that even mean? To me it should be more like `grant ALL ON index ALL`.

Here's the break down of Type (Thing-1) and associated grants that make sense to me. Thing-2 is used as a specific instance of the type. For most grants it probably only makes sense for that to be `ALL`.

| Type | Grants |
| --- | --- |
| Index | create, get, list, search, delete |
| Schema | create, get |

Here are some examples that I had in mind.

```
% grant all permissions on all indexes
grant ALL on index ALL TO user

% grant ability to create indexes, it doesn't make any sense to me to grant a user create privilage for a subset of indexes. either you can create an index or your can't. Same goes for get, list and delete.
grant create ON index ALL TO user
grant get ON index ALL TO user

% The search permission, however, can be limited to a specific index.
grant search ON index fruit TO user

% Or not.
grant search ON index ALL TO god

% Schemas are another item where either a user should have full access or none. Not per-schema permission.
grant create ON schema ALL TO user
```

I realize that this may not be feasible at the moment given implementation details but I really think it should look like something above. Now, if the permission need to be namespaced, e.g. `yokozuna.create`, then fine. I can live with that. But I strongly feel `grant yokozuna.index ON index TO user` is nonsensical.

coderoshi:
After another tea-party gab session, we've agreed to change the index/schema permission names to one permission `admin`, and instead use types (thing 1's)  to dictate what a permission is good for (`index` or `schema`). Example:
- `grant yokozuna.admin ON index TO user`
- `grant yokozuna.admin ON schema TO user`
- `grant yokozuna.admin ON ANY TO user` (both index and schema)

`yokozuna.search` will remain the same.

rzezeski:
After some discussion in backchannel with @coderoshi we came to the following solution.

| type | grants |
| --- | --- |
| index | admin, search |
| schema | admin |

The `admin` grant means that you can do all your basic admin stuff like put/get/list/delete.

The `search` grant means you can search. This only applies to the index type because it is non-sensical to "search" a schema in Yokozuna.

This would produce the following examples.

```
% grant admin on schemas
grant yokozuna.admin ON schema TO user

% grant admin on indexes
grant yokozuna.admin ON index TO user

% grant search on index fruit
grant yokozuna.search ON index fruit TO user

% grant search on all indexes
grant yokozuna.search ON index TO user

% or maybe ALL, for those of us that like that sorta thing
grant yokozuna.search ON index ALL TO user
```

rzezeski:
+1 to merge.

PR #196, opened 2013-10-03
skip_java option on build, for riak devel w/o yoko

jrwest:


rzezeski:
I'm going to allow this in for now.  But soon plan to replace this required building with a pre-built jar.

PR #198, opened 2013-10-04
Fix solaris build by removing -v flag from `cp` commands

jaredmorrow:
Solaris 10 does not have a `cp -v` flag available so in the
spirit of dumbing our builds down to the lowest common
denominator, this commit removes it.

I tested this branch on solaris-10u9


rzezeski:
Solaris is the ultimate troll.

+1 to merge.

Issue #200, opened 2013-10-08
Update for new hashtree API


rzezeski:
The hashtree API was changed by this commit.

https://github.com/basho/riak_core/commit/e7a4f9ee311f9d967ce839140dd83b21271bc5db#diff-61653aa3ba5c9c8b67d8e8953ea7e8c4

Need to update Yokozuna AAE to use `compare/4`.

PR #201, opened 2013-10-08
Update to new hashtree API

rzezeski:
The hashtree module only exports `compare/4` now.

Remove some unused functions and dead code.


rzezeski:
Pinging @engelsanchez and @slfritchie in case they want to overlook this change.

rzezeski:
This addresses #200.

slfritchie:
Hi, my badbad for not looking for the right things that might hit YZ.  Sorry!

I guess the good news is that the [Count] kludge in the accumulator disappears in favor of a simple integer. +1 from me.

PR #202, opened 2013-10-09
Don't convert index name to list

rzezeski:
All index names are binary now.


coderoshi:
Yes, that is the case. +1

Issue #132, opened 2013-06-24
Add tests to giddyup


rzezeski:
Part of integrating into master will include adding the Yokozuna riak tests to giddyup.

hazen:
Issue created: https://github.com/basho/giddyup/issues/86

rzezeski:
The tests have been added to giddyup but the following config needs to be added to the `riak` and `riak_ee` configs in the builders for the tests to be discovered and run.

``` erlang
  {yz_dir, "<path-to-yokozuna>"},
  {test_paths, ["<path-to-yokozuna>/riak_test/ebin"]},
```

Let me know if that works @jaredmorrow.

rzezeski:
@jaredmorrow and I witnessed a successful run of the Yokozuna tests on one of the builders today.  Closing this ticket.

hazen:
Yeah!

PR #203, opened 2013-10-09
Make YZ_BENCH_DIR environment variable optional

hazen:
Instead you can use `{rt_bench_path, ["/tmp/yz-verify/yokozuna/misc/bench"]},` in your .riak_test.config so Giddyup can actually run the tests.

### DIFF
--- a/docs/TESTING.md
+++ b/docs/TESTING.md
@@ -52,6 +52,7 @@ Open `~/.riak_test.config` and add the following to the end.
     {yokozuna, [
                 {rt_project, "yokozuna"},
                 {rt_deps, ["<path-to-testing-dir>/riak_yz/deps"]},
+                {yz_dir, ["<path-to-testing-dir>/riak_yz"]},
                 {rtdev_path, [{root, "/tmp/rt"},
                               {current, "/tmp/rt/riak_yz"}]}
                ]}.
--- a/docs/TESTING.md
+++ b/docs/TESTING.md
@@ -52,7 +52,7 @@ Open `~/.riak_test.config` and add the following to the end.
     {yokozuna, [
                 {rt_project, "yokozuna"},
                 {rt_deps, ["<path-to-testing-dir>/riak_yz/deps"]},
-                {yz_dir, ["<path-to-testing-dir>/riak_yz"]},
+                {yz_dir, ["<path-to-testing-dir>/riak_yz/deps/yokozuna"]},
                 {rtdev_path, [{root, "/tmp/rt"},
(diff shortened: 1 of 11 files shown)


rzezeski:
Isn't `rt_bench_path` already used by other tests? I imagine this would point to something like `~/work/basho_bench`.  If that is the case then this change probably won't play well with other riak tests run by giddy up because `YZ_BENCH_DIR` points to something like `~/work/yokozuna/misc/bench`.

Perhaps the real fix, to make this less confusing, is to use the `rt_bench_path` to actually call the `basho_bench` command but use `YZ_BENCHR_DIR`, with `yz_bench_dir` as backup, as a path to add to the `code_paths` in the generated basho bench config.

The entire point of this env var is so that the `yz_driver` module can be found.  Using the `basho_bench` command under `deps/basho_bench` just happened to be convenient when I wrote the first yokozuna riak test over a year ago.

rzezeski:
Maybe even better is just get rid of YZ_BENCH_DIR completely, and add add a config `yz_dir`.  Then we can use `rt_config:get(yz_dir)` and have access to any files in Yokozuna that are helpful for testing.

hazen:
@rzezeski I created `rt_bench_path`, so it's unique unless I happened to reinvent the name.

I was asking the other day about the need for YZ_BENCH_DIR at all.  It would certainly be cleaner if we just went with a single RT config parameter.  `yz_dir` sounds like a great name.

Did I interpret your last comment correctly?  Nuke YZ_BENCH_DIR?

rzezeski:
Yea, go ahead and nude the env var `YZ_BENCH_DIR` for now.  Let's see if we can get by with just a `yz_dir` rt config.

rzezeski:
LOL at "nude".  Nuke!  I meant nuke!

hazen:
LOL! Typing? How do you type? Must be my naked finger memory.

OK, so should I add a second config parameter just for basho_bench?  Or actually build it in the test code?

```
YZBenchDir = rt_config:get(yz_dir) ++ "/misc/bench"
```

I guess I could define the `yz_bench_dir` off of `yz_dir` and do something like this:

```
YZBenchDir = rt_config:get(yz_dir) ++ rt_config:get(yz_bench_dir)
```

where `yz_bench_dir` is just a subdirectory.

rzezeski:
Just build it all off `yz_dir` for now.

rzezeski:
+1 to merge

rzezeski:
Related to #132

PR #204, opened 2013-10-10
Use `basho_bench` riak test config

rzezeski:
Use the `basho_bench` riak test config to play nice with the automated testers.

/cc @jaredmorrow 

Related to #132.

### DIFF
--- a/riak_test/aae_test.erl
+++ b/riak_test/aae_test.erl
@@ -25,7 +25,6 @@
 
 confirm() ->
     YZBenchDir = rt_config:get(yz_dir) ++ "/misc/bench",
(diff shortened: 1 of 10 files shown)


hazen:
So `yz_rt.erl` is the only place where BB is actually called?

rzezeski:
Yep.

```
$ grep -inr --color 'basho_bench' riak_test/*
Binary file riak_test/ebin/yz_rt.beam matches
riak_test/yz_rt.erl:176:    BB = filename:join([rt_config:get(basho_bench), "basho_bench"]),
```

rzezeski:
Merging without +1 to Jared can test this path.  Deal with it.

PR #205, opened 2013-10-11
Don't store all _yz fields

rzezeski:
The entropy data, node, partition and first partition fields do not
need to be stored.  They are only used for filtering by Yokozuna and
are of no immediate use to the user.  The `_yz_id` field is also of no
use but Solr's distributed search requires the id field to be stored.

By not storing these fields a lot of space is saved in the index.  In
one benchmark disk usage was reduced by 25%.

### DIFF
--- a/include/yokozuna.hrl
+++ b/include/yokozuna.hrl
@@ -294,14 +294,14 @@
 -define(YZ_ED_FIELD, '_yz_ed').
 -define(YZ_ED_FIELD_S, "_yz_ed").
 -define(YZ_ED_FIELD_XML, ?YZ_FIELD_XML(?YZ_ED_FIELD_S)).
--define(YZ_ED_FIELD_XPATH, "/schema/fields/field[@name=\"_yz_ed\" and @type=\"_yz_str\" and @indexed=\"true\" and @stored=\"true\"]").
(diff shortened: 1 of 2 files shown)


rzezeski:
Here are results from my fruit benchmark storing 1 million small text objects.  No change in index or query throughput.  Reduced index size by 25%.  While this benchmark didn't show a change in throughput that would likely not be the case in a larger data set with less memory available as stored fields can cause pressure on the file page cache.  The less fields stored, the better.  See the following for more information on stored fields and page cache pressure.

http://blog.jpountz.net/post/33247161884/efficient-compressed-stored-fields-with-lucene

|  | Mean Writes/s | Mean Query/s | Index Size |
| --- | --- | --- | --- |
| Store All | 849 w/s | 1575 q/s | 21.4 MB |
| Store Some | 851 w/s | 1584 q/s | 16.2 MB (75%) |

coderoshi:
`_yz_vtag` doesn't need stored either

rzezeski:
Well I was wondering if we could want `_yz_vtag` so you could tell which sibling matches with which index?  But looking back over the code that is also in `_yz_id` field.  It could be extracted out of that if absolutely needed.  I'll revisit this tomorrow.

coderoshi:
But it's not really a value that needs to be returned. It's just used to filter siblings. If someone wants a vtag, they should probably use the riak key to get the value from KV.

rzezeski:
Yep, I'm in agreement here.  Will update PR.

coderoshi:
+1

PR #207, opened 2013-10-14
fix git repo url

nlf:


rzezeski:
Actually, the documentation is correct.  However it should probably say "Clone riak" now.  There is no longer a custom branch of Riak needed anymore.

nlf:
Really? All I have to do currently is clone the default branch of riak to get yokozuna?

rzezeski:
Yep.  The default branch of Riak is now `develop`.  If you clone that and run `make` you should see `deps/yokozuna`.

Keep in mind that the develop branch is in a lot of flux ATM.  Chances of a checkout with broken bits around integration are high.  Keep an eye out for dragons.

nlf:
In that case, I'm closing this since it's wrong anyway :)

rzezeski:
Hopefully #208 clarifies things a bit.

PR #210, opened 2013-10-16
Use common URL prefix

rzezeski:
For consistency, use the yz prefix for all HTTP resources.


coderoshi:
Also needs updated:

https://github.com/basho/yokozuna/blob/develop/riak_test/yz_wm_extract_test.erl#L57

PR #212, opened 2013-10-16
Schema bucket should disallow siblings

coderoshi:
The new `default_bucket_props` have `allow_mult=true`, which is hell on the `_yz_schema` bucket, which shouldn't have sibling schema values.

### DIFF
--- a/src/yz_app.erl
+++ b/src/yz_app.erl
@@ -57,6 +57,7 @@ maybe_setup(true) ->
     maybe_register_pb(?QUERY_SERVICES),
     maybe_register_pb(?ADMIN_SERVICES),
     setup_stats(),
+    ok = yz_schema:setup_schema_bucket(),
     ok.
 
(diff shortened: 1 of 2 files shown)


rzezeski:
I agree with the idea. For writing schemas, at least for now, LWW makes sense. However, this solution still has holes and it relies on the default bucket-type which will use ring bucket properties. I think having a new bucket-type named 'yokozuna' makes sense. The Yokozuna app would have full control over it and I think metadata validators can be used to enforce `allow_mult` (/cc @jrwest). Even if they can't use bucket types makes sense as they are the future. It also means we can keep all Yokozuna's private buckets under one type. E.g. schemas would live in `{<<"yokozuna">>, <<"schemas">>}`.

All that said I know the siblings are currently breaking apps built on top of Yokozuna. I'm going to run the tests on this PR and then merge it and make a new issue to create a proper Yokozuna bucket-type.

rzezeski:
+1 to merge.

PR #213, opened 2013-10-17
Add resources doc

rzezeski:


docs/RESOURCES.md
```diff
@@ -0,0 +1,49 @@
+Resources
+=========
+
+Yokozuna Talks
+--------------
+
+Yokozuna, Scaling Solr with Riak - Berlin Buzzwords 2013 - [video][bbuzz2013v] | [slides][bbuzz2013s]
```
coderoshi: This is markdown, you can
- make
- these
- all
- lists

rzezeski: I know, thought it looked fine without it.  But if you want bullet points you got it!

Issue #209, opened 2013-10-15
PB search failure on stored boolean fields


rzezeski:
Dave Martorana disovered an issue with retrieving stored boolean fields over the PB interface.  It looks to be a missing function head for boolean atoms.

http://lists.basho.com/pipermail/riak-users_lists.basho.com/2013-October/013642.html

```
2013-10-15 01:21:17.480 [error] <0.2872.0>@yz_pb_search:maybe_process:95
function_clause
[{yz_pb_search,to_binary,[false],[{file,"src/yz_pb_search.erl"},{line,154}]},{yz_pb_search,encode_field,2,[{file,"src/yz_pb_search.erl"},{line,152}]},{lists,foldl,3,[{file,"lists.erl"},{line,1197}]},{yz_pb_search,encode_doc,1,[{file,"src/yz_pb_search.erl"},{line,144}]},{yz_pb_search,'-maybe_process/3-lc$^0/1-0-',1,[{file,"src/yz_pb_search.erl"},{line,76}]},{yz_pb_search,maybe_process,3,[{file,"src/yz_pb_search.erl"},{line,76}]},{riak_api_pb_server,process_message,4,[{file,"src/riak_api_pb_server.erl"},{line,383}]},{riak_api_pb_server,connected,2,[{file,"src/riak_api_pb_server.erl"},{line,221}]}]
```

rzezeski:
Fixed by #214.

PR #214, opened 2013-10-20
Fix PB search for stored boolean fields

rzezeski:
Addresses #209.
- Fix encoding function so that it treats boolean fields properly.
- Add test to verify retrieval of string, float, and boolean fields via PB search.
- Update the BB driver to enabled benchmarking of PB search.

### DIFF
--- a/src/yz_pb_search.erl
+++ b/src/yz_pb_search.erl
@@ -140,17 +140,33 @@ default([H|T], _) ->
 default(Value, _) ->
     Value.
 
+%% @private
+%%
+%% NOTE: Bypass `riak_pb_search_codec' to avoid 2-pass on `Doc'.
+-spec encode_doc([{field_name(), term()}]) -> #rpbsearchdoc{}.
 encode_doc(Doc) ->
-    EncodedDoc = lists:foldl(fun ?MODULE:encode_field/2, [], Doc),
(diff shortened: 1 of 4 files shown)


src/yz_pb_search.erl
```diff
@@ -140,17 +140,33 @@ default([H|T], _) ->
...
+
+%% @private
+%%
+%% NOTE: Need to do this here because `riak_pb_codec' doesn't convert
+%% numbers.
+-spec encode_val(number() | atom() | binary()) -> atom() | binary().
+encode_val(Val) when is_integer(Val) ->
+    integer_to_binary(Val);
+encode_val(Val) when is_float(Val) ->
+    float_to_binary(Val);
```
coderoshi, line 169: integer_to_binary and float_to_binary don't exist before R16

coderoshi:
Search passes +1

rzezeski:
If we go back to 15B0x or a user absolutely needs to run Yokozuna on it then we'll patch these calls. But for now going to revel in 16B02 :)

src/yz_pb_search.erl
jaredmorrow, line 169: I also got bit by this trying to see if we could still build on R15B01.

PR #215, opened 2013-10-21
Log JVM stdout/err to INFO level

rzezeski:
Sometimes when the JVM/Jetty/Solr fails to start the only indication
is found in stdout/err. Log at the INFO level. If things are running
properly then there should be no logging.


coderoshi:
I'd love to review this but no time. Too huge. Let's just +1 for now, then analyze performance later.

PR #216, opened 2013-10-22
Add anti-entropy dir to cuttlefish schema

coderoshi:
Got this warning when I started Riak w/ YZ enabled:

```
[warning] Config yokozuna/anti_entropy_data_dir is missing. Defaulting to: "./data/yz_anti_entropy"
```

But there is no setting (or default) in https://github.com/basho/yokozuna/blob/develop/priv/yokozuna.schema

cc/ @joedevivo

### DIFF
--- a/priv/yokozuna.schema
+++ b/priv/yokozuna.schema
@@ -42,6 +42,6 @@
 ]}.
 
 %% @doc The directory where aae data files are stored
-{mapping, "yokozuna.anti_entropy_data_dir", "yokozuna.anti_entropy_data_dir", [
-  {default, "./data/yz_anti_entropy" }
+{mapping, "yokozuna.anti_entropy_data_dir", "yokozuna.anti_entropy.data_dir", [
+  {default, "{{platform_data_dir}}/yz_anti_entropy" }
 ]}.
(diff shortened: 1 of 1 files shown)


rzezeski:
This should use `{{platform_data_dir}}`. Also makes me wonder if using `anti_entropy.data_dir` to be location of KV's data dir is confusing? I.e. we have 3 systems with AAE right now, KV gets implicit namespace while other two are explicit? Just throwing it out there to @joedevivo who likes to ponder this stuff.

https://github.com/basho/riak_kv/blob/develop/priv/riak_kv.schema#L103

coderoshi:
Paging Dr. @joedevivo!

joedevivo:
So, @rzezeski's question doesn't really apply to this PR, since he's questioning the namespace of the setting defined in the riak_kv.schema. I do; however, think this one should be something like 'yokozuna.anti_entropy.data_dir'

just my opinion. you're on to something about the riak_kv bit. will ponder more.

rzezeski:
+1 to @joedevivo's `yokozuna.anti_entropy.data_dir`. @coderoshi still need to update the default to use `{{platform_data_dir}}`.

rzezeski:
+1 to merge

Issue #186, opened 2013-09-25
Remove bucket from WM GET /yz/index


coderoshi:
Currently getting an index via the wm interface will return something like: 

[{"name":"docs","bucket":"docs","schema":"_yz_default"}]

"bucket" makes no sense here anymore. Kill it.

PR #222, opened 2013-10-24
Remove 'bucket' field from index info

rzezeski:
This is leftover from days of one-to-one mapping from bucket to
index. Now many buckets can point to one index.


rzezeski:
This addresses #186.

coderoshi:
+1

PR #226, opened 2013-10-26
Add security support to BB driver for PB conns

rzezeski:
Add support for using secure protocol buffer connections in the basho
bench driver.


rzezeski:
Related to #145.

Issue #189, opened 2013-09-27
`_yz_id` is not unique


rzezeski:
It has come to my attention that the uniqueness semantic of `_yz_id` has been broken.  This happened when I added support from many-to-one bucket-to-index mappings.  The current id uses `<key>_<logical partition #>`.  if 2 or more buckets have the same key then data could be clobbered.  It depends on how they hash.

Furthermore, with upcoming support of bucket types Yokozuna has to consider the possibility that even bucket names can repeat now.  Thus the `_yz_id` field should ultimately become:

`<type>_<bucket>_<key>_<logical partition #>`

rzezeski:
This will get fixed with #176 which I'm currently working on.

Issue #176, opened 2013-09-20
Yokozuna and Bucket Types


rzezeski:
https://github.com/basho/riak/issues/362

Add bucket type support to Yokozuna.

Yokozuna will:
- Continue to base index association on bucket properties.
- If the bucket is under the default type then the bucket prop `yz_index` must be set.  This is also the scenario for users migrating from Riak Search.
- If the bucket has a type with `yz_index` set then that bucket will inherit that index association.
- If the bucket has a type with `yz_index` set but also has `yz_index` set as a bucket prop then the latter wins.  This means buckets under one type can point to different indexes.
- Multiple buckets can point to one index.
- A bucket can only have one associated index.
- When migrating Riak Search indexes keep them under default type as to avoid moving the KV data.
- Add logic to `yz_index:associated_buckets(Index)` to understand types.  I.e. if a type has `yz_index` set and has 3 buckets underneath it then those 3 buckets, along with any other buckets pointing to `Index`, should be returned.
- Add special field `_yz_bt` to designate the bucket type of the Solr document.

rzezeski:
After talking with @jrwest tonight it sounds like `get_bucket` will handle the property inheritance automatically.  I.e. Yokozuna doesn't have to care where the index name comes from.  It's up to the user to know that a type property applies to all its buckets but can be overridden by a bucket property.

It sounds like `yz_index:associated_buckets` could be a problem.  Perhaps the best thing to do there is just return any types with `yz_index` set along with and bucket overrides.  E.g.

``` erlang
-spec assocaited(Index) -> [{type, [type_name()]}, {buckets, [bucket_name()]}].
```

jrwest:
one thing I forgot to mention when talking about this earlier is `get_bucket` can now return `{error, no_type}` when passed `{BucketType, BucketName}` and `BucketType` d.n.e (or is inactive)

rzezeski:
This will rely on basho/riak_kv#711.

PR #230, opened 2013-11-12
Integrate with bucket types

rzezeski:
**NOTE:** This relies on basho/riak_kv#711.
## Overall

Bucket types are the future of namespacing and property creation in
Riak. They allow efficient storage of "bucket properties" outside of
the Ring and 2-level namespacing of `Type` and `Name`.

Essentially the bucket type can now be either a lone `binary()`
(legacy) or a 2-tuple of `{Type :: binary(), Name ::
binary()}`. Internally, when the legacy version is encountered it is
considered to live under the `default` bucket type. For example the
bucket `<<"my_bucket">>` would become `{<<"default">>, <<"my_bucket">>}`.

Up until this point Yokozuna has used the bucket property `yz_index`
to determine where to index data. This commit changes that in some
ways. Legacy users will have existing data in buckets. Those buckets,
in 2.0, will be considered to live under the default type as described
above. For legacy buckets (the default type) Yokozuna will NOT respect
the `yz_index` property. Rather it will act like Riak Search and use
an index named the same as the bucket AS LONG AS the `search` property
is set to true. Once users upgrade to 2.0 they should start opting
into non-default bucket types since it is more efficient and newer
features require the use of non-default type. For these types of
buckets Yokozuna will still use the `yz_index` property. This property
will typically be set at the type level but can also be overridden per
name under a type. Yokozuna doesn't care. If that `{Type, Name}` has a
`yz_index` property then it will be indexed.

In summary:
- Legacy buckets (default type) will act like Riak Search. The index
  used must have same name as bucket and `search` property must be
  true. This is used to users migrating from Riak Search.
- All new users MUST use new style buckets made of Type + Name. In
  most cases the `yz_index` property will be set on the type and thus
  inherited by all names under it (many buckets to one index). The
  index DOES NOT have to have the same name.
## Handoff

Another important change revolves around handoff. Since Yokozuna
leeches off the KV vnode it doesn't have control over handoff like it
would if it were a true vnode. When a node joins KV can start shipping
data before the bucket type data has been shipped over. In that case
there will be no `yz_index` property and indexes will be missing. AAE
would eventually catch this but it is poor form that node join would
cause a degradation in harvest, especially in a quiescent cluster.

To fix this Yokozuna needs more control over the lifecycle of the KV
vnode. Yokozuna needs to hook into the `handoff_starting` stage and
verify that the bucket types data is shipped before data handoff
begins. This is accomplished by adding the `yz_kv:should_handoff` hook
which is hard-coded in the KV vnode for now.

This is important for removing the hack around index creation as
well. Currently Yokozuna has a pretty horrible busy-wait hack in its
index hook to make sure indexes are created on joining nodes before
doing the first write of a handoff. This busy-wait blocks the KV vnode
and is dangerous for vnode latency. In a future commit this busy-wait
will be replaced with a check in this new handoff hook.
## Removal of Automatic AAE Tree Clearing

Remove all functionality around automatic clearing of trees when
adding or removing the `yz_index` property on a bucket with data. This
was referred to as `sync_data` in the `yz_events` module. Also called
"flags" harking back to when Yokozuna had a one-to-one bucket-to-index
mapping.

The original intention was that adding an index to a bucket with data
should clear the AAE trees so that exchanges would start repairing
missing indexes. If setting the index property to the tombstone value
(removal) then a) data for that bucket should be purged from the index
and b) AAE trees should be cleared. After much thought I think this
implicit behavior hurts more than helps.

Actions like clearing all AAE trees can be very expensive. It will not
be obvious to all users that adding or changing `yz_index` could cause
expensive operations to occur. For example, clearing the AAE trees for
a database with billions or trillions of objects will be expensive to
rebuild. Rather than relying on AAE a more direct operation could be
offered that allows the user to re-index a bucket or subset of
data. When removing an index it makes more sense to let the user
delete the index entirely rather than do an implicit delete-by-query
which is doing a bunch of extra work for a index that is going to be
deleted anyways.
## Misc Changes
- Update all tests to work with bucket types.
- Update Basho Bench driver to work with bucket types.
- Make map-reduce extraction more efficient. This is the ugly hack
  found in `yokozuna:positions`.


rzezeski:
This addresses #176 and also #189.

src/yz_doc.erl
```diff
@@ -197,10 +206,14 @@ doc_ts(MD) ->
 %%       iterate.  Otherwise the doc would have to be fetched for each
 %%       entry.
 gen_ed(O, Hash, Partition) ->
+    %% Store `Vsn' to allow future changes to this format.
+    Vsn = <<"1">>,
```
coderoshi, line 210: Should probably be in defined yokozua.hrl, or at least in the head of this module

coderoshi:
It looks like this PR also reflects changes to the ED object, which may have been implied by the addition of types, but should also note that objects are versioned.

Beyond that, it all looks good. The tests pass and the code is clear.

My only question lies behind the decision to drop support for M:1 bucket:index mapping on the `default` bucket type. I didn't quite grok the rationale. I understand the desire to promote bucket types, but is there a technical case to neuter the default bucket type that I'm not seeing? Because it seems awkward that types would function differently than default, since to the casual user (well, to me at least), `default` is just another type.

coderoshi:
fwiw, I'll +1 this for now, because it wouldn't hurt to add behaviors in the future if so desired.

rzezeski:
The reason I neutered the default type is to stay consistent with new features being introduced in 2.0 and strongly push users away from bucket properties which bloat the ring which is a know stability issue. The new CRDTs and strong consistency also require non-default bucket type. This is because the default type is really meant for legacy as it is not controlled by the new bucket-type command nor stored in cluster metadata. Also, the way I wrote this it means Yokozuna, under the default type, acts like Riak Search. It's a strict 1:1 mapping which assumes bucket=index. My thinking is that makes migration easier as it doesn't require adding a yz_index field or the chance of making a typo.

That said, you still have a good point. There's no technical reason the buckets under default type couldn't use `yz_index`. Once I start work on the RS migration code again perhaps I'll revisit this as I'm not 100% sure this is the best way to do migration. However, I do strongly feel that users need to stay far away from bucket properties once 2.0 comes out. And I think restricting new features like Yokozuna to only non-default types helps force the issue.

coderoshi:
I thought so. I ain't mad +1

PR #231, opened 2013-11-13
Replace nested separator with dots

coderoshi:
Currently, the extractors separate nested XML or JSON objects with underscore (`_`). Let's replace this with dots (`.`). This should clarify that values are nested, since dot is a common separator (KVC, mongo nested queries).


rzezeski:
I think this should be fine. Do it up.

rzezeski:
+1, all tests passed.

Issue #221, opened 2013-10-24
Use POST search


rzezeski:
The larger the ring size the larger the `fq` param will become. POST search should be used to avoid larger URL issues.
- Make sure to benchmark this change.

https://github.com/basho/yokozuna/compare/develop...post-get

rzezeski:
Addressed in #236.

PR #236, opened 2013-11-15
Use POST for search

rzezeski:
- Use the POST method for search to prevent a large `fq` parameter
  from causing error.
- Don't pass request headers to Solr as this can break the POST
  search, e.g. if `content-length: 0` is passed.

### DIFF
--- a/src/yz_solr.erl
+++ b/src/yz_solr.erl
@@ -223,11 +223,11 @@ dist_search(Core, Headers, Params) ->
     end.
 
 search(Core, Headers, Params) ->
-    Encoded = mochiweb_util:urlencode(Params),
-    URL = ?FMT("~s/~s/select?~s", [base_url(), Core, Encoded]),
-    Body = [],
+    Body = mochiweb_util:urlencode(Params),
+    URL = ?FMT("~s/~s/select", [base_url(), Core]),
+    Headers2 = [{content_type, "application/x-www-form-urlencoded"}|Headers],
     Opts = [{response_format, binary}],
-    case ibrowse:send_req(URL, Headers, get, Body, Opts) of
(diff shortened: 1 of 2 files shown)


rzezeski:
This addresses #221.

rzezeski:
A simple query benchmark on my linux workstation shows a slight bump of ~3.5% in throughput (queries per second), as well as slight reductions in latency. I didn't expect that but nice to see.

| --- | --- | --- | --- |
| GET | 2190 | 14.8 ms | 27.3 ms |
| POST | 2267 | 14.3 ms | 26.6 ms |

src/yz_solr.erl
```diff
@@ -223,11 +223,11 @@ dist_search(Core, Headers, Params) ->
     end.
 
 search(Core, Headers, Params) ->
-    Encoded = mochiweb_util:urlencode(Params),
-    URL = ?FMT("~s/~s/select?~s", [base_url(), Core, Encoded]),
-    Body = [],
+    Body = mochiweb_util:urlencode(Params),
+    URL = ?FMT("~s/~s/select", [base_url(), Core]),
+    Headers2 = [{content_type, "application/x-www-form-urlencoded"}|Headers],
```
coderoshi, line 228: Interestingly, if Headers already contains a `content_type`, ibrowse will just convert the header value into a CSV, which is not valid HTTP. But solr also doesn't seem to care, so we're probably good.

coderoshi:
Looks good +1

src/yz_solr.erl
rzezeski, line 228: Hopefully nothing else is passing `content_type` header to search. I wouldn't think anything is given that search has  been GET up to this point. If it does become an issue we can replace with `lists:keystore`.

PR #237, opened 2013-11-16
Increase Solr startup timeout

rzezeski:
The default timeout of 15 seconds seems to be too low for some users
and also some of our CI machines. Double the timeout for slower
machines.


coderoshi:
:+1: I've had to do this before, may as well encode it

PR #235, opened 2013-11-21
Rename public facing Yokozuna to Riak Search 2.0

coderoshi:
We need to rename all of the cases where the term yokozuna is public facing, with some variety of `search`, or `rs`, so as to clarify that this is a Riak Search 2.0 project.
- [x] Update HTTP admin resource names from `yz` prefix to `rs`
- [x] Update all yokozuna.schema cuttlefish mappings from `yokozuna` prefix to `search`
- [x] Rename `yz_index` bucket label to `search_index`
- [x] Update docs with use new resource paths and settings
- [x] Rename `yokozuna.*` security to `search.*` (also change `yokozuna.search` to `search.query`)

Testing this relies on basho/riak-erlang-client#134.

After this is merged:
- the mapred test needs updated basho/riak_test#454
- the remaining drivers need the renamed bucket prop `search_index`

### DIFF
--- a/docs/CONCEPTS.md
+++ b/docs/CONCEPTS.md
@@ -1,9 +1,9 @@
 Yokozuna Concepts
 ==========
 
-This document goes over important concepts in Yokozuna.  This will not
-cover every detail of every concept.  Instead it is meant as an
-overview of the most important concepts.
+This document goes over important concepts in Yokozuna.
+This will not cover every detail of every concept.
+Instead it is meant as an overview of the most important concepts.
 
 Yokozuna is an Erlang Application
 ---------------------------------
--- a/docs/RELEASE_NOTES.md
+++ b/docs/RELEASE_NOTES.md
@@ -1,10 +1,41 @@
 Yokozuna Release Notes
 ==========
 
-0.12.0 (WIP)
+
+0.12.0
 ------
 
 ### Breaking Changes ###
+
+Yokozuna has been publicly renamed to Search 2. Although internally
+the Yokozuna name will still be used (in the code and project name),
+all public facing options will remove the use of "Yokozuna" or "yz"
+in favor of "Yokozuna" or simply "search".
+
+Configuration (riak.conf) changes all yokozuna prefixes to search:
+
+```
+## To enable Search 2, set this 'on'.
+search = on
+
+## The port number which Solr binds to.
+search.solr_port = 10014
+
+## The port number which Solr JMX binds to.
+search.solr_jmx_port = 10013
+
+# and so on...
+```
+
+The index property to associate a bucket will changed from
+`yz_index` to `search_index`.
+
+All HTTP administration request have changed to rename the `yz`
+prefix to `search`. To manage an index or schema, use, respectively:
+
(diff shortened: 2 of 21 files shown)


docs/CONCEPTS.md
```diff
@@ -1,50 +1,50 @@
-Yokozuna Concepts
+Search 2 Concepts
```
rzezeski: Why are we changing the codename here? I thought I was clear about leaving the in-repo docs alone?

docs/RELEASE_NOTES.md
```diff
@@ -1,11 +1,44 @@
-Yokozuna Release Notes
```
rzezeski, line 1: Definitely -1 to changing title of project in release notes.

rzezeski:
So I thought we agreed not to touch the in-repo documentation? Given these docs live in the Yokozuna repo I feel using the code name is the right thing to do. The docs on docs.basho.com can use "Search".  Also, I notice a lot of "Search 2". I thought we also said it would just be "Search"?

coderoshi:
Reverted docs change, since they're internal.

priv/yokozuna.schema
```diff
@@ -31,17 +31,23 @@
...
 %% arguments, i.e. -XX, may not be portable across JVM
 %% implementations.  E.g. -XX:+UseCompressedStrings.
-{mapping, "yokozuna.solr_jvm_args", "yokozuna.solr_jvm_args", [
+{mapping, "search.solr_jvm_args", "yokozuna.solr_jvm_args", [
   {default, "-Xms1g -Xmx1g -XX:+UseStringCache -XX:+UseCompressedOops"}
 ]}.
 
+%% @doc The data under which to store all Search related data.
+%% Including the Solr index data.
+{mapping, "search.data_dir", "yokozuna.yz_dir", [
```
rzezeski: This was removed in develop. No need for it.

rzezeski:
All tests pass. Looks good. Just need to pull the `yz_dir` back out of config.

```
Test Results:
```

coderoshi:
@rzezeski removed excess config. could also use a +1 for basho/riak-erlang-client#134

priv/yokozuna.schema
```diff
@@ -1,12 +1,12 @@
...
 %% @datatype enum on, off
-{mapping, "yokozuna", "yokozuna.enabled", [
+{mapping, "search", "yokozuna.enabled", [
   {default, off},
   {datatype, {enum, [on, off]}}
 ]}.
 
 { translation,
-  "yokozuna.enabled",
+  "search.enabled",
```
rzezeski: This needs to stay `yokozuna.enabled` since it is a translation. Otherwise Yokozuna fails to start because it is being passed untranslated value of `on`. I found this via the erlang client tests. It can't be caught with riak tests since they use advanced config.

rzezeski: Actually, so the first string needs to be `"yokozuna.enabled"` but the `Conf` still uses the search namespace so need to change call to `conf_get_value("search", Conf)`. That was a bit of a PITA to figure out.

rzezeski:
+1, do it.

PR #238, opened 2013-11-26
RS migration via AAE

rzezeski:
Another installment of Riak Search migration code. This pull request does the following.
- Brings the migration test up to speed with the latest code.
- Factors out some common test logic related to AAE.
- Revert the default vs. non-default bucket type change. I.e. all buckets, once again, can have indexes associated with them.
- Remove some dead code and fix type spec.

Most importantly. As of this PR it is possible to migrate a Riak Search deployment to a Yokozuna one.

There is still more to do for Riak Search migration. The biggest task being a command to iterate + index all data in a bucket or list of buckets. This is useful because migration will not require waiting for AAE and it will be useful in future for schema migrations. However, that change will require some more fundamental changes in the Yokozuna indexing code path. Rather than make an even bigger PR I thought it makes sense to push this one out so that it has a chance to make it in the 0.12.0 release.

This is related to issue #20.

### DIFF
--- a/src/yz_app.erl
+++ b/src/yz_app.erl
@@ -51,11 +51,13 @@ stop(_State) ->
 maybe_setup(false) ->
     ok;
 maybe_setup(true) ->
+    Ring = yz_misc:get_ring(raw),
+    RSEnabled = yz_misc:is_riak_search_enabled(),
+    strip_rs_hooks(RSEnabled, Ring),
     Routes = yz_wm_search:routes() ++ yz_wm_extract:routes() ++
 	yz_wm_index:routes() ++ yz_wm_schema:routes(),
     yz_misc:add_routes(Routes),
-    maybe_register_pb(?QUERY_SERVICES),
-    maybe_register_pb(?ADMIN_SERVICES),
+    maybe_register_pb(RSEnabled),
     setup_stats(),
     ok = riak_core:register(yokozuna, [{permissions, [search,admin]}]),
     ok = yz_schema:setup_schema_bucket(),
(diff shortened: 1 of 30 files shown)


rzezeski:
This commit conflicts with #235. That one should land first and then this branch can be updated accordingly.

src/yz_app.erl
```diff
@@ -83,3 +85,84 @@ setup_stats() ->
...
+
+%% @private
+%%
+%% @doc Remove Riak Search pre-commit hook from all buckets when Riak
+%% Search is disabled.
+%%
+%% Previous versions of Riak had a bug in `set_bucket' which caused
+%% bucket fixups to leak into the raw ring.  If a user is upgrading
+%% from one of these versions, and enabled search on a bucket, then
+%% the Riak Searc hook will be in the raw ring.  After migrating to
```
coderoshi: *Search

src/yz_app.erl
```diff
@@ -51,11 +51,13 @@ stop(_State) ->
 maybe_setup(false) ->
     ok;
 maybe_setup(true) ->
+    Ring = yz_misc:get_ring(raw),
+    RSEnabled = yz_misc:is_riak_search_enabled(),
+    strip_rs_hooks(RSEnabled, Ring),
```
coderoshi: This code could be extracted into an external module, to keep this yz_app code cleaner

coderoshi:
+1, although I didn't grok what was happening in the `yz_app` fode copied from `riak_search_kv_hook`, which I feel would be better living in its own module, since it's effectively patch code that will live only as long as RS migrations.

PR #240, opened 2013-11-27
Replace references to yz with "search"

rzezeski:
Looks like we missed a few spots.


coderoshi:
hm, looks like a merge issue :+1:

PR #232, opened 2013-11-13
Correspondence to Apache Lucene 4.5.x, and detailed change of java_scr 

masahitojp:
- It corresponds to Lucene-5114 [remove boolean useCache param from TermsEnum.seekCeil/Exact].
  https://issues.apache.org/jira/browse/LUCENE-5114
  
  TermsEnum.SeekStatus seekCeil(BytesRef text, boolean useCache) is removed@Lucene 4.5.x
- "final" is added to the variable which is not changed.
- The processing which judges whether it is effective in logger is added.

PR #243, opened 2013-12-03
Upgrade to Solr 4.6.0

rzezeski:
As the title says. Still need to benchmark, but the tests pass.

```
Test Results:
```

This also includes changes made in #232 as they were required to upgrade.


java_src/com/basho/yokozuna/handler/EntropyData.java
```diff
@@ -117,24 +117,25 @@ public void handleRequestBody(SolrQueryRequest req, SolrQueryResponse rsp)
...
             }
 
-            String text = null;
-            String[] vals = null;
-            String docPartition = null;
-            String vsn = null;
-            String riakBType = null;
-            String riakBName = null;
-            String riakKey = null;
-            String hash = null;
```
hazen, line 127: Why was the null initialization removed here?

priv/conf/solrconfig.xml
```diff
@@ -123,63 +123,9 @@
     <unlockOnStartup>false</unlockOnStartup>
       -->
```
hazen, line 125: Was this following section removed by Apache or were we simply not using it?

rzezeski, line 125: It is all commented out so the defaults are in effect. I decided to remove it since some of these comments have changed in the example solfconfig from apache and I figured we are doing more harm by keeping around outdated, comment out config.

java_src/com/basho/yokozuna/handler/EntropyData.java
rzezeski, line 127: That was done by @masahitojp. I think he did this for stylistic reasons. If any of these variables were used without first initialization then IIRC the compiler will complain. So I think everything is fine. If it wasn't we would have seen NPEs long ago.

rzezeski:
Here are my benchmark results. Indexing seems to have improved a touch and querying seems to have slowed down a touch with 4.6.0, but could just be noise. In any case, I'm happy with it.

## Index

| --- | --- | --- | --- |
| 4.4.0 | 1756 | 14.00 ms | 83.46 ms |
| 4.6.0 | 1763 | 14.00 ms | 81.65 ms |

### 4.4.0

load-fruit-440

### 4.6.0

load-fruit-460

## Query

| --- | --- | --- | --- |
| 4.4.0 | 2206 | 14.64 ms | 27.75 ms |
| 4.6.0 | 2182 | 14.80 ms | 27.90 ms |

### 4.4.0

query-lima-440

### 4.6.0

query-lima-460

hazen:
All tests but one pass

:+1: With the stipulation that we address `yz_schema_admin` in another PR

rzezeski:
Yep, I wrote up an issue for the schema problem in #245.

PR #244, opened 2013-12-04
Use one-to-one mapping example in README

rzezeski:
Most users will probably have one bucket type that holds all bucket &
data. Use this style (one-to-one mapping) as the README example rather
than associating the index at the type level which causes it to be
used for all buckets (one-to-many).


hazen:
:+1:

PR #247, opened 2013-12-05
Fix logging of Solr's stdout/stderr

rzezeski:
As the title says.


src/yz_solr_proc.erl
```diff
@@ -121,7 +121,7 @@ handle_info({check_solr, WaitTimeSecs}, S=?S_MATCH) ->
             {stop, "solr didn't start in alloted time", S}
     end;
 handle_info({_Port, {data, Data}}, S=?S_MATCH) ->
-    ?INFO("solr stdout/err: ~p", Data),
+    ?INFO("solr stdout/err: ~p", [Data]),
```
seancribbs, line 124: OK, but I'd use `~s` since Data is always iodata().

rzezeski:
This is also fixed in #250. Rather than requiring Sean to rebase that branch I'll just close this issue.

Issue #245, opened 2013-12-04
Bad schema can cause indefinite errors


rzezeski:
WIth the upgrade to Solr 4.6.0 the CREATE core command now returns an error when the core already exists. This poses a problem for the "bad_class" test in [yz_schema_admin](https://github.com/basho/yokozuna/blob/develop/riak_test/yz_schema_admin.erl). @javajolt first hit this issue and I was able to replicate by hand. According to the Solr wiki this behavior should have existed in 4.4.0 but I'm guessing there was a bug because testing in 4.4.0 by hand showed that you could create an existing core without an error being returned.

> In Solr 4.3 and above, an error is returned, and RELOAD must be used.

http://wiki.apache.org/solr/CoreAdmin#CREATE

The fix will probably involve changing Yokozuna to differentiate between first time creation and reload of schema. Perhaps this can be accomplished with some state stored in cluster meta, calling core list from solr, or checking the data directory.

rzezeski:
The telltale sign of this issue is a log entry like the following. Specifically the `another core is already defined there`.

```
2013-12-04 17:33:05.978 [error] <0.530.0>@yz_index:local_create:156 Couldn't create index bad_class: {ok,"400",[{"Content-Type","application/xml; charset=UTF-8"},{"Transfer-Encoding","chunked"}],<<"<?xml version=\"1.0\" encoding=\"UTF-8\"?>\n<response>\n<lst name=\"responseHeader\"><int name=\"status\">400</int><int name=\"QTime\">3</int></lst><lst name=\"error\"><str name=\"msg\">Error CREATEing SolrCore 'bad_class': Could not create a new core in /spinning/bench/riak-460/dev/dev1/data/yz/bad_class/as another core is already defined there</str><int name=\"code\">400</int></lst>\n</response>\n">>}
```

PR #250, opened 2013-12-11
Add support for KV datatypes (counters, sets, maps)

seancribbs:
- Extracts top-level types into fields as described in `yz_dt_extractor`.
- Defines content types that match those used in [`riak_kv`](https://github.com/basho/riak_kv/blob/develop/include/riak_kv_types.hrl)
- Adds fields and dynamic fields to the default schema.

TODO:
- [x] riak_test for indexing CRDTs

### DIFF
--- a/priv/default_schema.xml
+++ b/priv/default_schema.xml
@@ -106,6 +106,16 @@
    <dynamicField name="*_ig" type="ignored" multiValued="true"/>
    <dynamicField name="random_*" type="random" />
 
+   <!-- Riak datatypes default fields-->
+   <field name="counter" type="int"    indexed="true" stored="true" />
+   <field name="set"     type="string" indexed="true" stored="false" multiValued="true" />
+
+   <!-- Riak datatypes embedded fields -->
+   <dynamicField name="*_flag"     type="boolean" indexed="true" stored="true" />
+   <dynamicField name="*_counter"  type="int"     indexed="true" stored="true" />
+   <dynamicField name="*_register" type="string"  indexed="true" stored="false" />
+   <dynamicField name="*_set"      type="string"  indexed="true" stored="false" multiValued="true" />
+
    <!-- catch-all field -->
    <dynamicField name="*" type="text_general" indexed="true" stored="false" multiValued="true" />
 
--- a/priv/default_schema.xml
+++ b/priv/default_schema.xml
@@ -113,7 +113,7 @@
    <!-- Riak datatypes embedded fields -->
    <dynamicField name="*_flag"     type="boolean" indexed="true" stored="true" />
    <dynamicField name="*_counter"  type="int"     indexed="true" stored="true" />
-   <dynamicField name="*_register" type="string"  indexed="true" stored="false" />
+   <dynamicField name="*_register" type="string"  indexed="true" stored="true" />
    <dynamicField name="*_set"      type="string"  indexed="true" stored="false" multiValued="true" />
 
    <!-- catch-all field -->
--- a/src/yz_dt_extractor.erl
+++ b/src/yz_dt_extractor.erl
@@ -1,0 +1,136 @@
+%% -------------------------------------------------------------------
+%%
+%% Copyright (c) 2013 Basho Technologies, Inc.  All Rights Reserved.
+%%
+%% This file is provided to you under the Apache License,
+%% Version 2.0 (the "License"); you may not use this file
+%% except in compliance with the License.  You may obtain
+%% a copy of the License at
+%%
+%%   http://www.apache.org/licenses/LICENSE-2.0
+%%
+%% Unless required by applicable law or agreed to in writing,
+%% software distributed under the License is distributed on an
+%% "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+%% KIND, either express or implied.  See the License for the
+%% specific language governing permissions and limitations
+%% under the License.
+%%
+%% -------------------------------------------------------------------
+
+%% @doc An extractor for Riak KV's datatypes. Maps have their fields
+%% concatenated with `field_separator'. Sets are converted into
+%% multi-valued fields. Counters are treated as integer fields.
+%% Embedded flags are treated as boolean fields. Embedded registers
+%% are treated as regular fields.
+%%
+%% Counter example:
+%%
+%%    [{<<"counter">>, <<"5">>}]
+%%
+%% Set example:
+%%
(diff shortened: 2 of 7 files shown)


seancribbs:
@rzezeski I'd love a little help here, for some reason yokozuna with my branch creates an unusable cluster in riak_test. In the original directory, the dev nodes start up just fine. In my rtdev, solr bombs. Here's a console output: https://gist.github.com/7936269

src/yz_extractor.erl
```diff
@@ -32,7 +32,11 @@
                       {"application/json",yz_json_extractor},
                       {"application/xml",yz_xml_extractor},
                       {"text/plain",yz_text_extractor},
-                      {"text/xml",yz_xml_extractor}]).
+                      {"text/xml",yz_xml_extractor},
```
rzezeski: This list needs to be ordered as the `ordsets` modules is used to determine extractor.

seancribbs: Addressed in a0fb872

src/yz_dt_extractor.erl
```diff
@@ -0,0 +1,136 @@
...
+%%     {{<<"page_views">>, riak_dt_pncounter}, 1502},
+%%     {{<<"events">>, riak_dt_map},
+%%      [{{<<"RICON">>, riak_dt_lwwreg}, <<"spoke">>},
+%%       {{<<"Surge">>, riak_dt_lwwreg}, <<"attended">>}]}]
+%%
+%%    [{<<"activated_flag">>, true},
+%%     {<<"name_register">>, <<"Ryan Zezeski">>},
+%%     {<<"phones_set">>, <<"555-5555">>},
+%%     {<<"phones_set">>, <<"867-5309">>},
+%%     {<<"page_views_counter">>, <<"1502">>},
```
coderoshi, line 54: I'm curious why *_flag is boolean, but *_counter is binary?

seancribbs, line 54: I'm following convention of the JSON extractor. Should it be an integer?

lenary:
:+1: seems all good to me (though I should really run the tests)

rzezeski:
Yes, please someone make sure they paste the riak_test results output in the PR. All tests but `yz_schema_admin` should pass.

seancribbs:
@rzezeski Is it that my schema changes broke `yz_schema_admin` or something else?

rzezeski:
@seancribbs No, caused by upgrade to solr 4.6 (#245).

src/yz_dt_extractor.erl
coderoshi, line 54: Nevermind, this will work fine. I forgot we converted json numbers to binary too.

priv/default_schema.xml
```diff
@@ -106,6 +106,16 @@
...
    <dynamicField name="random_*" type="random" />
 
+   <!-- Riak datatypes default fields-->
+   <field name="counter" type="int"    indexed="true" stored="true" />
+   <field name="set"     type="string" indexed="true" stored="false" multiValued="true" />
+
+   <!-- Riak datatypes embedded fields -->
+   <dynamicField name="*_flag"     type="boolean" indexed="true" stored="true" />
+   <dynamicField name="*_counter"  type="int"     indexed="true" stored="true" />
+   <dynamicField name="*_register" type="string"  indexed="true" stored="false" />
```
coderoshi: I can see why set would be not be stored, but why is register not stored? By default all dynamic fields are stored except text_general because of their size. Is register generally going to be larger than a string type?

seancribbs: We don't know what people will put in a register (or a set for that matter), so the size could be unbounded. Does Solr have a limit on the size of string fields? Is there an opportunity here for some documentation?

coderoshi: The difference between a string and text is that text is analyzed (stemmed, downcase, punctuation ignored). We don't store text simply because it tends to store larger values, not because of any hard limit. Since it's really about reasonable defaults, it should depend on how people tend to use registers. Are users going to be more commonly storing a few bytes like strings, or kilo/megabytes like text?

seancribbs: I would hope smallish values (i.e. < 1KB), but there's no hard limit. How about I switch it to `stored="true"` and then we plan document this particular quirk.

coderoshi: :thumbsup:

src/yz_dt_extractor.erl
```diff
@@ -0,0 +1,136 @@
...
+extract_fields(Name, register, Value, #state{fields=Fields, field_separator=Sep}=State) ->
+    FieldName = field_name(Name, register, Sep),
+    State#state{fields=[{FieldName, Value}|Fields]};
+extract_fields(Name, flag, Value, #state{fields=Fields, field_separator=Sep}=State) ->
+    FieldName = field_name(Name, flag, Sep),
+    State#state{fields=[{FieldName, Value}|Fields]}.
+
+-spec field_name(field_path_name(), datatype(), binary()) -> binary().
+field_name(undefined, map, _Sep) ->
+    %% (timeforthat)
```
seancribbs: Got some time for that in 0a31d51

priv/default_schema.xml
seancribbs: Fixed in 0a31d51

coderoshi:
+1

seancribbs:
:dancers: WOO!

rzezeski:
Where's my riak test output?

seancribbs:
@rzezeski Would mine be sufficient? Do you need a full run of verify.sh?

rzezeski:
@seancribbs I just want to make sure someone ran the full suite. If no one did it's not a huge deal as I will run it on develop soon enough. I know I don't always follow this rule myself but moving forward it would be nice if the result output was a prerequisite for a PR being merged.

rzezeski:
Dialyzer had two issues with new extractor. I haven't looked at them closely as I'm trying to do something else ATM.

```
yz_dt_extractor.erl:93: The variable Err can never match since previous clauses completely covered the type #state{}
yz_dt_extractor.erl:99: The call yz_dt_extractor:extract_map(Name::binary()) breaks the contract (field_path_name()) -> fun(({{binary(),module()},term()},state()) -> state())
```

seancribbs:
@rzezeski Thanks, will fix.

coderoshi:
@rzezeski Sorry about that. My security failure is unrelated to this

```
yz_security-bitcask          : fail
1 Tests Failed
16 Tests Passed
```

rzezeski:
@coderoshi No problem, thanks for attaching the output.

PR #257, opened 2013-12-19
Delete core.properties file during core creation

rzezeski:
As of Solr 4.6.0 a Core can get into a weird state when it is CREATEd
with a bad schema file, e.g. a schema including a bad class name.
Solr will create a core.properties file but will ultimately fail to
initialize the Core because of the bad class. Yokozuna has been built to
treat CREATE as in idempotent action that it can retry until it works.
If the user uploads a fixed schema then the next instance of CREATE
should work. However, in 4.6.0 Solr now complains about creating a core
which doesn't exist in memory but has a core.properties file on disk.
The Solr documentation says that a RELOAD operation must be used but the
results in Solr returning an exception saying the core doesn't exist.
Furthermore, an UNLOAD action results in a NullPointerException when the
Core is in this state.

The core.properties file is generated by Solr. It contains minimal
information and deleting it doesn't damage the index. Thus, the easiest
solution was to simply delete the core.properties before calling CREATE.
If this is the first time the Core is being create then the file doesn't
exist anyways. If the Core is in this purgatory state then it allows
CREATE to retry initialization rather than fail.

This addresses #245.

### DIFF
--- a/riak_test/yz_schema_admin.erl
+++ b/riak_test/yz_schema_admin.erl
@@ -320,19 +320,18 @@ confirm_bad_schema(Cluster) ->
     lager:info("give solr time to attempt to create core ~s", [Name]),
     timer:sleep(5000),
     lager:info("verify solr core ~s is not up", [Name]),
-    ?assertNot(yz_solr:ping(binary_to_list(Name))),
+    ?assertNot(yz_solr:ping(Name)),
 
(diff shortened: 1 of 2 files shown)


coderoshi:
+1

```
yz_schema_admin-bitcask: pass
0 Tests Failed
1 Tests Passed
```

hazen:
Thanks, @coderoshi!

Issue #258, opened 2013-12-20
Two bugs in `yz_kv:should_handoff`


rzezeski:
@jrwest brought 2 issues to my attention related to `yz_kv:should_handoff`.
1. If Yokozuna is disable then this function should return `true`. Otherwise KV handoff could be blocked.
2. If Yokozuna is enabled a capability check must be made with remote node before doing remote call to get hash. See https://github.com/basho/yokozuna/blob/develop/src/yz_kv.erl#L165

The second issue can be solved by checking for the Yokozuna service on the remote side. If the service is not on remote then return `true` to allow handoff to complete. I don't think capabilities are needed.

jrwest:
capabilities would allow you to perform a local-only call but since you are already making one remote call... {shrug}

EDIT: reworded, sorry tired.

rzezeski:
@jrwest Does capabilities take into account if service is disabled? There could be case where user doesn't enable Yokozuna on all nodes (which would be a mistake, sigh non-global config). In that case the service would be down and `should_handoff` could return true so as not to block KV handoff. I don't think capabilities can deal with this situation, can it?

jrwest:
@rzezeski if I understand you correctly, what I've usually done in this situation is based on the app.config change how the capability is registered when the application starts. This way the cluster can't converge to "is enabled" when a single node is still disabled. 

https://github.com/basho/riak_kv/blob/develop/src/riak_kv_app.erl#L170-L172
https://github.com/basho/riak_kv/blob/develop/src/riak_kv_app.erl#L322-L327

rzezeski:
I found that condition 1 is not an issue. KV handoff will proceed just fine with the existing check. I verified this via the `verify_handoff` test. However, I'm still going to change the `should_handoff` function to only make calls if Yokozuna is enabled to avoid any breakage in the future.

rzezeski:
My fix for the second condition will be to check for the Yokozuna service on the other side. I tried to use capabilities but it currently has a race condition which causes the 2.0 node to think the 1.4.2 node has Yokozuna (which is impossible) and cause a crash on the call to the metadata server.

If the service is down (meaning Yokozuna is disabled, Solr is down, or pre 2.0 node) then bypass the metadata check. Otherwise perform it. Simple. A modified version of `verify_handoff_mixed` failed before the modification with the expected rpc failures and it passes after the modification. PR incoming once I get another PR in.

jonmeredith:
[...]

PR #275, opened 2014-01-15
Fix handoff check in mixed cluster scenario

rzezeski:
Perform the metadata check only when Yokozuna is enabled on both the local and remote node. Otherwise just return `true` and let KV perform handoff.

This addresses #258. See commit messages for more detail.

I tested with a modified `verify_handoff_mixed` test. Below is the diff.

``` diff
diff --git a/tests/verify_handoff_mixed.erl b/tests/verify_handoff_mixed.erl
index c4a7441..b278574 100644
--- a/tests/verify_handoff_mixed.erl
+++ b/tests/verify_handoff_mixed.erl
@@ -54,11 +54,13 @@ confirm() ->
     UpgradeVsn = proplists:get_value(upgrade_version,
                                      riak_test_runner:metadata(),
                                      previous),
-    SearchEnabled = [{riak_search, [{enabled, true}]}],
-    Versions = [{current, SearchEnabled},
-                {UpgradeVsn, SearchEnabled}],
+    RSEnabled = [{riak_search, [{enabled, true}]}],
+    YZEnabled = [{yokozuna, [{enabled, true}]}],
+    Versions = [{current, RSEnabled ++ YZEnabled},
+                {UpgradeVsn, RSEnabled}],
     Services = [riak_kv, riak_search, riak_pipe],
     [Current, Old] = Nodes = rt:deploy_nodes(Versions, Services),
+    rt:wait_for_service(Current, yokozuna),

     prepare_vnodes(Current),

@@ -122,7 +124,7 @@ prepare_search_vnodes(Node) ->
       [ list_to_binary(integer_to_list(N))
         || N <- lists:seq(1000, 1000+?SEARCH_COUNT) ]),
     riakc_pb_socket:stop(C).
-                     
+
 prepare_pipe_vnodes(Node) ->
     %% the riak_pipe_w_pass worker produces no archive, but the vnode
     %% still sends its queue (even if empty) through handoff
```

### DIFF
--- a/src/yz_kv.erl
+++ b/src/yz_kv.erl
@@ -160,7 +160,7 @@ get_short_preflist({Bucket, _} = BKey, Ring) ->
 %% check that.
 -spec should_handoff({term(), {p(), node()}}) -> boolean().
 should_handoff({_Reason, {_Partition, TargetNode}}) ->
-    case ?YZ_ENABLED of
+    case ?YZ_ENABLED andalso is_service_up(?YZ_SVC_NAME, TargetNode) of
         true ->
             case is_metadata_consistent(TargetNode) of
                 true ->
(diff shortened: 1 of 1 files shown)


russelldb:
Works for me +1 (tested with `verify_2i_mixed_cluster`, too!)

PR #234, opened 2014-01-15
Create internal bucket type for schemas

coderoshi:
Currently schemas are stored in a bucket under the default type and the `allow_mult` property must be set at app start time which adds to the size of the ring.

Create a bucket type dedicated for internal Yokozuna usage. Initially this could be used for schemas but it would be nice to also support solr configs so that they can be more easily upgraded across an entire index (perhaps that should be a separate PR).

Alternate idea: forget about storing stuff in KV completely and store in cluster metadata using gzip or other forms of compression.

@jrwest is there any practical limit or concerns about size of value stored in cluster meta?


coderoshi:
dammit master

PR #281, opened 2014-01-16
Use tools.mk in Makefile

Vagabond:
This also fixes dialyzing the tests, if you set the path to riak_test.


reiddraper:
Looks good to me. @rzezeski please confirm too before merging.

Vagabond:
I squished this branch, because it was getting too fat.

rzezeski:
+1, works for me.

PR #283, opened 2014-01-23
Better essential test

rzezeski:
After talking with a user on IRC I decided to make two small tweaks to the essential test. These changes make the test more rigorous around verification of the number of results returned.
1. Use a random query with 1-3 terms of different cardinality as opposed to querying for the same term over and over.
2. Transition from 2 -> 4 nodes rather than 3 -> 4. This forces co-located index replicas to exist thus making sure that cluster sizes less than the N value still produce the correct results.


coderoshi:
Makes sense. +1

```
Test Results:
yokozuna_essential-bitcask: pass
0 Tests Failed
1 Tests Passed
```

Issue #233, opened 2013-11-13
Use new handoff hook to wait for index creation


rzezeski:
Currently there is a very ugly hack in the index function to busy-wait for index creation during handoff so that first N objects don't fail to index causing missing indexes. This hack uses the pdict which means it ASSUMES it lives on KV vnode and it also can block the vnode for seconds. Using the new handoff hook would allow to wait for indexes to migrate to a joining node first before starting the handoff sequence. This would remove use of the pdict and blocking of the vnode. It should also simplify the code in general.

Handoff hook: basho/riak_kv#711.

PR #287, opened 2014-01-29
Verify Index Exists Locally Before Handoff

coderoshi:
Altered the should_handoff check (called by the new [handoff hook](https://github.com/basho/riak_kv/pull/711)) to return false until all indexes expected to exist on a node handoff exist (respond to a solr ping).

This resolves issue #233, where previous, yokozuna would busy-wait when a put was indexed by a handoff.

### DIFF
--- a/src/yz_kv.erl
+++ b/src/yz_kv.erl
@@ -25,11 +25,6 @@
 -compile(export_all).
 -include("yokozuna.hrl").
 
--define(ONE_SECOND, 1000).
--define(WAIT_FLAG(Index), {wait_flag, Index}).
--define(MAX_WAIT_FOR_INDEX, app_helper:get_env(?YZ_APP_NAME, max_wait_for_index_seconds, 5)).
-
--type check() :: {module(), atom(), list()}.
 -type write_reason() :: delete | handoff | put | anti_entropy.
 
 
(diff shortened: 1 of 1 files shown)


hazen:
```
Test Results:
0 Tests Failed
17 Tests Passed
```

coderoshi:
Sorry @javajolt, I rewrote the index verification in it's entirety. Another pass?

hazen:
No problem. Just as long as cleverly crafted code, upon reflection, has not been replaced by three boring lines.

src/yz_kv.erl
```diff
@@ -164,6 +159,17 @@ should_handoff({_Reason, {_Partition, TargetNode}}) ->
             true
     end.
 
+%% @doc Checks pending node changes for remote nodes'
+%% indexes, and tests locally that they currently exist
+-spec pending_indexes_exist() -> boolean().
+pending_indexes_exist() ->
```
rzezeski: Already went over this in backchannel but adding quick recap. This function needs to a) get index list from local ring and then b) verify those indexes exist on the `TargetNode`. Instead this function is pinging the nodes on the node _sending_ data. The reason the tests passed is because the target nodes just so happened to create the indexes before the `should_handoff` function returned true. I.e. there is a race. I was able to remove the race by adding a 3s sleep to the `yz_solr:core` function (with that modification the test will fail and you will see index failures in the dev3 log file). It would be good to add an intercept which adds a 10s sleep to `yz_solr:core` (but only when it is called for create) during the `yokozuna_essential` test.

Also, rather than using ping this function could use an rpc call to `yz_solr:cores` and compare that list with the index list from the local ring. Keep in mind both lists may need to be sorted first to us `==` for comparison.

Finally, I'm afraid we still need the logic that only waits so long for the indexes to exist. Remember that the `should_handoff` callback is blocking KV handoff from starting. Yokozuna should do its best to not interfere with KV. In this case, if the indexes still don't exist after N tries then we need to give up and let KV proceed. Otherwise people might get very upset about KV handoff being stuff.

src/yz_kv.erl
```diff
@@ -164,6 +160,14 @@ should_handoff({_Reason, {_Partition, TargetNode}}) ->
             true
     end.
 
+%% @doc Checks pending node changes for remote nodes'
+%% indexes, and tests locally that they currently exist
+-spec target_indexes_exist(node()) -> boolean().
+target_indexes_exist(RemoteNode) ->
+    LocalIndexes = yz_solr:cores(),
```
rzezeski: The `yz_solr:cores` function returns `{ok, Cores}` and so the sort functions below will fail. Running `yokozuna_essential` shows this as it never completes and you'll see errors in the dev1/2 logs.

src/yz_kv.erl
```diff
@@ -164,6 +160,14 @@ should_handoff({_Reason, {_Partition, TargetNode}}) ->
             true
     end.
 
+%% @doc Checks pending node changes for remote nodes'
+%% indexes, and tests locally that they currently exist
```
rzezeski: I find this doc hard to follow. I don't understand what "pending node changes" are nor do I follow the "tests locally". The point of this function is to verify that the remote node has the indexes that the local node thinks it should have. Furthermore this function is a predicate function so it would be nice if the function name made that more clear. Perhaps `has_indexes(Node, IndexList)` which makes it more clear this function takes a list of indexes and determines whether or not they exist on the target node.

rzezeski:
Okay, essential test passes now, even with the added sleep. Please squash into 1 commit so history doesn't include the broken commits. In the meantime I'm going to run full suite of tests.

rzezeski:
Tests passed, dialyzer looks good. +1 to merge.

```
Test Results:
0 Tests Failed
17 Tests Passed
```

Eric noted that he saw the `yz_index_admin` test fail once but I got 10/10 good runs and he cannot make it fail again either. Noting it here in case of future failures.

Issue #253, opened 2013-12-16
Port AAE changes


rzezeski:
Port over changes that also apply to Yokozuna.

basho/riak_kv#765

rzezeski:
Another PR to potentially port.

basho/riak_kv#819

rzezeski:
I'm choosing to not port the AAE rebuild throttle to Yokozuna as it is implemented in KV. The work YZ AAE does compared to KV is fundamentally different. It retrieves `{BKey, Hash}` tuples from the Solr term dictionary (it should probably be using DocValues to exploit column storage + less memory pressure). I think the throttle is a good idea but investigation needs to be done to see how to best apply the idea to Yokozuna.

rzezeski:
I do not believe basho/riak_kv#819 needs to be ported because a `riak_object:riak_object()` is always passed to `yz_kv:hash_object`.

Issue #248, opened 2013-12-05
Unnecessary AAE repair after restart


rzezeski:
I'm noticing that after a cluster restart unnecessary AAE repair will occur. I've seen this a few times in the past and I thought I fixed it but perhaps it regressed. My first hunch is it could have something to do with the fact that Riak KV doesn't flush it's tree buffer on vnode shutdown. This buffer can hold 200 entries and I notice all repair is < 200 keys per partition. I think @jtuple might be fixing the flush issue in KV in an upcoming PR.

As part of the fix the `aae_test` should be modified to restart cluster and verify no repair happens afterwards in order to make sure this regression doesn't happen again.

rzezeski:
I added a case to `aae_test` to verify that no repair occurs after cluster restart. It passes. For good measure I ran the `yokozuna_essential` test and restarted the cluster after it completed. It performed repair. I waited to repair to settle and then restarted the cluster. It repaired again on the second start. A quick glance shows it's all Yokozuna repair and it's always < 200 keys per repair. This makes me suspicious of the `hashtree.erl` buffer. I find it odd that there is no repair from restart after the `aae_test` but there is for the essential test. Another potential lead. I need to investigate further.
- [x] Make sure I'm using KV/Core with latest changes.
- [ ] <del>Try changing hashtree buffer to something smaller and see if repaired keys is less than that.</del>
- [x] Figure out why `yokozuna_essential` exhibits this behavior but not `aae_test`.

rzezeski:
Verified that I still see the issue with latest Core + KV. Comparing code I notice Yokozuna calls `hashtree:close` and assumes that flushes the buffer whereas KV calls `flush_buffer`. Perhaps there was a change to the hashtree module or close never flushed to begin with. In any case I'll update the Yokozuna code and see if that fixes the issue.

https://github.com/basho/riak_kv/blob/3f763d6f0f1d9157119cea3a572b533bee27972a/src/riak_kv_index_hashtree.erl#L876

https://github.com/basho/yokozuna/blob/0e47e1ea7d305ef448772a06a641f4bc63364c36/src/yz_index_hashtree.erl#L488

Also, I still should figure out why `yokozuna_essential` has this behavior but `aae_test` doesn't.

rzezeski:
The rabbit hole gets deeper...

Riak Test uses a hard shutdown (kill -9 against beam.smp pids) because
of hanging shutdown issues in the past. When restarting the nodes I
[...]

PR #293, opened 2014-02-05
Add AAE flush + new test

rzezeski:
Two things:
1. Add a test to verify that AAE doesn't repair after a restart when it doesn't need to.
2. Make sure to flush hashtrees during shutdown. This is one of the changes needed for #253.

The test is not necessarily meant to test the flushing logic. Even without the added flush call during close the test still passes (probably because of flush getting called via other code paths). I added it to for #248 but realized that my problems are probably related to riak test and not Yokozuna itself. Since I already did the work of writing the test and added the new flush logic I figured I might as well make a PR to add both of them.

### DIFF
--- a/riak_test/aae_test.erl
+++ b/riak_test/aae_test.erl
@@ -73,7 +73,7 @@ confirm() ->
 
     _ = verify_removal_of_orphan_postings(Cluster),
 
-    verify_no_repair(Cluster),
+    verify_no_indefinite_repair(Cluster),
 
     _ = verify_no_repair_for_non_indexed_data(Cluster, PBConns),
     yz_rt:close_pb_conns(PBConns),
@@ -162,7 +162,7 @@ verify_removal_of_orphan_postings(Cluster) ->
 %% @doc Verify that there is no indefinite repair.  There have been
 %% several bugs in the past where Yokozuna AAE would indefinitely
 %% repair.
-verify_no_repair(Cluster) ->
+verify_no_indefinite_repair(Cluster) ->
     lager:info("Verify no indefinite repair"),
     yz_rt:count_calls(Cluster, ?REPAIR_MFA),
     TS = erlang:now(),
--- a/riak_test/aae_test.erl
+++ b/riak_test/aae_test.erl
@@ -73,15 +73,13 @@ confirm() ->
 
     _ = verify_removal_of_orphan_postings(Cluster),
 
-    %% Verify that there is no indefinite repair.  The have been
-    %% several bugs in the past where Yokozuna AAE would indefinitely
-    %% repair.
-    lager:info("Verify no indefinite repair"),
     verify_no_repair(Cluster),
 
     _ = verify_no_repair_for_non_indexed_data(Cluster, PBConns),
-
     yz_rt:close_pb_conns(PBConns),
+
+    _ = verify_no_repair_after_restart(Cluster),
+
     pass.
 
 %% @doc Create a tuple containing a riak object and the first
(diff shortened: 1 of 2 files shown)


hazen:
```
Test Results:
0 Tests Failed
17 Tests Passed
```

From your description it sounds like the test will still pass even without your changes, so I won't bother seeing if I can get it to fail.  Please correct me if I'm wrong.

rzezeski:
> From your description it sounds like the test will still pass even without your changes, so I won't bother seeing if I can get it to fail. Please correct me if I'm wrong.

That's correct. The test is not necessarily related to the flush change. The test was added to catch any potential future regressions and the flush was added because it is one of the ports required for #253.

riak_test/aae_test.erl
```diff
@@ -159,8 +159,11 @@ verify_removal_of_orphan_postings(Cluster) ->
     ?assertEqual(Num, yz_rt:get_call_count(Cluster, ?REPAIR_MFA)),
     ok.
 
-%% @doc Verify that no repair has happened since `TS'.
+%% @doc Verify that there is no indefinite repair.  The have been
```
coderoshi: "There have been"

riak_test/aae_test.erl
```diff
@@ -159,8 +159,11 @@ verify_removal_of_orphan_postings(Cluster) ->
...
+%% several bugs in the past where Yokozuna AAE would indefinitely
+%% repair.
 verify_no_repair(Cluster) ->
```
coderoshi: maybe rename the function?

src/yz_index_hashtree.erl
```diff
@@ -485,9 +485,23 @@ maybe_build(S) ->
...
-    S#state{trees=Trees2, closed=true};
+    Trees2 = [begin
+                  NewTree =
+                      try
+                          hashtree:flush_buffer(Tree)
+                      catch _:_ ->
+                              lager:warning("Failed to flush trees during close"),
+                              Tree
+                      end,
+                  {IdxN, NewTree}
```
coderoshi, line 501: If you called `hashtree:close(NewTree)` here, you could avoid the next loop

coderoshi:
Everything looks great, so I pushed nitpick's turbo button.

src/yz_index_hashtree.erl
rzezeski, line 501: We could but perhaps we don't want to catch if close fails. Also I copied straight from KV as I want to keep the two as identical as possible so that applying patches is easier.

rzezeski:
Renamed test function. Test still passes.

```
Test Results:
aae_test-bitcask: pass
0 Tests Failed
1 Tests Passed
```

Do I get a +1?

coderoshi:
Oh yeah `1 Tests Passed` +1

jonmeredith:
Perhaps yz_aae_test-bitcask?

On Thu, Feb 6, 2014 at 8:47 AM, Ryan Zezeski notifications@github.comwrote:

> Renamed test function. Test still passes.
> 
> Test Results:
> 
> ## aae_test-bitcask: pass
> 
> 0 Tests Failed
> 1 Tests Passed
> That's 100.0% for those keeping score
> 
> Do I get a +1?
> 
> ## 
> 
> Reply to this email directly or view it on GitHubhttps://github.com/basho/yokozuna/pull/293#issuecomment-34336077
> .

## 

Jon Meredith
VP, Engineering
Basho Technologies, Inc.
jmeredith@basho.com

coderoshi:
Not a bad idea. It's about time that gets renamed

rzezeski:
> Perhaps yz_aae_test-bitcask?

I'll do that in another PR.

Issue #290, opened 2014-02-04
Add n_val to Index in Clients


coderoshi:
Adding n_val to the index requires setting it on the client.
- [x] [Add n_val to yokozuna PB interface](https://github.com/basho/yokozuna/pull/296)

Add n_val to:
- [x] [protobufs](https://github.com/basho/riak_pb/pull/80)
- [x] [Erlang client](https://github.com/basho/riak-erlang-client/pull/154)
- [x] [Ruby client](https://github.com/basho/riak-ruby-client/pull/146)
- [x] [Python client](https://github.com/basho/riak-python-client/pull/307)
- [x] Java client

PR #295, opened 2014-02-06
Add notion of n_val to index

rzezeski:
Add the notion of `n_val` to Yokozuna indexes. This reduces coupling between KV buckets and Yokozuna indexes and fixes a bug in the query plan generation. Removing the coupling to buckets also allows for usage of indexes not yet envisioned, e.g. indexing internal events like audit logs.

This patch also makes small tweeks to the essential test to use a non-default `n_val` to make sure that works. It also uses an index name different from the bucket as that would have caught this bug many months ago.

Finally, this patch extends the HTTP Index API to handle `n_val` but changes are still required for PB as well as the official clients. See #290.

### DIFF
--- a/src/yz_cover.erl
+++ b/src/yz_cover.erl
@@ -146,9 +146,8 @@ cache_plan(Index, Ring) ->
 -spec calc_plan(index_name(), ring()) -> {ok, plan()} | {error, term()}.
 calc_plan(Index, Ring) ->
     Q = riak_core_ring:num_partitions(Ring),
-    BProps = riak_core_bucket:get_bucket(Index),
     Selector = all,
-    NVal = riak_core_bucket:n_val(BProps),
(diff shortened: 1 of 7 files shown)


src/yz_cover.erl
```diff
@@ -146,9 +146,8 @@ cache_plan(Index, Ring) ->
 -spec calc_plan(index_name(), ring()) -> {ok, plan()} | {error, term()}.
 calc_plan(Index, Ring) ->
     Q = riak_core_ring:num_partitions(Ring),
-    BProps = riak_core_bucket:get_bucket(Index),
     Selector = all,
-    NVal = riak_core_bucket:n_val(BProps),
+    NVal = yz_index:get_n_val(yz_index:get_index_info(Index)),
```
coderoshi, line 150: :fist:

coderoshi:
+1. All verify.sh tests pass.

PR #303, opened 2014-02-13
Add bucket type/index n_val equivalency validator

coderoshi:
Registers a create/update bucket type property validator for yokozuna. This ensures that setting properties on a bucket type will validate that 1) the index exists, and 2) the bucket types n_val equals the index's n_val.

This forced a change to the tests, whereby an index must first be created, then `wait_for_index`, then associate the index with the bucket. Before this check, waiting didn't matter until values were written.

### DIFF
--- a/riak_test/yokozuna_essential.erl
+++ b/riak_test/yokozuna_essential.erl
@@ -276,8 +276,6 @@ setup_indexing(Cluster, PBConns, YZBenchDir) ->
     [yz_rt:wait_for_index(Cluster, I)
      || I <- [?INDEX, <<"tagging">>, <<"escaped">>, <<"unique">>]],
 
-    ok = rpc:call(Node, riak_core_bucket, set_bucket,
-                  [?BUCKET, [{n_val, ?INDEX_N_VAL}]]),
     yz_rt:set_index(Node, ?BUCKET, ?INDEX, ?INDEX_N_VAL),
     yz_rt:set_index(Node, {?BUCKET_TYPE, <<"tagging">>}, <<"tagging">>),
     yz_rt:set_index(Node, {?BUCKET_TYPE, <<"escaped">>}, <<"escaped">>).
--- a/riak_test/yokozuna_essential.erl
+++ b/riak_test/yokozuna_essential.erl
@@ -269,19 +269,18 @@ setup_indexing(Cluster, PBConns, YZBenchDir) ->
     yz_rt:store_schema(PBConn, ?FRUIT_SCHEMA_NAME, RawSchema),
     yz_rt:wait_for_schema(Cluster, ?FRUIT_SCHEMA_NAME, RawSchema),
     ok = yz_rt:create_index(Node, ?INDEX, ?FRUIT_SCHEMA_NAME, ?INDEX_N_VAL),
-    yz_rt:set_index(Node, ?BUCKET, ?INDEX),
+    ok = yz_rt:create_index(Node, <<"tagging">>),
+    ok = yz_rt:create_index(Node, <<"escaped">>),
+    ok = yz_rt:create_index(Node, <<"unique">>),
+
+    [yz_rt:wait_for_index(Cluster, I)
+     || I <- [?INDEX, <<"tagging">>, <<"escaped">>, <<"unique">>]],
+
     ok = rpc:call(Node, riak_core_bucket, set_bucket,
                   [?BUCKET, [{n_val, ?INDEX_N_VAL}]]),
-
-    ok = yz_rt:create_index(Node, <<"tagging">>),
+    yz_rt:set_index(Node, ?BUCKET, ?INDEX, ?INDEX_N_VAL),
     yz_rt:set_index(Node, {?BUCKET_TYPE, <<"tagging">>}, <<"tagging">>),
-
-    ok = yz_rt:create_index(Node, <<"escaped">>),
-    yz_rt:set_index(Node, {?BUCKET_TYPE, <<"escaped">>}, <<"escaped">>),
-
-    ok = yz_rt:create_index(Node, <<"unique">>),
-    [yz_rt:wait_for_index(Cluster, I)
-     || I <- [?INDEX, <<"tagging">>, <<"escaped">>, <<"unique">>]].
+    yz_rt:set_index(Node, {?BUCKET_TYPE, <<"escaped">>}, <<"escaped">>).
 
 verify_deletes(Cluster, KeysDeleted, YZBenchDir) ->
     NumDeleted = length(KeysDeleted),
--- a/src/yz_bucket_validator.erl
+++ b/src/yz_bucket_validator.erl
@@ -65,7 +65,7 @@ validate_n_val(Index, INVal, BNVal, BucketProps) ->
         INVal when INVal == BNVal ->
             {BucketProps, []};
         _ ->
-            Error = ?FMT("Bucket type n_val ~p must match the associated "
+            Error = ?FMT("Bucket n_val ~p must match the associated "
                          "search_index ~s n_val ~p", [BNVal,Index,INVal]),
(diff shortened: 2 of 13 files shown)


rzezeski:
- [x] code review
- [x] compile/eunit
- [x] dialyzer
- [x] riak tests
- [x] manual test

riak_test/yokozuna_essential.erl
```diff
@@ -269,19 +269,18 @@ setup_indexing(Cluster, PBConns, YZBenchDir) ->
...
     ok = yz_rt:create_index(Node, <<"escaped">>),
-    yz_rt:set_index(Node, {?BUCKET_TYPE, <<"escaped">>}, <<"escaped">>),
-
     ok = yz_rt:create_index(Node, <<"unique">>),
+
     [yz_rt:wait_for_index(Cluster, I)
-     || I <- [?INDEX, <<"tagging">>, <<"escaped">>, <<"unique">>]].
+     || I <- [?INDEX, <<"tagging">>, <<"escaped">>, <<"unique">>]],
+
+    ok = rpc:call(Node, riak_core_bucket, set_bucket,
```
rzezeski: This `rpc:call` shouldn't be needed anymore now that you are passing `NVal` to `set_index`.

src/yz_bucket_validator.erl
```diff
@@ -0,0 +1,98 @@
...
+            %% since we cannot remove n_val
+            {proplists:delete(search_index, BucketProps),
+                [{search_index, Error}]}
+    end.
+
+-spec get_search_index_info(props()) -> {error,binary()} | {index_name(),n()}.
+get_search_index_info(Props) ->
+    case proplists:get_value(search_index, Props) of
+        undefined ->
+            {error, no_search_index};
```
rzezeski, line 79: This doesn't match your type spec:

```
yz_bucket_validator.erl:51: The pattern {'error', 'no_search_index'} can never match the type {'error' | binary(),binary() | pos_integer()}
```

src/yz_bucket_validator.erl
```diff
@@ -0,0 +1,98 @@
...
+                false ->
+                    {error, ?FMT("~s does not exist", [Index])};
+                true ->
+                    {Index, index_n_val(Index)}
+            end
+    end.
+
+%% @doc Get search_index from the bucket props, and return the
+%%      index's n_val. If it doesn't exist, return undefined.
+% index_n_val(Props) ->
```
rzezeski: Commented out code, also add `@priavte`.

src/yz_bucket_validator.erl
```diff
@@ -0,0 +1,98 @@
...
+            Error = ?FMT("Bucket type n_val ~p must match the associated "
+                         "search_index ~s n_val ~p ", [BNVal,Index,INVal]),
+            %% Removing search_index from valid props,
+            %% since we cannot remove n_val
+            {proplists:delete(search_index, BucketProps),
+                [{search_index, Error}]}
+    end.
+
+-spec get_search_index_info(props()) -> {error,binary()} | {index_name(),n()}.
+get_search_index_info(Props) ->
```
rzezeski, line 76: Add `@private`.

src/yz_bucket_validator.erl
```diff
@@ -0,0 +1,98 @@
...
+            {proplists:delete(search_index, BucketProps),
+                [{search_index, Msg}]};
+        {Index, INVal} ->
+            BNVal = proplists:get_value(n_val, Props),
+            validate_n_val(Index, INVal, BNVal, BucketProps)
+    end.
+
+%% @private
+%%
+
```
rzezeski: Dat whitespace.

src/yz_bucket_validator.erl
```diff
@@ -0,0 +1,98 @@
...
+-define(FMT(S, Args), list_to_binary(lists:flatten(io_lib:format(S, Args)))).
+
+-type prop() :: {PropName::atom(), PropValue::any()}.
+-type error() :: {PropName::atom(), ErrorReason::atom()}.
+-type props() :: [prop()].
+-type errors() :: [error()].
+
+%% @doc Performs two validations. The first is validating that an index
+%% exists before a bucket type can be associated to it by setting search_index.
```
rzezeski: This applies to buckets in general, not just bucket types, right?

rzezeski:
This doesn't seem to actually be working. Here are steps from manual
test I did:

```
# curl -X PUT 'http://localhost:10018/search/index/foo'

# curl 'http://localhost:10018/search/index/foo' | json_pp
{
   "n_val" : 3,
   "name" : "foo",
   "schema" : "_yz_default"
}

# curl -i -X PUT -H 'content-type: application/json' 'http://localhost:10018/types/data/buckets/foo/props' -d '{"props":{"search_index":"foo"}}'

# curl 'http://localhost:10018/types/data/buckets/foo/props' | json_pp
{
   "props" : {
      ...
      "n_val" : 3,
      ...
      "search_index" : "foo",
   }
}

# curl -i -X PUT -H 'content-type: application/json' 'http://localhost:10018/types/data/buckets/foo/props' -d '{"props":{"n_val":4}}'
HTTP/1.1 204 No Content
Vary: Accept-Encoding

// What? That shouldn't have work.

# curl 'http://localhost:10018/types/data/buckets/foo/props' | json_pp
{
   "props" : {
      ...
      "search_index" : "foo",
      ...
      "n_val" : 4,
      ...
   }
}

// This should not have been possible. There is nothing in the logs to
// indicate error. Then for heck of it I try again with n_val:5

# curl -i -X PUT -H 'content-type: application/json' 'http://localhost:10018/types/data/buckets/foo/props' -d '{"props":{"n_val":5}}'
HTTP/1.1 400 Bad Request

// Okay that one failed, I check the logs:

==> /root/yz-verify/rt/riak_yz/dev/dev1/log/console.log <==
2014-02-14 17:02:24.472 [error] <0.1688.0>@riak_core_bucket:set_bucket:77 Bucket properties validation failed [{search_index,<<"Bucket type n_val 4 must match the associated search_index foo n_val 3 ">>}]

// Interesting so even though I set to 5 it's reading the current
// value of 4. Something strange is happening with this validator code.
```

We should add two tests to `yz_index_admin`.
1. That non-existing index cannot be associated to bucket.
2. That n_val must match of the set property should fail.

src/yz_bucket_validator.erl
```diff
@@ -0,0 +1,98 @@
...
+        {ExistingProps, undefined} ->
+            ExistingProps;
+        {ExistingProps, BucketProps} ->
+            riak_core_bucket_props:merge(ExistingProps, BucketProps)
+    end,
+    case get_search_index_info(Props) of
+        {error, no_search_index} ->
+            {BucketProps, []};
+        {error, Msg} ->
+            {proplists:delete(search_index, BucketProps),
```
rzezeski: I haven't read up on how validators work so I may be misunderstanding what this code does but I hope this is not deleting the `search_index` property from the bucket properties. If the validation fails it should fail the user request and not mutate the stored properties.

coderoshi: A tuple of {valid properties, error messages} are to be returned. We aren't mutating the properties here, we're only removing search_index from the list of acceptable ones. See `riak_kv_bucket:validate/4`.

rzezeski:
Riak Tests pass and manual test works:

```
curl -i -X PUT -H 'content-type: application/json' 'http://localhost:10018/types/data/buckets/foo/props' -d '{"props":{"n_val":4}}'
HTTP/1.1 400 Bad Request
Vary: Accept-Encoding
Date: Sat, 15 Feb 2014 01:14:53 GMT
Content-Type: application/json
Content-Length: 82

{"n_val":"Bucket type n_val 4 must match the associated search_index foo n_val 3"}
```

I think the error message should be updated to just say "Bucket" but other than that looks good.

+1 to merge.

Issue #306, opened 2014-02-13
Cyclic behaviour when creating yokozuna index with broken schema


oleksiyk:
1) create a schema (`RpbYokozunaSchemaPutReq`), succeeds:

Notice the schema doesn't have a `text` field which is by default is `defaultSearchField`

``` xml
<?xml version="1.0" encoding="UTF-8" ?>
<schema name="test" version="1.5">
  <fields>
    <field name="num" type="long" indexed="true" stored="true"/>

    <field name="_yz_id" type="_yz_str" indexed="true" stored="true" required="true" />
    <field name="_version_" type="long" indexed="true" stored="true"/>
    <field name="_yz_ed" type="_yz_str" indexed="true" stored="true"/>
    <field name="_yz_pn" type="_yz_str" indexed="true" stored="true"/>
    <field name="_yz_fpn" type="_yz_str" indexed="true" stored="true"/>
    <field name="_yz_vtag" type="_yz_str" indexed="true" stored="true"/>
    <field name="_yz_node" type="_yz_str" indexed="true" stored="true"/>
    <field name="_yz_rb" type="_yz_str" indexed="true" stored="true"/>
    <field name="_yz_rk" type="_yz_str" indexed="true" stored="true"/>
    <field name="_yz_rt" type="_yz_str" indexed="true" stored="true"/>
    <field name="_yz_err" type="_yz_str" indexed="true"/>
  </fields>

  <uniqueKey>_yz_id</uniqueKey>

  <types>
    <fieldType name="_yz_str" class="solr.StrField" sortMissingLast="true" />
    <fieldType name="long" class="solr.TrieLongField" precisionStep="0" positionIncrementGap="0"/>
  </types>
</schema>
```

2) create index (`RpbYokozunaIndexPutReq`), no error returned: 

``` javascript
index: {
      name: 'test-search-index-1392332139243',
      schema: 'test-search-schema-1392332139243' // the schema from step 1
}
```

Solr reports the following:

```
2014-02-14 00:55:40,803 [WARN] <qtp655787180-23>@SolrCore.java:503 [test-search-index-1392332139243] Solr index directory '/usr/local/Cellar/riak/2.0.0-pre11/libexec/data/yz/test-search-index-139233213924
3/data/index' doesn't exist. Creating new index...
2014-02-14 00:55:43,506 [ERROR] <qtp655787180-16>@SolrException.java:108 org.apache.solr.common.SolrException: undefined field text
```

[...]

PR #307, opened 2014-02-14
Add validator for required text field

coderoshi:
Related to #306, the default text field is required. This adds a text validator to enforce it exists.


rzezeski:
We may be able to avoid requiring this field. Issue #306 makes me wonder about a few of our settings in `solrconfig.xml`.

PR #308, opened 2014-02-14
More AAE ports

rzezeski:
More AAE ports from basho/riak_kv#765.
- Allow use of `never` for value of `anti_entropy_expire` to indicate
  that trees should never expire automatically (they can still be
  expired manually).
- Clear any pending AAE exchanges when switching from automatic to
  manual mode. Active exchanges will be allowed to finish. The reason
  for clearing pending exchanges is because manual mode is about
  letting the user control what exchanges happen and thus all pending
  automatic exchanges should be dropped.
- Modify AAE trees to delay clearing trees up until the moment before
  rebuild. Thus when trees expire they are eligible for exchange up
  until the moment they are rebuilt. This is a big improvement from
  before which would immediately clear (delete) the tree upon
  expiration and could mean hours or days of time before that tree
  could be exchanged again thus creating a window of no AAE.
- Add ability to manually expire all trees at once via
  `yz_entropy_mgr:expire_trees/0`. This may be useful in a production
  scenario where the trees are no longer trusted but you still want to
  be able to exchange before they are rebuilt. It may also be useful
  for testing.
- Add a test to verify that trees can still be exchanged while they
  are expired but before they are rebuilt.

### DIFF
--- a/riak_test/aae_test.erl
+++ b/riak_test/aae_test.erl
@@ -80,6 +80,8 @@ confirm() ->
 
     _ = verify_no_repair_after_restart(Cluster),
 
+    _ = verify_exchange_after_expire(Cluster),
+
     pass.
 
 %% @doc Create a tuple containing a riak object and the first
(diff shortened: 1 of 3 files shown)


riak_test/aae_test.erl
```diff
@@ -145,6 +147,29 @@ setup_index(Cluster, PBConn, YZBenchDir) ->
...
+    lager:info("Expire all trees and verify exchange still happens"),
+    lager:info("Set anti_entropy_build_limit to 0 so that trees can't be built"),
+    _ = [ok = rpc:call(Node, application, set_env, [riak_kv, anti_entropy_build_limit, {0, 1000}])
+         || Node <- Cluster],
+
+    lager:info("Expire all trees"),
+    _ = [ok = rpc:call(Node, yz_entropy_mgr, expire_trees, [])
+         || Node <- Cluster],
+
+    %% The expire is done very cast so just give it a moment
```
coderoshi: very cast. such spelling. wow

src/yz_entropy_mgr.erl
```diff
@@ -125,6 +125,16 @@ cancel_exchanges() ->
 clear_trees() ->
     gen_server:cast(?MODULE, clear_trees).
 
+%% @doc Expire all the trees. Expired trees can still be exchanged up
+%% until the point they are rebult versus clearing trees which
```
coderoshi: rebuilt

rzezeski:
Forgot to link this to #253.

riak_test/aae_test.erl
rzezeski: LOLOL.

src/yz_index_hashtree.erl
```diff
@@ -514,23 +530,62 @@ close_trees(S=#state{trees=Trees, closed=false}) ->
...
+            S2#state{built=Pid};
+        _ ->
+            Pid ! stop,
+            S
+    end;
+maybe_rebuild(S) ->
+    S.
+
+-spec get_all_locks(build | rehash, pid()) -> boolean().
+get_all_locks(Type, Pid) ->
```
coderoshi, line 583: This is a strangely named function. It's only getting one lock, not all.

rzezeski, line 583: I use this name to match KV AAE. A lot of this code is strange but want to keep it as close to KV as possible until these two systems merge. It makes porting patches easier.

coderoshi:
+1
`aae_test-bitcask: pass`

PR #304, opened 2014-02-13
Add stats formatting for riak-admin and http stats output

bowrocker:
YZ stats are not currently displayed on the riak http stats endpoint, or by riak-admin status. Add YZ stats formatting appropriate for console output with 'riak-admin status' and the htts /stats endpoint. The stats namespace for Search is transformed to /search/[query | index]/... instead of 'yokozuna'.

This PR also depends on changes to riak_kv in order to publish the stats to the 2 interfaces:
https://github.com/basho/riak_kv/pull/839

Example stats output:

search_stat_ts : 1392310824
search_query_fail_count : 0
search_query_fail_one : 0
search_index_throughput_count : 0
search_index_throughput_one : 0
search_index_fail_count : 0
search_index_fail_one : 0
search_query_throughput_count : 0
search_query_throughput_one : 0
search_query_latency_min : 0.0
search_query_latency_max : 0.0
search_query_latency_median : 0.0
search_query_latency_percentile_95 : 0.0
search_query_latency_percentile_99 : 0.0
search_index_latency_min : 0.0
search_index_latency_max : 0.0
search_index_latency_median : 0.0
search_index_latency_percentile_95 : 0.0
search_index_latency_percentile_99 : 0.0

PR #311, opened 2014-02-18
Update previous hash during event tick + general refactoring

rzezeski:
Currently `yz_events:sync_indexes` is called on every tick because the previous hash is not being updated. I noticed this when investigating #304.
- Sync indexes IFF full check or hash has changed.
- Fix off-by-one error in full check.
- Add `MAYBE` macro for tighter code.

This can be verified by using redbug to verify `sync_indexes` is called every 60 seconds.

```
> redbug:start("yz_events:sync_indexes -> return", [{time, 180000}]).
```

### DIFF
--- a/include/yokozuna.hrl
+++ b/include/yokozuna.hrl
@@ -127,6 +127,16 @@
 
 -define(DATA_DIR, application:get_env(riak_core, platform_data_dir)).
 
+
+-define(MAYBE(Check, Expression, Default),
+        case Check of
+            true -> Expression;
+            false -> Default
+        end).
+
+-define(MAYBE(Check, Expression),
+        ?MAYBE(Check, Expression, ok)).
+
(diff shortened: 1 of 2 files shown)


include/yokozuna.hrl
```diff
@@ -127,6 +127,16 @@
...
 
+
+-define(MAYBE(Check, Expression, Default),
+        case Check of
+            true -> Expression;
+            false -> Default
+        end).
+
+-define(MAYBE(Check, Expression),
+        ?MAYBE(Check, Expression, ok)).
```
coderoshi, line 138: Welcome to CPP ;)

coderoshi:
+1, if you're into this kind of macro thing. Tested and works.

rzezeski:
Gah, I mean to link this is issue #306, not 304.

PR #313, opened 2014-02-19
Don't require the 'text' field

rzezeski:
The fact that the `text` field was being set as `df` by the select handler was just artifact of copying Solr example and not paying close enough attention. This removes that mapping so the user must either explicitly use a field for every term or pass the `df` request parameter.
- Use `_yz_id` for the ping handler.
- Remove `df` default from select handler.
- Update basho bench driver.
- Update migration test.

This is related to #306.


coderoshi:
+1, `17 Tests Passed`, Closing #307

Issue #314, opened 2014-02-19
Build custom handlers once and reuse JAR


rzezeski:
Currently the custom handlers are build from scratch every time. This
means byte code for the local JDK version will be generated. Recently
we had an issue where a user couldn't start the 2.0.0pre11 Fedora
package because the official Fedora builder has javac 1.8.0-internal
and the user has JDK 1.7.

http://lists.basho.com/pipermail/riak-users_lists.basho.com/2014-February/014703.html

The fix is to do like we do with the custom Solr package and build the
`yokozuna.jar` separately from the yokozuna Erlang application. The
JAR would be downloaded from s3 like our custom Solr version and put
it the right place. This has the added benefit of making the Yokozuna
build simpler as it will no longer require `javac` on the PATH.

As Solr is compile with JDK 1.6 so should this be. Don't forget to
version the JAR so updates don't overwrite previous versions.

PR #315, opened 2014-02-20
Download custom jars from s3

rzezeski:
This should address #314. See that PR or commit msgs for more info.

In order to test this you should run `make clean && rm -rf deps` to make sure you get latest. You'll also want to wipe out your verify dir.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -17,7 +17,7 @@ clean:
 	$(REBAR) clean
 	rm -rf riak_test/ebin
 	rm -rf build
-	git clean -dfx priv/solr
+	git clean -dfx priv/
 
 distclean: clean
 	$(REBAR) delete-deps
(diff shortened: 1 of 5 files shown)


bowrocker:
Works for me: https://gist.github.com/bowrocker/9117066

Subsequent compiles without 'make clean' do not download them again. This is a good thing, but what happens if the jars change? Do we have to remember to do a clean to update? If so, perhaps a follow on is to check and track a checksum, and download only if it changes?

rzezeski:
Actually they should be downloaded again after clean because the files are not cached like the custom Solr package. They are small so hopefully that is fine for now.

The jar are versioned and the only way they change is if the scripts are updated to a new version. To get the new jars you'll want to run a make clean followed by make.

bowrocker:
Makes sense to me, just checking. Things like that have bitten me in the past.

I am :+1:

rzezeski:
Yea this is why I added the version. The yokozuna-1.jar will never change. If we add or modify handlers that will require the following steps:
1. Build yokozuna-2.jar.
2. Upload to s3.
3. Set `YZ_JAR_VSN=2` in `grab-solr.sh`.

PR #325, opened 2014-02-26
tools.mk update and dialyzer fixes

reiddraper:


reiddraper:
There are a couple of outstanding errors left, one of the most interesting is with the `tree()` type, which include says should be a `pid()`, but inside `riak_kv_index_hashtree:get_build_time/1`, it expects a record, or something. Worth investigating.

coderoshi:
+1 Along with #327, dialyzer is happy, the changes look good

reiddraper:
_I'm going to push another commit to update tools.mk, so please avoid merging until then._

Vagabond:
I updated tools.mk to fix another bug we found in tools.mk.

reiddraper:
+1

PR #333, opened 2014-03-05
Refactor yz_cover:calc_plan

bowrocker:
Some light refactoring of yz_cover:calc_plan and related functions for readability.

Eunits and riak_tests pass.

### DIFF
--- a/src/yz_cover.erl
+++ b/src/yz_cover.erl
@@ -145,59 +145,61 @@ cache_plan(Index, Ring) ->
 %% @doc Calculate a plan for the `Index'.
 -spec calc_plan(index_name(), ring()) -> {ok, plan()} | {error, term()}.
 calc_plan(Index, Ring) ->
-    [NumPartitions, NVal, Result] = create_plan(Index, Ring),
-    handle_plan(Result, Ring, NVal, NumPartitions).
+    NumPartitions = riak_core_ring:num_partitions(Ring),
+    NVal = yz_index:get_n_val(yz_index:get_index_info(Index)),
+    CoveragePlan = create_coverage_plan(NVal),
+    maybe_filter_plan(CoveragePlan, Ring, NVal, NumPartitions).
 
 %% @private
 %%
-%% @doc Handle the result of create plan, for return to caller
--spec handle_plan(term(), ring(), integer(), integer()) ->  {ok, plan()} | {error, term()}.
-handle_plan({error, Error}, _, _, _) ->
+%% @doc Filter plan or error
+-spec maybe_filter_plan(term(), ring(), integer(), integer()) ->  {ok, plan()} | {error, term()}.
+maybe_filter_plan({error, Error}, _, _, _) ->
     Error;
-handle_plan({CoverSet, _}, Ring, NVal, NumPartitions) ->
-    [UniqNodes, LogicalCoverSet] = calc_plan_add_filtering(CoverSet, Ring, NVal, NumPartitions),
+maybe_filter_plan({CoverSet, _}, Ring, NVal, NumPartitions) ->
+    LogicalCoverSet = calc_plan_add_filtering(CoverSet, Ring, NVal, NumPartitions),
+    UniqNodes = get_uniq_nodes(CoverSet),
     Mapping = yz_solr:build_mapping(UniqNodes),
-    calc_result(length(Mapping) == length(UniqNodes), UniqNodes, LogicalCoverSet, Mapping).
+    calc_plan_return(length(Mapping) == length(UniqNodes), UniqNodes, LogicalCoverSet, Mapping).
 
 %% @private
 %%
 %% @doc handle the result of the calculation result with the 
 %%      result of filtering
--spec calc_result(true | false, list(), logical_cover_set(), list()) ->  {ok, plan()} | {error, term()}.
-calc_result(false, _, _, _) ->
+-spec calc_plan_return(true | false, list(), logical_cover_set(), list()) ->  {ok, plan()} | {error, term()}.
+calc_plan_return(false, _, _, _) ->
     {error, "Failed to determine Solr port for all nodes in search plan"};
-calc_result(true, UniqNodes, LogicalCoverSet, Mapping) ->
+calc_plan_return(true, UniqNodes, LogicalCoverSet, Mapping) ->
     {ok, {UniqNodes, LogicalCoverSet, Mapping}}.
+
+%% @private
+%%
+%% @doc return unique list of nodes
+-spec get_uniq_nodes(logical_cover_set()) -> list().
+get_uniq_nodes(CoverSet) ->
+    {_Partitions, Nodes} = lists:unzip(CoverSet),
+    lists:usort(Nodes).
 
 %% @private
 %%
 %%  @doc add filtering with the NVal, NumPartitionsand CoverSet and return
 %% 
--spec calc_plan_add_filtering(logical_cover_set(), ring(), integer(), integer()) -> list().
+-spec calc_plan_add_filtering(logical_cover_set(), ring(), integer(), integer()) -> logical_cover_set().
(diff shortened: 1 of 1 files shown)


src/yz_cover.erl
```diff
@@ -145,33 +145,61 @@ cache_plan(Index, Ring) ->
...
+-spec calc_plan_add_filtering(logical_cover_set(), ring(), integer(), integer()) -> logical_cover_set().
+calc_plan_add_filtering(CoverSet, Ring, NVal, NumPartitions) ->
+    LPI = logical_index(Ring),
+    add_filtering(NVal, NumPartitions, LPI, CoverSet).
+    
+%% @private
+%%
+%% @doc create coverage plan
+-spec create_coverage_plan(integer()) -> term().
+create_coverage_plan(NVal) ->
```
rzezeski, line 157: While moving this functionality you got rid of `Selector` and `NumPrimaries` but I think it helps to have those variables as it gives a name to the values.

src/yz_cover.erl
```diff
@@ -145,33 +145,61 @@ cache_plan(Index, Ring) ->
...
+%% 
+-spec calc_plan_add_filtering(logical_cover_set(), ring(), integer(), integer()) -> logical_cover_set().
+calc_plan_add_filtering(CoverSet, Ring, NVal, NumPartitions) ->
+    LPI = logical_index(Ring),
+    add_filtering(NVal, NumPartitions, LPI, CoverSet).
+    
+%% @private
+%%
+%% @doc create coverage plan
+-spec create_coverage_plan(integer()) -> term().
```
rzezeski: Use `n()`.

src/yz_cover.erl
```diff
@@ -145,33 +145,61 @@ cache_plan(Index, Ring) ->
...
-                    {ok, {UniqNodes, LogicalCoverSet, Mapping}};
-                false ->
-                    {error, "Failed to determine Solr port for all nodes in search plan"}
-            end
-    end.
+%% @private
+%%
+%% @doc Filter plan or error
+-spec maybe_filter_plan(term(), ring(), integer(), integer()) ->  {ok, plan()} | {error, term()}.
+maybe_filter_plan({error, Error}, _, _, _) ->
```
rzezeski: This is pulling the `Error` out of the tuple which breaks the spec.

src/yz_cover.erl
```diff
@@ -145,33 +145,61 @@ cache_plan(Index, Ring) ->
...
-                true ->
-                    {ok, {UniqNodes, LogicalCoverSet, Mapping}};
-                false ->
-                    {error, "Failed to determine Solr port for all nodes in search plan"}
-            end
-    end.
+%% @private
+%%
+%% @doc Filter plan or error
+-spec maybe_filter_plan(term(), ring(), integer(), integer()) ->  {ok, plan()} | {error, term()}.
```
rzezeski: Use `n()` for `NVal` and `pos_integer()` for NumPartitions.

src/yz_cover.erl
```diff
@@ -145,33 +145,61 @@ cache_plan(Index, Ring) ->
...
+%%
+%% @doc return unique list of nodes
+-spec get_uniq_nodes(logical_cover_set()) -> list().
+get_uniq_nodes(CoverSet) ->
+    {_Partitions, Nodes} = lists:unzip(CoverSet),
+    lists:usort(Nodes).
+
+%% @private
+%%
+%%  @doc add filtering with the NVal, NumPartitionsand CoverSet and return
```
rzezeski: Extra space before `@doc` as well as unnecessary line following it.

src/yz_cover.erl
```diff
@@ -145,33 +145,61 @@ cache_plan(Index, Ring) ->
...
+%% 
+-spec calc_plan_add_filtering(logical_cover_set(), ring(), integer(), integer()) -> logical_cover_set().
+calc_plan_add_filtering(CoverSet, Ring, NVal, NumPartitions) ->
```
rzezeski: Rename this to `make_logical_and_filter`. The plan as already been calculated. This function is simply transforming it.

src/yz_cover.erl
rzezeski: A better doc would be `This function converts CovertSet into logical partitions and adds filtering information.`

src/yz_cover.erl
```diff
@@ -145,33 +145,61 @@ cache_plan(Index, Ring) ->
...
+-spec calc_plan_return(true | false, list(), logical_cover_set(), list()) ->  {ok, plan()} | {error, term()}.
+calc_plan_return(false, _, _, _) ->
+    {error, "Failed to determine Solr port for all nodes in search plan"};
+calc_plan_return(true, UniqNodes, LogicalCoverSet, Mapping) ->
+    {ok, {UniqNodes, LogicalCoverSet, Mapping}}.
+
+%% @private
+%%
+%% @doc return unique list of nodes
+-spec get_uniq_nodes(logical_cover_set()) -> list().
```
rzezeski: `-> [node()]`

src/yz_cover.erl
```diff
@@ -145,33 +145,61 @@ cache_plan(Index, Ring) ->
...
+    LogicalCoverSet = calc_plan_add_filtering(CoverSet, Ring, NVal, NumPartitions),
+    UniqNodes = get_uniq_nodes(CoverSet),
+    Mapping = yz_solr:build_mapping(UniqNodes),
+    calc_plan_return(length(Mapping) == length(UniqNodes), UniqNodes, LogicalCoverSet, Mapping).
+
+%% @private
+%%
+%% @doc handle the result of the calculation result with the 
+%%      result of filtering
+-spec calc_plan_return(true | false, list(), logical_cover_set(), list()) ->  {ok, plan()} | {error, term()}.
```
rzezeski: Use `boolean()` and `[node()]`.

rzezeski: Rename this function to `plan_return`. It's being used inside `maybe_filter_plan` at which point no calculation is happening.

rzezeski:
Overall, this is a good refactoring. I made some specific comments on the code but have two general ones as well:
1. All `@doc` should be full sentences.
2. All the private functions should be moved into the private section and kept in alphabetical order.

src/yz_cover.erl
```diff
@@ -245,6 +244,26 @@ make_cover_set(N, Q, Cover) ->
 make_logical(LogicalIndex, PSet) ->
     [{logical_partition(LogicalIndex, P), Node} || {P, Node} <- PSet].
 
+%% @private
+%%
+%% @doc Add filtering to the cover set and return a logical cover set.
+-spec make_logical_and_filter(logical_cover_set(), ring(), integer(), integer()) -> logical_cover_set().
```
rzezeski: NVal should be `n()` and NumPartitions should be `pos_integer()`.

src/yz_cover.erl
```diff
@@ -184,6 +175,14 @@ get_distance(Q, {LPA,_}, {LPB,_}) when LPB < LPA ->
 get_distance(_Q, {LPA,_}, {LPB,_}) ->
     LPB - LPA.
 
+%% @private
+%%
+%% @doc This function converts CovertSet into logical partitions and adds filtering information.
```
rzezeski: This doc comment was meant for `make_logical_and_filter`.

src/yz_cover.erl
```diff
@@ -253,6 +272,15 @@ partition(LogicalIndex, LP) ->
 
 %% @private
 %%
+%% @doc Handle the result of the calculation result with the result of filtering.
```
rzezeski: This doc doesn't really tell me anything. It would probably be better to a) have no doc at all or b) maybe something like "Return the plan only if there exists a Solr host-port mapping for each node in the plan.".

bowrocker:
https://github.com/basho/yokozuna/pull/339

PR #337, opened 2014-03-07
Disallow bad schema or index names

coderoshi:
Currently, yz allows names that contain a `/` char, which breaks things terribly when yz and solr attempt to create corresponding files/directories. Disallow it.

### DIFF
--- a/riak_test/yz_pb.erl
+++ b/riak_test/yz_pb.erl
@@ -35,7 +35,9 @@ confirm() ->
     Cluster = rt:build_cluster(4, ?CFG),
     rt:wait_for_cluster_service(Cluster, yokozuna),
     confirm_admin_schema(Cluster),
+    confirm_admin_bad_schema_name(Cluster, <<"bad/name">>),
     confirm_admin_index(Cluster),
+    confirm_admin_bad_index_name(Cluster, <<"bad/name">>),
     confirm_basic_search(Cluster),
     confirm_encoded_search(Cluster),
     confirm_multivalued_field(Cluster),
@@ -129,6 +131,16 @@ confirm_admin_schema(Cluster) ->
     riakc_pb_socket:stop(Pid),
     ok.
 
+confirm_admin_bad_schema_name(Cluster, Name) ->
+    Name = <<"bad/name">>,
+    Node = select_random(Cluster),
+    {Host, Port} = yz_rt:riak_pb(hd(rt:connection_info([Node]))),
+    lager:info("confirm_admin_bad_schema_name ~s [~p]", [Name, {Host, Port}]),
+    {ok, Pid} = riakc_pb_socket:start_link(Host, Port),
+    {error,_} = riakc_pb_socket:create_search_schema(Pid, Name, ?SCHEMA_CONTENT),
+    riakc_pb_socket:stop(Pid),
+    ok.
+
 confirm_admin_index(Cluster) ->
     Index = <<"index">>,
     create_index(Cluster, Index, Index, 4),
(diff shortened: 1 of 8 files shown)


riak_test/yz_pb.erl
```diff
@@ -129,6 +131,16 @@ confirm_admin_schema(Cluster) ->
     riakc_pb_socket:stop(Pid),
     ok.
 
+confirm_admin_bad_schema_name(Cluster, Name) ->
+    Name = <<"bad/name">>,
```
rzezeski, line 135: This was passed in as argument and set in the function. Doesn't break since the values are the same but might as well not bother taking the `Name` as an argument.

src/yz_index.erl
```diff
@@ -252,6 +261,15 @@ remove_non_owned_data(Index, Ring) ->
 schema_name(Info) ->
     Info#index_info.schema_name.
 
+%% @doc Verify that the index is a name that Solr can use. Some chars
+%%      are invalid, namely "/".
+-spec verify_name(index_name()) -> {ok, index_name()} | {error, invalid_name}.
+verify_name(Name) ->
```
rzezeski, line 267: Is there any point to returning `{ok, Name}` when the caller already knows the name it passed it?

coderoshi, line 267: I'm following the same pattern set by the yz_schema verify functions.

src/yz_pb_admin.erl
```diff
@@ -114,7 +114,9 @@ process(#rpbyokozunaindexputreq{
         ok ->
             {reply, #rpbputresp{}, State};
         {error, schema_not_found} ->
-            {error, "Schema not found", State}
+            {error, "Schema not found", State};
+        {error, invalid_name} ->
+            {error, "Invalid character '/' in index name", State}
```
rzezeski, line 119: This error is assuming that `invalid_name` indicates the existence of `/`. That is true the code as it stands today but if the verification function has characters added to then this message will have to be changed with it. Good chance for human error. Perhaps the verification function should return something more like `{error, {invalid_name, "Invalid character..."}}` so that the message is part of the error tuple.

I'm fine with leaving it as is given other things that need to get done and that, from a user perspective, there is no difference.

rzezeski:
- [x] compile/unit
- [x] dialyzer
- [x] riak tests
- [x] code review
- [x] manual check

rzezeski:
All the riak tests passed but the migration test but I've been seeing that fail on other branches so I don't think it should hold up this PR.

```
yz_rs_migration_test-bitcask : fail
1 Tests Failed
16 Tests Passed
```

rzezeski:
Manual check of HTTP schema create:

```
# curl -i  -XPUT -H'content-type: application/xml' 'http://localhost:10018/search/schema/foo%2Fbar' --data-binary @/root/good-schema.xml
HTTP/1.1 100 Continue

HTTP/1.1 400 Bad Request
Date: Mon, 10 Mar 2014 17:51:59 GMT
Content-Type: text/plain
Content-Length: 59

Error storing schema: Invalid character '/' in schema name
```

Manual check for PB schema create:

```
1> {ok, B} = file:read_file("/root/good-schema.xml").
{ok,<<"<?xml version=\"1.0\" encoding=\"UTF-8\" ?>\n<!--\n Licensed to the Apache Software Foundation (ASF) under one or "...>>}
2> {ok,C} = riakc_pb_socket:start("localhost", 10017).
{ok,<0.35.0>}
3> riakc_pb_socket:create_search_schema(C, <<"foo/bar">>, B).
{error,<<"Error storing schema Invalid character '/' in schema name\n">>}
```

Manual check of HTTP index create:

```
# curl -i  -XPUT 'http://localhost:10018/search/index/foo%2Fbar'
HTTP/1.1 400 Bad Request
Date: Mon, 10 Mar 2014 18:00:39 GMT
Content-Type: text/plain
Content-Length: 44

Invalid character '/' in index name foo/bar
```

Manual check of PB index create:

```
4> riakc_pb_socket:create_search_index(C, <<"foo/bar">>).
{error,<<"Invalid character '/' in index name">>}
```

rzezeski:
+1 to merge.

PR #338, opened 2014-03-10
Add riak_test for ring resizing during YZ operations

bowrocker:
Add yz_ring_resizing, which:
- Creates a 4-node cluster, ring size 16
- Sets up indexing using the "fruit" schema
- Loads up some data with bb
- Fires up an asynchronous query
- Resizes ring to 64
- Waits for ring resize to finish and checks query results

Note this test has a ring size reduction commented out which should be re-instated once the issue with YZ and ring sizing downward is resolve.

### DIFF
--- a/riak_test/yz_ring_resizing.erl
+++ b/riak_test/yz_ring_resizing.erl
@@ -1,0 +1,141 @@
+-module(yz_ring_resizing).
+-compile(export_all).
+-include_lib("eunit/include/eunit.hrl").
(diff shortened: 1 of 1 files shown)


rzezeski:
+1 to merge
- [x] code review
- [x] compile/eunit
- [x] dialyzer
- [x] riak test

```
Test Results:
yz_ring_resizing-bitcask: pass
0 Tests Failed
1 Tests Passed
```

PR #347, opened 2014-03-20
tools to check and repair inconsistency between kv data and solr index

kuenishi:
In current implementation, there could be inconsistency between
kv data and solr index, in case of write-intensive workloads where
just a log "failed to index object ...." left in console.log.
This cannot be fixed by AAE because AAE only fixes inconsistency
between replicas, but not kv data and index.

Mainly used functions would be `yz_repair:check_type/1` and
`yz_repair:check_bucket/1` , each for bucket types and default buckets.
Both functions iterates over all vnodes along the cluster and compares
keys from vnode and keys from solr with `_yz_pn:N` where N is the
partition number.

Potential problems: Those functions runs sequencially, thus very slow
in large datasets. This is because correctness is the priority for now
and it is easy to make repair scripts work in parallel. Also keylisting
strategy is awful because both keylists from kv and solr are stored in
memory at once for subtraction. This is also a future work to seek for
and use smarter join algorithm.
## 

Writing riak_test is also future work. I'm also a newbie to yz dev, so any tiny nitpicks and tableflipping will be appreciated.

### DIFF
--- a/src/yz_repair.erl
+++ b/src/yz_repair.erl
@@ -1,0 +1,209 @@
+%% -------------------------------------------------------------------
+%%
+%% Copyright (c) 2014 Basho Technologies, Inc.  All Rights Reserved.
+%%
+%% This file is provided to you under the Apache License,
+%% Version 2.0 (the "License"); you may not use this file
+%% except in compliance with the License.  You may obtain
+%% a copy of the License at
+%%
+%%   http://www.apache.org/licenses/LICENSE-2.0
+%%
+%% Unless required by applicable law or agreed to in writing,
+%% software distributed under the License is distributed on an
+%% "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+%% KIND, either express or implied.  See the License for the
+%% specific language governing permissions and limitations
+%% under the License.
+%%
+%% -------------------------------------------------------------------
+
+-module(yz_repair).
+
+%% @doc a module to check yz consistency
+%% use this via attached console like this:
+%% > yz_repair:check_type(<<"some_buckettype">>)
+%% this finds a yokozuna index tied to bucket-types property
+%% or
+%% > yz_repair:check_type(<<"some_bucket">>)
+%% this finds a yokozuna index tied to bucket property
+
+%% note that this tool is not so much efficient, because
+%% this fetches all keys from memory ... in near future
+%% this script should be a vnode iterator that joins with
+%% keys from yokozuna popped out
+
+-export([check/1, repair/1,
+         check_type/1, repair_type/1,
+         check_bucket/1, repair_bucket/1,
+         check_vnode/4]).
+
+%% somewhat different from original bucket types,
+%% just for comprehensiveness in code below.
+-type bucket_or_type() :: bucket() | type().
+-type bucket() :: {bucket, binary()}.
+-type type() :: {type, binary()}.
+
+-spec check_type(binary()) -> ok.
+check_type(BucketType) ->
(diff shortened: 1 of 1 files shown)


bowrocker:
@kuenishi I quickly reviewed this with Ryan, and we had some questions:
- We are wondering why the failed to index errors are occurring the the first place, and would like to look into that more.
- @rzezeski feels that it should, in fact, be fixed by AAE -- perhaps it has not had a chance to kick in yet?
- Is it possible to post the data, logs, your configuration, and possibly some instructions, on how to re-create this problem?

Thank you very much.

kuenishi:
Thank you for your quick look, here're answers:
- I pasted one log into [My gist](https://gist.github.com/kuenishi/43dd212521ad4289611b) that indicated writing to  kv succeeded but indexing failed - that was the only suspicious log.
- I checked aae-status and AAE seemed working correctly. Or is there other way to check whether yz AAE working correctly? I think even if AAE is working correctly this could happen. If all 3 indexing request failed after putting to kv succeeded, then there could be a gap, kv has that key while solr doesn't. ..though this doesn't explain why search result varies.
- It was easy: just start a five-node stagedevrel, and write a 1-10KB JSON data with random key, ~10 thousands of keys. Backend is leveldb for 2i queries ( I believe the backend choice doesn't matter).

kuenishi:
one more: I chose default bucket type and default schema for reproduction. This also happened with default-schema-enabled bucket type in customer's test environment.

```
> http http://localhost:10018/buckets/fluentlog/props                                              
HTTP/1.1 200 OK
Content-Encoding: gzip
Content-Length: 262
Content-Type: application/json
Date: Thu, 20 Mar 2014 16:09:52 GMT
Vary: Accept-Encoding

{
    "props": {
        "allow_mult": false, 
        "basic_quorum": false, 
        "big_vclock": 50, 
        "chash_keyfun": {
            "fun": "chash_std_keyfun", 
            "mod": "riak_core_util"
        }, 
        "dw": "quorum", 
        "last_write_wins": false, 
        "linkfun": {
            "fun": "mapreduce_linkfun", 
            "mod": "riak_kv_wm_link_walker"
        }, 
        "n_val": 3, 
        "name": "fluentlog", 
        "notfound_ok": true, 
        "old_vclock": 86400, 
        "postcommit": [], 
        "pr": 0, 
        "precommit": [], 
        "pw": 0, 
        "r": "quorum", 
        "rw": "quorum", 
        "search_index": "fluentlog_index", 
        "small_vclock": 50, 
        "w": "quorum", 
        "young_vclock": 20
    }
}
```

rzezeski:
I'm marking this as 2.0-RC. I'm not actually sure we want the code patch yet. First we need to investigate and fully understand why the documents were failing to index in the first place and why AAE supposedly didn't repair them.

kuenishi:
It would be the best if AAE fixes it. Yes, we have to know whey indexing failed and also why AAE didn't repair it.

Here's the whole log: https://gist.github.com/kuenishi/9722017#file-console-log-0-L206
See around Line 206; right before the failure there's a AAE log that claims it fixed one key, though I'm not sure it's the one to be fixed. `solr.log` is also attached, but nothing worth mentioning as there are no logs around 3/19 13:08 - the time when it happened.

kuenishi:
The original issue in customer's environment was not this issue (my fault...) . I'm feeling like to close this issue. If this has reproduced please reopen this issue. I also leave this branch not deleted, because the gap-checking code `yz_repair:check_bucket/1` would be useful in future. Just in case. But I hope it never used again.

Issue #352, opened 2014-04-01
Don't log ops on index failure


rzezeski:
Don't log the ops being sent to Solr. Since lager will truncate overly
long messages it is better to not log the ops being sent to Solr. If
the doc is too large it causes the return from Solr to get truncated
and leaves no indication in the log of why Solr failed to index the
object.

This patch is pretty simple. Just remove the `Ops` variable from the
throw in `yz_solr:index`.

This is related to #347 which had a failure but the return body from
Solr was truncated from the log.

PR #354, opened 2014-04-01
Remove logging of ops sent to Solr when indexing fails.

bowrocker:
If the doc being logged is too large, lager will truncate it, leaving no indication why the actual index failure occurred. Remove the logging of it.

Issue:

https://github.com/basho/yokozuna/issues/352


rzezeski:
+1 to merge

PR #351, opened 2014-04-02
Normalize riak_test execution process

kellymclaughlin:
Currently the execution of the riak_test modules for yokozuna relies on some specific setup assumption contained in the `tools/verify.sh` script. The focus of this issue is to bring the execution of the yokozuna tests inline with way other riak_tests are executed. This seems appropriate now that yokozuna is a part of riak proper. This will make it easier for people to get the tests up and running and facilitate the testing of yokozuna features in concert with other riak features.

My initial thoughts of the things needed are as follows: 
- [x] Remove requirement of extra step to build the yokozuna basho_bench driver prior to executing riak_test modules.
- [x] Change the expectation for where the `yz_security` test looks for the cert file. 
- [x] Update basho_bench setup and expectations to break the reliance on `verify.sh` and `misc/bench` in the yokozuna repo
- [x] Update the TESTING.md document to reflect the new procedures to run riak_test with yokozuna

There may be other items or hurdles that crop up, but this issue is to represent the exploratory work to attempt this.

cc: @engelsanchez

### DIFF
--- a/docs/TESTING.md
+++ b/docs/TESTING.md
@@ -6,69 +6,44 @@ Testing
 Riak Test is a tool for running integration tests against a Riak
 cluster.  See the [Riak Test README][rt_readme] for more details.
 
-### Build a Riak/Yokozuna devrel
+Follow the instructions in the [Riak Test README][rt_readme] to build
+a `devrel` release and set it up for testing with riak_test.
 
-Make a directory to build the devrel.
+### Add Yokozuna Config
 
-    mkdir ~/testing
-    cd testing
+Open `~/.riak_test.config` and add the following to the configuration
+stanza to be used for yokozuna testing:
 
-The rest is like the [Yokozuna Getting Started][yz_gs] guide but
-checkout riak as `riak_yz`.  Don't start the cluster.  Just build the
-devrel.
+    {basho_bench, "<path-to-basho_bench-repo>"},
+    {yz_dir, "<path-to-yokozuna-repo>"},
 
-    git clone git://github.com/basho/riak.git riak_yz
-    cd riak_yz
-    make stagedevrel
+This will result in a configuration stanza similar to the following:
 
-### Setup rtdev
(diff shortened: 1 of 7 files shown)


coderoshi:
I originally had to clone `yz_security` to use yokozuna's priv dir, since the r_t priv dir wasn't passing for verify.sh tests. I don't recall the reason why. But the r_t security test have been fixed since then. Just some historical fyi.

coderoshi:
@kellymclaughlin following your instructions, the tests pass (after merging `develop`) except for `aae_test`. If you change the definition of `YZBenchDir` to use `yz_rt:bb_driver_setup` it works, and I'll +1 this (or I'll push my patch, whichever you like)

coderoshi:
+1 dea8f58

kellymclaughlin:
@borshop merge

rzezeski:
I'm not sure how we want to play it with bors when another repo causes a break.  To be pedantic I feel like #384 should be addressed first and have develop green again and then merge this. On the other hand we know these changes didn't cause the break.  @reiddraper given you are the bors/CI person what would you suggest?

kellymclaughlin:
I am fine with waiting for green. I can try to run down #384 this afternoon if nobody is already looking at it.

reiddraper:
> To be pedantic I feel like #384 should be addressed first and have develop green again and then merge this.

That would be my suggestion too. We'll get #384 addressed today if at all possible.

rzezeski:
So it turns out this branch broke tests relying on the `yz_driver`.  The compile calls are not passing the lager parse transform which is problematic because the driver contains calls to BB's `?DEBUG` macro which is a lager call.  When BB tries to run you get undefs and the job fails.  I have no idea how others were able to get their tests to pass.  I discovered this while rebasing a branch on develop so I'm just going to roll my fixes into that branch for sake of time.

Issue #330, opened 2014-03-04
yz/ibrowse configuration and interaction with Solr


wbrown:
Currentlly, Yokozuna uses the `ibrowse` application to stage index updates to Solr.  The default `ibrowse` settings are decidedly non-optimal with a default per host:port pair maximum of 10 sessions and 10 pipeline depth.

As the Solr indexes grow to millions of entries, the latency of the index updates exerts a backpressure upon Riak, slowing down key updates and PUTs.  `htop` shows a back and forth pattern of Riak and Solr exchanging CPU cycles.  At this point, Riak performance completely tanks, especially with the LevelDB backend.

Working with @evanmcc -- we found that the `ibrowse` author doesn't like to do configuration like the rest of the Erlang world, unfortunately.  He suggested a configuration line update, that I had to alter -- hooray for documentation that doesn't match reality.

`${RIAK_HOME}/lib/ibrowse-4.0.1/priv/ibrowse.conf` had the following line added:

```
{dest, "localhost", 8093, 100, 100000, []}.
```

Note that `{127,0,0,1}` in place of `localhost` did not work.  It really wants the string hostname, probably because `yz` is using `localhost`.  By the way, if you do this, this will default to IPV6 for localhost on some systems, which is an interesting and potentially problematic side effect.

The above settings set the connection maximum to `100` with a pipeline depth of `10000`.  I restarted Riak, and sure enough, I got that amount of connections which is an indicator that I may need to raise the limit higher for my throughput of about 400-500 keys indexed a second.

```
[root@altostratus log]#  netstat -anp | grep 8093 | grep EST | wc -l200
200
```

Checking for `TIME_WAIT` revealed hundreds of connections left over, which is indicative that there is no pipelining. 

```
[root@altostratus log]#  netstat -anp | grep 8093 | grep TIME_WAIT | wc -l
594
```

@coderoshi confirmed my suspicion.  This is problematic for deployments that index hundreds or thousands of keys a second with the overhead of establishing a socket connection for every index update.

So, I'd suggest the following changes:
- Add a more reasonable default max session value for `localhost:8093`
- Use HTTP pipelining so that we don't run out of socket descriptors/file handles and also so we don't incur the punishing latency and overhead of a TCP connection for every key update.

rzezeski:
[...]

Issue #316, opened 2014-02-21
Catch-all "*" in default schema XML is problematic


wbrown:
There are a couple of lines in the default schema that is problematic for people who are potentially just starting out in Riak.

https://github.com/basho/yokozuna/blob/develop/priv/default_schema.xml#L119-L120

``` XML
   <!-- catch-all field -->
   <dynamicField name="*" type="text_general" indexed="true" stored="false" multiValued="true" />
```

I was after a least-obstacle evaluation of Yokozuna, so I used the default fields and typed my fields using the dynamic typing.  I'd noticed that the Yokozuna database was larger than my Riak database, which was unusual and wasteful to me.

I tried turning off the `stored` parameters in the types that I was using, but saw no appreciable difference.

Finally, I noticed the catchall `*` and facepalmed.  Disabling that catchall reduced my Solr database size from gigabytes to dozens of megabytes, and mitigated the `no available vnode` issues that I was seeing with Bitcask.

I appreciate that you're kind of supposed to have your own schema in Solr, but this is error prone for getting started, and gives someone a false impression of Solr's efficiency.

rzezeski:
@wbrown Interesting, the reasoning for adding the catch-all field was so that beginners could search data without learning about the schema or using suffix fields (#95). It will be tough to satisfy everyone. Anyone running Yokozuna in production absolutely needs to create their own schemas that are tailored for the data being indexed.

I guess for me the question is:

Is it better for users to fail to index data and learn about the schema in a very explicit manner or is it better to implicitly index for "ease of use" when first experimenting with Yokozuna?

One thing I don't like about the current situation is it allows users to remain blissfully unaware that most/all of their data is being indexed as text when it shouldn't. If the catch-all was removed users would be forced to understand the mechanisms at play which is a good thing in the long run.

Pinging @coderoshi @jonmeredith @javajolt @bowrocker @tsantero @hectcastro @drewkerrigan @seancribbs @gburd and @michellep for any input on this.

coderoshi:
[...]

PR #358, opened 2014-04-09
Soften JVM overload failures

coderoshi:
After a week of replicating #330, I've concluded that there is little that we can do about an overloaded Solr JVM. What we can do is reduce the flood of failure messages by making some simple adjustments in key configuration files:
### solrconfig.xml

Suggested by http://wiki.apache.org/solr/SolrPerformanceProblems, slow or overly frequent commits can overload the cache warmer. Since all other `autowarmCount`s were set to 0, I followed suit with the `fieldValueCache`.

Since all cache auto warmers were set to zero, I also bumped up with default `maxWarmingSearchers` from 2 (suitable for read-only slaves, as mentioned in the config comments) to 4. I found that on high loads of large objects, I'd get an [exceeded limit of maxWarmingSearchers=X](http://wiki.apache.org/solr/FAQ#What_does_.22exceeded_limit_of_maxWarmingSearchers.3DX.22_mean.3F) message, which this resolved.
### default_schema.xml

For objects with many fields, I set the `catch-all field` to false by default, as per #316. This helped tremendously in the reduction of incidental field indexing.
### jetty.xml

I bumped up the `maxIdleTime` from 50 seconds to 60, since that's the default timeout for many of our PB clients. It's also important that this value matches the ibrowse client timeout, explained below.

The `acceptors` I kept at their default, which is 2. Solr recommends that you have 2 acceptors per CPU, so this is something that we should document. A better long term option is to make this a riak.conf, or possibly even automatically set acceptor count by interrogating the system, but I'm unsure this is something we want to tackle for beta.
### yz_solr.erl

Issue #320 was rife with `org.eclipse.jetty.io.EofException`. This is because ibrowse default times out at 30 seconds. Since jetty defaulted to 50 seconds, it was Solr complaining about the unexpected closed socket. This at least stops that complaint.
### ibrowse.conf

I made no changes to ibrowse.conf. Though we could increase the number of pooled and opened sockets, similar to this `{dest, "localhost", 10014, 100, 100000, []}.` it doesn't resolve the underlying problem of a slow and overloaded JVM. It might be a advanced setting in isolated cases, but in all of my tests increasing the connection pool just prolonged the inevitable crash from a JVM that wasn't indexing large objects fast enough.
### riak.conf

Followed many JVM tuning suggestions for larger heaps (http://wiki.apache.org/solr/ShawnHeisey#GC_Tuning), from tweaking pages sizes, settings per heap space, ratios, etc.

```
-d64 -Xms8g -Xmx32g -XX:+UseStringCache -XX:+UseCompressedOops -XX:NewRatio=3 \
-XX:SurvivorRatio=4 -XX:TargetSurvivorRatio=90 -XX:MaxTenuringThreshold=8  \
-XX:+UseConcMarkSweepGC -XX:+CMSScavengeBeforeRemark -XX:PretenureSizeThreshold=64m \
-XX:CMSFullGCsBeforeCompaction=1 -XX:+UseCMSInitiatingOccupancyOnly \
-XX:CMSInitiatingOccupancyFraction=70 -XX:CMSTriggerPermRatio=80 \
-XX:CMSMaxAbortablePrecleanTime=6000 -XX:+CMSParallelRemarkEnabled \
-XX:+ParallelRefProcEnabled -XX:+UseLargePages -XX:+AggressiveOpts
```

I found some values that allowed me to support slightly larger objects sizes on my test cluster, for my particular test set (large JSON objects, 1-2M with 40-10k indexes of various types, updated 8-25 client threads), but expanding those options to be yokozuna default seems too pedantic. 

I analyzed opened sockets in my tests, and when the JVM takes a long time, the connection remains active until the process is complete. If the data coming in is too large and fast, with insufficient JVM resources to manage the data being indexed in a timely manner, you'll see ports in use. The only solution I could see to this problem is ensuring that the JVM runs fast enough that the connections aren't used for more than a few seconds, which depends entirely on Solr being fast.
## Other change

There's also a minor change in here. While making these _yz_solr changes, Dmitri found an error where his key was named for the state of the Oregon, causing the delete query to fail on a reserved word (_yz_rk:OR), so I wrapped it in quotes.
## Documentation

Issue #330 gives us a glimpse of some problems that an ill-fitting JVM can cause. We're going to need to find and share plenty of documentation about how/where a user can got o investigate their resource requirements, and troubleshoot problems: basho/basho_docs#1012.

### DIFF
--- a/priv/default_schema.xml
+++ b/priv/default_schema.xml
@@ -104,7 +104,7 @@
    <dynamicField name="*_set"      type="string"  indexed="true" stored="false" multiValued="true" />
 
    <!-- catch-all field -->
-   <dynamicField name="*" type="text_general" indexed="false" stored="false" multiValued="true" />
+   <dynamicField name="*" type="ignored" />
 
    <field name="_yz_id" type="_yz_str" indexed="true" stored="true" multiValued="false" required="true"/>
 
--- a/priv/default_schema.xml
+++ b/priv/default_schema.xml
@@ -104,7 +104,7 @@
    <dynamicField name="*_set"      type="string"  indexed="true" stored="false" multiValued="true" />
 
    <!-- catch-all field -->
-   <dynamicField name="*" type="text_general" indexed="true" stored="false" multiValued="true" />
+   <dynamicField name="*" type="text_general" indexed="false" stored="false" multiValued="true" />
 
    <field name="_yz_id" type="_yz_str" indexed="true" stored="true" multiValued="false" required="true"/>
 
--- a/src/yz_solr.erl
+++ b/src/yz_solr.erl
@@ -31,6 +31,7 @@
 -define(DEFAULT_URL, "http://localhost:8983/solr").
 -define(DEFAULT_VCLOCK_N, 1000).
 -define(QUERY(Str), {struct, [{'query', Str}]}).
+-define(SOLR_TIMEOUT, 60000).
 
 %% @doc This module provides the interface for making calls to Solr.
 %%      All interaction with Solr should go through this API.
@@ -63,7 +64,7 @@ commit(Core) ->
     URL = ?FMT("~s/~s/update?~s", [base_url(), Core, Encoded]),
     Headers = [{content_type, "application/json"}],
     Opts = [{response_format, binary}],
-    case ibrowse:send_req(URL, Headers, post, JSON, Opts) of
+    case ibrowse:send_req(URL, Headers, post, JSON, Opts, ?SOLR_TIMEOUT) of
         {ok, "200", _, _} -> ok;
         Err -> throw({"Failed to commit", Err})
     end.
@@ -72,7 +73,7 @@ commit(Core) ->
 -spec core(atom(), proplist()) -> {ok, list(), binary()} |
                                   {error, term()}.
 core(Action, Props) ->
-    core(Action, Props, 5000).
+    core(Action, Props, ?SOLR_TIMEOUT).
 
 -spec core(atom(), proplist(), ms()) -> {ok, list(), binary()} |
                                         {error, term()}.
@@ -110,7 +111,7 @@ delete(Index, Ops) ->
     URL = ?FMT("~s/~s/update", [base_url(), Index]),
     Headers = [{content_type, "application/json"}],
     Opts = [{response_format, binary}],
-    case ibrowse:send_req(URL, Headers, post, JSON, Opts) of
+    case ibrowse:send_req(URL, Headers, post, JSON, Opts, ?SOLR_TIMEOUT) of
         {ok, "200", _, _} -> ok;
         Err -> {error, Err}
     end.
@@ -169,7 +170,7 @@ index(Core, Docs, DelOps) ->
     URL = ?FMT("~s/~s/update", [base_url(), Core]),
     Headers = [{content_type, "application/json"}],
     Opts = [{response_format, binary}],
-    case ibrowse:send_req(URL, Headers, post, JSON, Opts) of
+    case ibrowse:send_req(URL, Headers, post, JSON, Opts, ?SOLR_TIMEOUT) of
         {ok, "200", _, _} -> ok;
         Err -> throw({"Failed to index docs", Ops, Err})
     end.
@@ -197,7 +198,7 @@ partition_list(Core) ->
     Encoded = mochiweb_util:urlencode(Params),
     URL = ?FMT("~s/~s/select?~s", [base_url(), Core, Encoded]),
     Opts = [{response_format, binary}],
-    case ibrowse:send_req(URL, [], get, [], Opts) of
+    case ibrowse:send_req(URL, [], get, [], Opts, ?SOLR_TIMEOUT) of
         {ok, "200", _, Resp} -> {ok, Resp};
         Err -> {error, Err}
     end.
@@ -240,7 +241,7 @@ search(Core, Headers, Params) ->
     URL = ?FMT("~s/~s/select", [base_url(), Core]),
     Headers2 = [{content_type, "application/x-www-form-urlencoded"}|Headers],
     Opts = [{response_format, binary}],
-    case ibrowse:send_req(URL, Headers2, post, Body, Opts) of
+    case ibrowse:send_req(URL, Headers2, post, Body, Opts, ?SOLR_TIMEOUT) of
         {ok, "200", RHeaders, Resp} -> {RHeaders, Resp};
         {ok, CodeStr, _, Err} ->
             {Code, _} = string:to_integer(CodeStr),
@@ -320,10 +321,10 @@ encode_commit() ->
 %% @doc Encode a delete operation into a mochijson2 compatiable term.
 -spec encode_delete(delete_op()) -> term().
 encode_delete({key,Key}) ->
-    Query = ?YZ_RK_FIELD_S ++ ":" ++ ibrowse_lib:url_encode(binary_to_list(Key)),
+    Query = ?YZ_RK_FIELD_S ++ ":\"" ++ ibrowse_lib:url_encode(binary_to_list(Key)) ++ "\"",
     ?QUERY(list_to_binary(Query));
 encode_delete({siblings,Key}) ->
-    Query = ?YZ_RK_FIELD_S ++ ":" ++ ibrowse_lib:url_encode(binary_to_list(Key)) ++ " AND " ++ ?YZ_VTAG_FIELD_S ++ ":[* TO *]",
+    Query = ?YZ_RK_FIELD_S ++ ":\"" ++ ibrowse_lib:url_encode(binary_to_list(Key)) ++ "\" AND " ++ ?YZ_VTAG_FIELD_S ++ ":[* TO *]",
     ?QUERY(list_to_binary(Query));
 encode_delete({'query', Query}) ->
     ?QUERY(Query);
--- a/priv/conf/solrconfig.xml
+++ b/priv/conf/solrconfig.xml
@@ -387,7 +387,7 @@
          Recommend values of 1-2 for read-only slaves, higher for
          masters w/o cache warming.
       -->
-    <maxWarmingSearchers>4</maxWarmingSearchers>
+    <maxWarmingSearchers>2</maxWarmingSearchers>
 
   </query>
 
--- a/priv/conf/solrconfig.xml
+++ b/priv/conf/solrconfig.xml
@@ -51,7 +51,7 @@
          If both ramBufferSizeMB and maxBufferedDocs is set, then
          Lucene will flush based on whichever limit is hit first.  -->
     <!-- Reduce disk flushes. If YZ loses data on crash, AAE will repair anyway -->
-    <ramBufferSizeMB>32</ramBufferSizeMB>
+    <!-- <ramBufferSizeMB>100</ramBufferSizeMB> -->
 
     <!-- Expert: Merge Policy
          The Merge Policy in Lucene controls how merging of segments is done.
@@ -324,7 +324,7 @@
 
     <fieldValueCache class="solr.FastLRUCache"
                      size="512"
-                     autowarmCount="128"
+                     autowarmCount="0"
                      showItems="32" />
   
 
@@ -387,7 +387,7 @@
          Recommend values of 1-2 for read-only slaves, higher for
          masters w/o cache warming.
       -->
-    <maxWarmingSearchers>2</maxWarmingSearchers>
+    <maxWarmingSearchers>4</maxWarmingSearchers>
 
   </query>
 
--- a/priv/solr/etc/jetty.xml
+++ b/priv/solr/etc/jetty.xml
@@ -60,8 +60,11 @@
             <Call class="java.lang.System" name="setProperty"> <Arg>log4j.configuration</Arg> <Arg>etc/log4j.properties</Arg> </Call>
             <Set name="host"><SystemProperty name="jetty.host" /></Set>
             <Set name="port"><SystemProperty name="jetty.port" default="8983"/></Set>
-            <Set name="maxIdleTime">50000</Set>
-            <Set name="lowResourceMaxIdleTime">1500</Set>
+            <Set name="maxIdleTime">60000</Set>
+            <!-- this could prematurely close sockets, but is better than tying up resources -->
+            <Set name="lowResourceMaxIdleTime">5000</Set>
+            <!-- acceptors should equal number of CPUs * 2 -->
+            <!-- <Set name="acceptors">16</Set> -->
             <Set name="statsOn">false</Set>
           </New>
       </Arg>
--- a/riak_test/yz_languages.erl
+++ b/riak_test/yz_languages.erl
@@ -26,6 +26,7 @@ confirm() ->
     confirm_body_search_encoding(Cluster),
     confirm_language_field_type(Cluster),
     confirm_tag_encoding(Cluster),
+    confirm_reserved_word_safety(Cluster),
     pass.
 
 select_random(List) ->
@@ -61,9 +62,12 @@ store_and_search(Cluster, Bucket, Index, CT, Body, Field, Term) ->
     store_and_search(Cluster, Bucket, Index, Headers, CT, Body, Field, Term).
 
 store_and_search(Cluster, Bucket, Index, Headers, CT, Body, Field, Term) ->
+    store_and_search(Cluster, Bucket, Index, "test", Headers, CT, Body, Field, Term).
+
+store_and_search(Cluster, Bucket, Index, Key, Headers, CT, Body, Field, Term) ->
     HP = select_random(host_entries(rt:connection_info(Cluster))),
     create_index(Cluster, HP, Index),
-    URL = bucket_url(HP, Bucket, "test"),
+    URL = bucket_url(HP, Bucket, Key),
     lager:info("Storing to bucket ~s", [URL]),
     {ok, "204", _, _} = ibrowse:send_req(URL, Headers, put, Body),
     %% Sleep for soft commit
@@ -97,3 +101,12 @@ confirm_tag_encoding(Cluster) ->
                {"x-riak-meta-yz-tags", "x-riak-meta-arabic_s"},
                {"x-riak-meta-arabic_s", <<"أقرأ"/utf8>>}],
     store_and_search(Cluster, Bucket, Index, Headers, "text/plain", Body, "arabic_s", "أقرأ").
+
+confirm_reserved_word_safety(Cluster) ->
+    Index = <<"reserved">>,
+    Bucket = {Index, <<"b">>},
+    lager:info("confirm_reserved_word_safety ~s", [Index]),
+    Body = <<"whatever">>,
+    Headers = [{"Content-Type", "text/plain"}],
+    RKey = "ON",
+    store_and_search(Cluster, Bucket, Index, RKey, Headers, "text/plain", Body, "text", "whatever").
--- a/riak_test/yz_languages.erl
+++ b/riak_test/yz_languages.erl
@@ -108,5 +108,5 @@ confirm_reserved_word_safety(Cluster) ->
     lager:info("confirm_reserved_word_safety ~s", [Index]),
     Body = <<"whatever">>,
     Headers = [{"Content-Type", "text/plain"}],
-    RKey = "ON",
+    RKey = "OR",
     store_and_search(Cluster, Bucket, Index, RKey, Headers, "text/plain", Body, "text", "whatever").


src/yz_solr.erl
```diff
@@ -320,10 +321,10 @@ encode_commit() ->
 %% @doc Encode a delete operation into a mochijson2 compatiable term.
 -spec encode_delete(delete_op()) -> term().
 encode_delete({key,Key}) ->
-    Query = ?YZ_RK_FIELD_S ++ ":" ++ ibrowse_lib:url_encode(binary_to_list(Key)),
+    Query = ?YZ_RK_FIELD_S ++ ":\"" ++ ibrowse_lib:url_encode(binary_to_list(Key)) ++ "\"",
```
rzezeski, line 336: I wonder if we are breatking UTF-8 here? I don't think we have tests for indexing/searching/deleting UTF-8 keys. I'll write up a separate issue.

rzezeski:
@wbrown If you have the chance I would love to hear if this patch improves the issues you noticed in #330 & #320.

wbrown:
@rzezeski Awesome, I've read these notes.  So, this is informative and tells me that if I want good Solr performance, I need to architect for it.  i.e. highly clocked machines.  I'll give the patch set a shot.  Do I apply this against the 2.0.0-pre20 download?

rzezeski:
@wbrown I'm not sure if it will apply cleanly to pre20 download, but you can certainly try.

priv/default_schema.xml
```diff
@@ -104,7 +104,7 @@
    <dynamicField name="*_set"      type="string"  indexed="true" stored="false" multiValued="true" />
 
    <!-- catch-all field -->
-   <dynamicField name="*" type="text_general" indexed="true" stored="false" multiValued="true" />
+   <dynamicField name="*" type="text_general" indexed="false" stored="false" multiValued="true" />
```
rzezeski: If we are not going to index the catch-all field then set the type to `ignored` and drop the other attributes so they are inherited from the type. IMO this makes the drop semantic obvious. It also may prevent unnecessary analyzing since it uses the `StrField` type (I would hope `indexed=false` prevents analysis but use a non-analyzed type just in case).

rzezeski:
> Since all cache auto warmers were set to zero, I also bumped up with default maxWarmingSearchers from 2 (suitable for read-only slaves, as mentioned in the config comments) to 4. I found that on high loads of large objects, I'd get an exceeded limit of maxWarmingSearchers=X message, which this resolved.

My reading of this is we should NOT raise the limit to 4. If Solr is
overloaded and can't keep up with the load then allowing more
concurrent warming searchers is only going to drag it down more. I
would rather shed load than avoid error messages in the log.

> before the warming searcher from first commit has enough time to warm up, then there can be multiple searchers all competeing for resources at the same time, _EVEN HTOUGH ONE OF THEM WILL BE THROWN AWAY AS SOON AS THE NEXT ONE IS READY_. (emphasis mine)

This sentence makes me want to set the max to 1. If Solr can't keep up
with the commit load then throwing more concurrent load on the pile
will not help. Given that we auto soft-commit at 1s interval what
sense does it make to pile up warmers that are just going to
invalidate each other? That is, if a commit takes 5s to warm then I
don't want 4 concurrent warmers running, 3 of which are going to
quickly invalidated once the fourth finishes. Yes, updates will take
longer to be visible but in this case I think it's worth trading
harvest for yield because piling up concurrent warmers could end up
taking down the node; or worse, slowing it down just enough to make
latency balloon.

> If you encounter this error a lot, you can (in theory) increase the number in your maxWarmingSearchers, but that is risky to do unless you are confident you have the system resources (RAM, CPU, etc...) to do it safely. _A MORE CORRECT WAY TO DEAL WITH THE SITUATION IS TO REDUCE HOW FREQUENTLY YOU SEND COMMITS_. (emphases mine)

I think this is the crux of the issue. In the future I would like
Yokozuna to expose a soft/hard commit call with the ability to turn
off the auto commits. It won't make sense for everyone but it could
allow more efficient use of resources for those that have more regular
data load patterns.

@coderoshi Did the `exceeded limit of maxWarmingSearchers=X` error
cause any actual issues like crashing the node or cause failed writes?
Even if it failed writes I think I'm still inclined to prefer that
behavior over adding more load. I'm tempted to play around with this a
bit myself. Is there code I could run to reproduce your benchmark?

coderoshi:
@rzezeski no, it didn't cause any failures, it was just filling the logs.

Here's a script I was running. https://gist.github.com/coderoshi/329b93ee7987b1beecdf

I later ran some tests directly against solr in order to riak/yz out of the equation, for straight solr tuning

``` ruby
bh = big_hash.clone
bh['_yz_id'] = SecureRandom.hex.to_s
doc = {"commit" => {}, "add" => {"doc" => bh}}.to_json
File.open("tmp/#{bh['_yz_id']}.json", 'w+') {|f| f.write(doc)}
`curl 'http://192.168.1.135:10034/solr/herp/update' -H'Content-type:application/json' --data-binary @tmp/#{bh['_yz_id']}.json 2> /dev/null`
```

hectcastro:
Putting this together as a way to verify my understanding of the `maxWarmingSearchers` changes proposed:
- We're setting all of the cache `autowarmCount` settings to `0` in order to minimize the duration of auto warming so that `maxWarmingSearchers` is not exceeded
- Here we are trading the value of caching for speed of making new values available in query results
- We have no `solr.QuerySenderListener` configured by default, so combining that with setting all of the `autowarmCount` settings to `0`, we've done pretty much all we can in minimizing the warming done before searchers are moved out of the "warming" state
- ~~`maxWarmingSearchers` was bumped from `2` to `4` because we removed as much warming overhead as possible~~
- ~~We may want `maxWarmingSearchers` to be `> 1` because it is something that happens per index (having lots of indexes and setting this to `1` would hurt overall ability to provide fresh query results across all indexes)~~

rzezeski:
> We may want maxWarmingSearchers to be > 1 because it is something that happens per index (having lots of indexes and setting this to 1 would hurt overall ability to provide fresh query results across all indexes)

Are you sure this applies across all cores? Each core gets its own `solrconfig.xml` so that is surprising behavior to me.

rzezeski:
> maxWarmingSearchers was bumped from 2 to 4 because we removed as much warming overhead as possible

My understand is it was bumped to avoid logging noise. AFAICT running
out of warmers is not a hard error but rather a safety valve to
prevent overloading the core. Reading the following wiki entry makes
me think lower would be better given our 1s auto soft-commit and
preference for availability. This becomes even more important if
warming searchers is per-core. If we have 16 cores that could be 48
threads crunching.

http://wiki.apache.org/solr/FAQ#What_does_.22exceeded_limit_of_maxWarmingSearchers.3DX.22_mean.3F

coderoshi:
So my reasoning for zeroing out `autowarmCount` was in response to the solr documentation suggestions for slow commit times, and the fact that autowarmCount="0" is actually the default for current solr.

As for bumping up `maxWarmingSearchers`, that was, as @rzezeski suggested, solely to reduce the complaints in the log.

I do believe we should keep autowarmCount at zero, however, I'm not so strident about keeping maxWarmingSearchers at 4. Increasing the size didn't seem to make much of a difference for my tests, so I figured what the hell?

If log noise doesn't bother anyone, I'm fine with dropping `maxWarmingSearchers` back to 2 or even 1.

rzezeski:
> I do believe we should keep autowarmCount at zero

I agree. Given our 1s auto commit I think auto warming may be hurting
more than helping. This is something I've meant to benchmark for a
while now.

The real solution, in the future, is to provide APIs for manipulating
certain aspects of the `solrconfig.xml` because things like auto
commit, caches and auto warming are all highly dependent on query
profile. One size does not fit all here. I foresee plenty of complaints
in the future about the hardcoded `solrconfig.xml` but the 2.0 feature
ship has sailed so until then we can try to protect ourselves (if
users REALLY want a custom setup they should be able to override
manually at filesystem level).

wbrown:
@rzezeski I grabbed the patch from this commit, and applied it to the riak 2.0beta1 sources that went up for download.  I even have a nice Arch Linux packaging script that will build me a Riak 2.0 beta1 package with that patch installed.

Unfortunately, I ended up beating my head against the desk for hours.  What had worked before doesn't appear to work anymore, and it possibly has to do with the bucket types stuff.  It's worked in the past by setting the bucket's index_search value to the index and schema in question.

So I got the latest bleeding edge Python client that had bucket types support added, worked my way through some of the bugs and lack of documentation there, and transitioned my indexing engine classes to use bucket types.  I have indexes being created in Solr, buckets being assigned to bucket types, and bucket objects being grabbed from the bucket type object to ensure that we use the Yokozuna indexing.  But I am absolutely not seeing any documents going into Yokozuna/Solr.

My setup step are:
- Initialize new Riak cluster.
- Create bucket types from `riak-admin` with the following script:

``` bash
bts=(obs entity raw)
for bt in "${bts[@]}";
do
    bin/riak-admin bucket-type create $bt '{"props":{}}'
    bin/riak-admin bucket-type activate $bt
done
```
- Upon initialization of my application, it does the following:

``` python
        ### create our schema if it doesn't exist
        try:
            self.riak.get_search_schema(schemaName)
        except riak.RiakError:
            self.riak.create_search_schema(schemaName,
                open(schemaFile).read())

       ### our bucket type is the same name as our bucket name
       indexedBucketType = self.riak.bucket_type( self.bucketName )
       self.bucket = indexedBucketType.bucket( self.bucketName )

        ### create our index in Yokozuna if it doesn't exist
        if not self.bucketName in [ x['name']
                            for x in self.riak.list_search_indexes() ]:
            self.riak.create_search_index( RKeyString( self.bucketName ),
                                           RKeyString( schemaName ),
                                           self.replicas )
            print "* Index created for", self.bucketName
            time.sleep(5)

        ### Set the bucket type to our desired n_val.
        if indexedBucketType.get_property( "n_val", None ) != self.replicas:
            indexedBucketType.set_property( "n_val", self.replicas )

        ### set our bucket type's index to our desired index
        try:
            searchIndex = indexedBucketType.get_property("search_index", None)
        except KeyError:
            indexedBucketType.set_property("search_index", self.bucketName)
            searchIndex = indexedBucketType.get_property("search_index", None)
        if searchIndex != self.bucketName:
            indexedBucketType.set_property("search_index", self.bucketName)
```

Am I missing any obvious steps here?

When I store to an indexed bucket, I ensure that I acquire the bucket object via the `bucket_type` instantiation method:

``` python
        if indexEngine:
            self.getBucket = self.riak.bucket_type( bucketName ).bucket
        else:
            self.getBucket = self.riak.bucket

        self.bucket = self.getBucket(self.name)
```

When I double check the object representation with print, it is giving the expected representation -- a bucket type of `raw` and a bucket of `raw`.

rzezeski:
There seem to dialyzer issues on develop so going to figure those out first before getting the final review done on this issue.

wbrown:
@rzezeski @coderoshi So, reporting in now that I've successfully gotten this to work.  It was some subtle issues brought up by the bleeding edge version of the Python client that I had to use to get the `bucket_type` stuff working that had an impact on my code.  It's not ready for prime time, unfortunately, with some breaking API changes.  I will need to sit down and figure out what issues and patches I can make against the client when I have more time than I do at the moment.

It is feeling like for high load situations like mine, JVM tuning is absolutely essential.  I took the JVM settings that were outlined in the initial comment opening this PR, and applied it.  To make it work, I had to enable Huge Pages, and allocate about 2048 huge pages at 2MB each.  This is with a slightly reduced `-Xms4g -Xmx8g` setting as my development machines have 32GB each.  I also had to raise the kernel shared memory maximum.

Things are looking better now:

```
node_puts : 22254
node_puts_total : 1189134
node_put_fsm_time_mean : 4309
node_put_fsm_time_median : 1325
node_put_fsm_time_95 : 3574
node_put_fsm_time_99 : 119076
node_put_fsm_time_100 : 259319
```

The above numbers put me in the ballpark of `300` to `400` new indexed keys a second, with an apparent median latency of `1325us`

I'm seeing client side put latency of `4.84ms` to `10.96ms` with a load of about 10-30 active client-side threads, with Yokozuna and a Bitcask backend.  This is interleaved with a  client get latency of about `1.01ms` to `2.28ms`.  Not bad at all considering that the client is connecting to my Riak nodes other Gigabit.

Looking at the network side of the picture:

```
[riak@cumulus ~]$ netstat -anp | grep :8093 | grep TIME_WAIT | wc -l
49

[riak@cumulus ~]$ netstat -anp | grep :8093 | grep ESTABLISHED | wc -l
20
```

So, about `20` in `TIME_WAIT` and `10` in `ESTABLISHED`, and the number isn't growing.

We'll see how this does once I've fully reloaded my dataset thus far into the new database.  It's been my observation that Yokozuna usually starts out great, and then slows down.  A part of this improvement is that I've improved my selection and versioning algorithms since I last used Yokozuna, and had been having good performance with `2i` -- but this is quite promising.  Yokozuna allows a lot more flexibility in query terms than `2i`, as well as multiple query terms.

Thank you -- and let me know if there's anything I can test or check out here.  I am quite motivated to see a performant Yokozuna in 2.0 release.  I'll keep you guys in the loop as I rebuild my dataset.

jonmeredith:
Thanks for the detailed report @wbrown.

On Mon, Apr 21, 2014 at 7:10 PM, wbrown notifications@github.com wrote:

> @rzezeski https://github.com/rzezeski @coderoshihttps://github.com/coderoshiSo, reporting in now that I've successfully gotten this to work. It was
> some subtle issues brought up by the bleeding edge version of the Python
> client that I had to use to get the bucket_type stuff working that had an
> impact on my code. It's not ready for prime time, unfortunately, with some
> breaking API changes. I will need to sit down and figure out what issues
> and patches I can make against the client when I have more time than I do
> at the moment.
> 
> It is feeling like for high load situations like mine, JVM tuning is
> absolutely essential. I took the JVM settings that were outlined in the
> initial comment opening this PR, and applied it. To make it work, I had to
> enable Huge Pages, and allocate about 2048 huge pages at 2MB each. This is
> with a slightly reduced -Xms4g -Xmx8g setting as my development machines
> have 32GB each. I also had to raise the kernel shared memory maximum.
> 
> Things are looking better now:
> 
> node_puts : 22254
> node_puts_total : 1189134
> node_put_fsm_time_mean : 4309
> node_put_fsm_time_median : 1325
> node_put_fsm_time_95 : 3574
> node_put_fsm_time_99 : 119076
> node_put_fsm_time_100 : 259319
> 
> The above numbers put me in the ballpark of 300 to 400 new indexed keys a
> second, with an apparent median latency of 1325us
> 
> I'm seeing client side put latency of 4.84ms to 10.96ms with a load of
> about 10-30 active client-side threads, with Yokozuna and a Bitcask
> backend. This is interleaved with a client get latency of about 1.01ms to
> 2.28ms. Not bad at all considering that the client is connecting to my
> Riak nodes other Gigabit.
> 
> Looking at the network side of the picture:
> 
> [riak@cumulus ~]$ netstat -anp | grep :8093 | grep TIME_WAIT | wc -l
> 49
> 
> [riak@cumulus ~]$ netstat -anp | grep :8093 | grep ESTABLISHED | wc -l
> 20
> 
> So, about 20 in TIME_WAIT and 10 in ESTABLISHED, and the number isn't
> growing.
> 
> We'll see how this does once I've fully reloaded my dataset thus far into
> the new database. It's been my observation that Yokozuna usually starts out
> great, and then slows down. A part of this improvement is that I've
> improved my selection and versioning algorithms since I last used Yokozuna,
> and had been having good performance with 2i -- but this is quite
> promising. Yokozuna allows a lot more flexibility in query terms than 2i,
> as well as multiple query terms.
> 
> Thank you -- and let me know if there's anything I can test or check out
> here. I am quite motivated to see a performant Yokozuna in 2.0 release.
> I'll keep you guys in the loop as I rebuild my dataset.
> 
> ## 
> 
> Reply to this email directly or view it on GitHubhttps://github.com/basho/yokozuna/pull/358#issuecomment-40994183
> .

## 

Jon Meredith
VP, Engineering
Basho Technologies, Inc.
jmeredith@basho.com

wbrown:
Update after the load test/data reloading had been running for about 14 hours.  Shortly before I went to bed, I went ahead and started a second client feeding another set of data in.

I'm at about 30 million keys indexed now.  With two observation engine clients and a GeoIP database client populating Riak over Gigabit on a single workstation, my client-side latency stats are:
- `GET` ranges from `5.07ms` to `20.26ms` with an apparent median of ~`11ms`
- `PUT` ranges from `7.20ms` to `42.53ms` with an apparent median of ~`30ms`

This is with a current rate of about `562` keys a second `PUT` into Riak.

```
node_put_fsm_active : 5
node_put_fsm_active_60s : 34074
node_put_fsm_in_rate : 429
node_put_fsm_out_rate : 427
```

Latency measured by Riak is pretty good, though 95th and 99th percentile times aren't great at `327ms` to `2679ms`:

```
node_put_fsm_time_mean : 13050
node_put_fsm_time_median : 1413
node_put_fsm_time_95 : 11724
node_put_fsm_time_99 : 327825
node_put_fsm_time_100 : 2679843
```

`GET` latencies are pretty good from the Riak side still, though 99th and 100th is horrible at `173ms` and `2799ms`:

```
node_get_fsm_time_mean : 5374
node_get_fsm_time_median : 254
node_get_fsm_time_95 : 5424
node_get_fsm_time_99 : 173151
node_get_fsm_time_100 : 2799747
```

With 30 million keys, that's enough to outsize the `32GB` of RAM in my two test nodes, and `32GB` of RAM in my Infiniband ZFS filer.  Each node is up to about `129GB` of data stored, with a `compressratio` of `2.35x`, for an actual dataset size of `303GB`, so we're getting a pretty good picture of 'real world' dataset load.

Riak with Bitcask is at `9.1GB` of RAM and growing, while Solr is at `2.57GB` of RAM.

So, things are overall improved versus my first attempt basically causing Riak, Yokozuna, and Solr to crash and burn.  However, some confusing flies in the ointment:
- I am still seeing a fair amount of socket turnover.  The numbers are even, but there's still connections dropped and connected despite being under load.

```
[root@cumulus wbrown]# netstat -anp | grep :8093  | grep EST | grep java | wc -l
10
[root@cumulus wbrown]# netstat -anp | grep :8093  | grep TIME_WAIT  | wc -l
45
```

Looking closer:

```
[root@cumulus wbrown]# netstat -anp | grep :8093  | grep EST | grep java; echo "Waiting 10s"; sleep 10; netstat -anp | grep :8093  | grep EST | grep java;
tcp6       0      0 127.0.0.1:8093          127.0.0.1:56964         ESTABLISHED 24086/java          
tcp6       0      0 127.0.0.1:8093          127.0.0.1:43465         ESTABLISHED 24086/java          
tcp6       0      0 127.0.0.1:8093          127.0.0.1:40233         ESTABLISHED 24086/java          
tcp6       0      0 127.0.0.1:8093          127.0.0.1:40236         ESTABLISHED 24086/java          
tcp6     946      0 127.0.0.1:8093          127.0.0.1:45451         ESTABLISHED 24086/java          
tcp6       0      0 127.0.0.1:8093          127.0.0.1:47951         ESTABLISHED 24086/java          
tcp6    5962      0 127.0.0.1:8093          127.0.0.1:38766         ESTABLISHED 24086/java          
Waiting 10s
tcp6       0      0 127.0.0.1:8093          127.0.0.1:40469         ESTABLISHED 24086/java          
tcp6       0      0 127.0.0.1:8093          127.0.0.1:50734         ESTABLISHED 24086/java          
tcp6       0      0 127.0.0.1:8093          127.0.0.1:42569         ESTABLISHED 24086/java          
tcp6       0      0 127.0.0.1:8093          127.0.0.1:45451         ESTABLISHED 24086/java          
tcp6       0      0 127.0.0.1:8093          127.0.0.1:60645         ESTABLISHED 24086/java          
tcp6       0      0 127.0.0.1:8093          127.0.0.1:48799         ESTABLISHED 24086/java          
tcp6       0      0 127.0.0.1:8093          127.0.0.1:39427         ESTABLISHED 24086/java          
tcp6       0      0 127.0.0.1:8093          127.0.0.1:47883         ESTABLISHED 24086/java          
tcp6       0    147 127.0.0.1:8093          127.0.0.1:38766         ESTABLISHED 24086/java          
tcp6       0      0 127.0.0.1:8093          127.0.0.1:51918         ESTABLISHED 24086/java 
```

We can see that the connections have completely turned over in a `10s` interval with a single connection maintained..  It's almost as if the timeout settings are not working.  Another interesting note to make -- the connections via `lo` are via `IPv6` rather than `IPv4`, so I wonder if that's making an impact.  There shouldn't be a near complete connection turnover in `10s` if the timeouts are set to `60s`.

A side question - I'm really big on metrics.  You cannot do performance analysis without metrics.  Do we have any Yokozuna latency metrics exposed anywhere, or do I need to set up Riak on OmniOS for `dtrace`? :)  It'd be nice to have in the `riak-admin status` pane.  I can open an issue for this, if indicated helpful.

rzezeski:
@wbrown First, thank you so much for all the great information from
the field. The detail of your reporting is much appreciated. That
said, here is a reply to one of your comments:

> I am still seeing a fair amount of socket turnover. The numbers are even, but there's still connections dropped and connected despite being under load.
> ...
> We can see that the connections have completely turned over in a 10s interval with a single connection maintained.. It's almost as if the timeout settings are not working. Another interesting note to make -- the connections via lo are via IPv6 rather than IPv4, so I wonder if that's making an impact. There shouldn't be a near complete connection turnover in 10s if the timeouts are set to 60s.

Yes this is what I was trying to explain in my
[comment about the ibrowse load balancer](https://github.com/basho/yokozuna/issues/330#issuecomment-40783996) the other day. It
defaults to a 10s `inactivity_timeout`. That, along with its poor
algorithm causes connections to constantly his this 10s timeout. This
timeout was **NOT** adjusted in this patch. I tried setting it in the
global config but ibrowse didn't seem to pick it up. I want to try
again but setting the timeout in the code but I may open a more
focused PR as this one is already making too many changes at once.

rzezeski:
My findings are that higher time out doesn't prevent the
`EofException`; it simply requires higher latency of the request to
hit it. I verified this by setting both the ibrowse request timeout
and Jetty idle timeout to 5 seconds and applying load. It managed to
produce plenty of `EofException`. The only method to completely
prevent `EofException` was to make the ibrowse request time out
substantially higher than the Jetty idle time out. This allows Jetty
to be in control of closing connections. This can be explored in a
different patch if need be.

I also found that these changes don't solve connection churn as that
is related to a combination of the `inactivity_timeout` and bad load
balancing algorithm [[1]](https://github.com/basho/yokozuna/issues/330#issuecomment-40783996). This, in theory, could be fixed with a
raise to the `inactivity_timeout` but I will leave that to a focused
PR rather than appending to this one.

Overall, the 60 second time out did make things quieter since it makes
it harder to time out. I also noticed a bump in throughput of about
27% from 1.1 writes/s to 1.4 writes/s. This patch actually causes the
median latency to raise slightly because it is allowing requests to
take longer than 30 seconds. The 99th is lower, probably due to less
crashes and thus less busy work on the server.

| --- | --- | --- | --- |
| develop | 1.1 | 17.1 s | 44.5 s |
| patch 358 | 1.4 | 19.5 s | 40.3 s |

In general, I think time outs need to be looked at more closely,
especially in regards to user semantics. If a user sets a timeout of 5
seconds on the write request should the index operation allow to block
the vnode for 60 seconds?

With that I'm going to consider the time outs reviewed so that I can
get to the other changes made in this PR.

rzezeski:
The riak tests look good:

```
Test Results:
0 Tests Failed
18 Tests Passed
```

rzezeski:
I don't recall `ON` being a reserved word in Lucene query syntax so I changed to `OR`. Test still passes.

```
Test Results:
yz_languages-bitcask: pass
0 Tests Failed
1 Tests Passed
```

src/yz_solr.erl
rzezeski, line 336: This solves the reserved word problem by using a phrase query but the more appropriate solution down the line is to use the raw or term query parsers.

http://wiki.apache.org/solr/QueryParser

rzezeski:
+1 61d632

List of the things I reviewed:
- [x] compile/unit
- [x] dialyzer
- [x] riak test
- [x] review timeouts
- [x] review OR change
- [x] review `solrconfig.xml` changes
- [x] review schema changes
- [x] review jetty.xml changes

rzezeski:
+1 61d6326

rzezeski:
@borshop merge

PR #361, opened 2014-04-14
Add support for pulling down dependencies via curl

hectcastro:
OS X doesn't have `wget` installed by default. Despite the fact that most of us have figured out a way to install it, I'd like to add support for `curl` so that we can make one less inconvenience for those who don't have `wget` installed.

### DIFF
--- a/tools/build-solr.sh
+++ b/tools/build-solr.sh
@@ -97,8 +97,10 @@ else
     SOLR_FILE=$(basename $URL)
     SOLR_DIR=${SOLR_FILE%-src.tgz}
 
-    if test ! -e $SOLR_FILE; then
+    if which wget > /dev/null && test ! -e $SOLR_FILE; then
         wget $URL
+    elif which curl > /dev/null && test ! -e $SOLR_FILE; then
+        curl -O $URL
     fi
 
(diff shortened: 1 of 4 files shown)


tools/build-solr.sh
```diff
@@ -97,8 +97,10 @@ else
     SOLR_FILE=$(basename $URL)
     SOLR_DIR=${SOLR_FILE%-src.tgz}
 
-    if test ! -e $SOLR_FILE; then
+    if which wget > /dev/null && test ! -e $SOLR_FILE; then
```
macintux: Should `download()` be copied to this script as well, or at least use the same `curl` and `wget` progress arguments?

hectcastro: I'm not opposed to making each download attempt uniform. I'll defer the decision to @rzezeski.

coderoshi: @hectcastro if you're in here, may as well extract this to a download func

hectcastro:
@coderoshi @macintux I pulled `download()` into `build-solr.sh`.

coderoshi:
+1 201ac6c

hectcastro:
@borshop merge

PR #362, opened 2014-04-15
Fixes calls into security

lordnull:
Without these, the yz_security riak_test is unable to pass.

### DIFF
--- a/riak_test/yz_security.erl
+++ b/riak_test/yz_security.erl
@@ -137,7 +137,7 @@ confirm_index_pb(Node) ->
     riakc_pb_socket:stop(Pid0),
 
     lager:info("Grant index permission to user"),
-    ok = ?GRANT(Node, [[?YZ_SECURITY_ADMIN_PERM,"ON","index","TO",?USER]]),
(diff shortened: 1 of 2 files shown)


jonmeredith:
Are the end user commands case sensitive?

rzezeski:
> Are the end user commands case sensitive?

Good question. I was under the impression we recently patched security so that is is case _insensitive_. Why does the test fail if uppercase is used?

lordnull:
Seems it is case sensitive; using the command-line script, grant fails if the "on" or "to" have capitals.

rzezeski:
> Seems it is case sensitive; using the command-line script, grant fails if the "on" or "to" have capitals.

Perhaps it was changed from upper to lower and I misread it as case insensitive. IIRC @macintux did the work?

macintux:
Sorry I didn't dig for code that uses this command, mea culpa.

Correct, the code is case-sensitive (like all of our riak-admin commands that I can tell) and I changed it from all-upper to all-lower.

rzezeski:
+1 3eba618

```
Test Results:
yz_security-bitcask: pass
0 Tests Failed
1 Tests Passed
```

rzezeski:
@borshop merge

PR #363, opened 2014-04-21
fix security code

rzezeski:
I'm opening this PR on behalf of @macintux because there is a traffic jam of Yokozuna PRs that are all blocked on fixes for security code. At the same time I'm taking over review on this issue. After this issue #362 is next and then develop should once again work with the latest security code.

### DIFF
--- a/src/yz_wm_search.erl
+++ b/src/yz_wm_search.erl
@@ -77,25 +77,13 @@ is_authorized(ReqData, Ctx) ->
     end.
 
 
-%%
-%% Written with the idea of using this from multiple modules
-check_permissions(RD, Ctx,
-                  _Permission, {_Bucket, undefined}) ->
+resource_forbidden(RD, Ctx, _Perm, {_Resource, undefined}) ->
     {true, wrq:append_to_resp_body("Unknown index", RD), Ctx};
-check_permissions(RD, Ctx,
-                  _Permission, {_Bucket, N}) when is_integer(N) ->
-    {true, wrq:append_to_resp_body("Unknown index", RD), Ctx};
-check_permissions(RD, Ctx,
-                  Permission, {Bucket, Index}) ->
-    check_permissions_aux(RD, Ctx, Permission,
-                          {Bucket, mochiweb_util:unquote(Index)});
-check_permissions(RD, Ctx, Permission, Bucket) ->
-    check_permissions_aux(RD, Ctx, Permission, Bucket).
-
-check_permissions_aux(RD, Ctx=#ctx{security=Security},
-                      Permission, Resource) ->
+resource_forbidden(RD, Ctx=#ctx{security=Security}, Permission,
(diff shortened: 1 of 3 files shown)


src/yz_wm_search.erl
```diff
@@ -76,24 +76,47 @@ is_authorized(ReqData, Ctx) ->
                     "instead.">>, ReqData), Ctx}
     end.
 
+
+%%
+%% Written with the idea of using this from multiple modules
+check_permissions(RD, Ctx,
+                  _Permission, {_Bucket, undefined}) ->
+    {true, wrq:append_to_resp_body("Unknown index", RD), Ctx};
```
rzezeski: This is not an unknown index, it's a failure to specify an index and it will be checked by `resource_exists`. Remove this clause.

rzezeski: Move this function into the `Private` section and mark with `@private`. Delete extra whitespace.

src/yz_wm_search.erl
```diff
@@ -76,24 +76,47 @@ is_authorized(ReqData, Ctx) ->
...
+check_permissions(RD, Ctx,
+                  _Permission, {_Bucket, N}) when is_integer(N) ->
+    {true, wrq:append_to_resp_body("Unknown index", RD), Ctx};
```
rzezeski, line 81: I'm confused by this function clause. Is there ever a case where an integer is passed to this function? All our calls to `path_info` should be for part of the path, not the port. Also, even if this case was possible the error string is misleading. It wouldn't be an unknown index, it would be bad input data. I vote to remove this function clause.

rzezeski:
`check_permissions` is naming the 3rd param `{Bucket, Index}` and `Bucket` but that's not what it is. It is `{Resource, SubResource}` or more generally `{NameSpace1, NameSpace2}`. For this particular resource it is `{<<"index">>, IndexName}`. Here's how I've explained it given what I know about security. We need to be careful about our wording because the scope reaches beyond just buckets.

https://github.com/basho/yokozuna/blob/develop/docs/SECURITY.md#resources-and-permissions

src/yz_wm_search.erl
macintux, line 81: Isn't dialyzer going to be unhappy about the prospect of passing an integer along to `mochiweb_util:unquote`? I wish `path_info` didn't return multiple different types, but as long as it does, seems like we're stuck preparing for that.

rzezeski, line 81: I would think so too but it passed just fine for me. I'll leave it at your discretion.

macintux:
I think it's now much cleaner. Have at it, @rzezeski

rzezeski:
+1 3e37541

I had to apply #362 by hand to get `yz_security` to pass and I saw one failure for `yz_rs_migration` but it passed the next run so I'm considering it not related to this change.

rzezeski:
@borshop merge

PR #364, opened 2014-04-21
Update README notice for beta

rzezeski:
Yokozuna is now tracking Riak releases and thus is in beta stage. Make
it clear that Yokozuna replaces riak_search in 2.0, where the beta may
be downloaded from, and point to our official search documentation.


coderoshi:
I love deleting text, so +1

PR #368, opened 2014-04-23
Add JSON Object Value Generator

rzezeski:
Add `json_obj_valgen` to the Yokozuna BB driver. I used this to test
indexing "large" JSON objects while investigating #358.  The config
line to use this value generator looks like so:

```
{value_generator, {function, yz_driver, json_obj_valgen,
[<NumFields>]}}.
```

This will generate a static JSON object (i.e. generated once at start
up) which contains `<NumFields>` integer fields.  All workers will
index the identical object.  Here is an example config from one of my
testing runs:

```
{mode, max}.
{concurrent, 32}.
{driver, yz_driver}.
{code_paths, ["/root/work/yz-bb/misc/bench"]}.
{secure, false}.
{bucket, {<<"data">>, <<"largeobj">>}}.
{index, <<"largeobj">>}.
{pb_conns, [{"10.0.1.201", 8087}]}.
{http_conns, []}.
{duration, infinity}.
{key_generator, {to_binstr, "~B", {partitioned_sequential_int, 0,
1000}}}.
{value_generator, {function, yz_driver, json_obj_valgen, [4000]}}.
{operations, [{load_pb, 1}]}.
```

This uses 32 workers to write the same JSON object containing 4000
integer fields over the key space 0 to 999.

NOTE: I purposely hard coded the timeout to 90s for the purposes of my
testing.  Yes, ideally this would be configurable but this patch is
about minimal amount of work to achieve the task at hand.  Please keep
this in mind while reviewing.  This is a tool, not production code.

### DIFF
--- a/misc/bench/src/yz_driver.erl
+++ b/misc/bench/src/yz_driver.erl
@@ -169,6 +169,19 @@ run(load_fruit_pb, KeyValGen, _, S=#state{bucket=Bucket, pb_conns=Conns}) ->
         Err -> {error, Err, S2}
     end;
 
+run(load_pb, KeyGen, ValGen, S) ->
+    Bucket = S#state.bucket,
+    Conns = S#state.pb_conns,
+    Conn = get_conn(Conns),
+    Key = KeyGen(),
+    {CT, Val} = ValGen(),
+    Obj = riakc_obj:new(Bucket, Key, Val, CT),
+    S2 = S#state{pb_conns=wrap(Conns)},
+    case riakc_pb_socket:put(Conn, Obj, [{timeout, 90000}], 90000) of
+        ok -> {ok, S2};
+        Err -> {error, Err, S2}
+    end;
+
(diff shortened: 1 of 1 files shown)


lordnull:
Seems to work, after running with similar config (changed to default bucket and loopback IP), I'm able to see very large objects in my riaks. Only remaining concern is dialyzer failure.

rzezeski:
> Only remaining concern is dialyzer failure.

Yea this needs to be rebased.  I'll get to that tomorrow.  Thanks for reviewing this.

rzezeski:
@lordnull Rebased and buildbot shows a good build.

lordnull:
:+1:

rzezeski:
@lordnull you need to +1 the commit for bors to pick it up.

lordnull:
:+1:  f44cee2

rzezeski:
@borshop merge

PR #369, opened 2014-04-23
Security "bucket" is a 2-tuple of binary

rzezeski:
The security API expects the "bucket" param to be a `{binary(),
binary()}`.  I quote the word bucket because security is actually
broader than buckets but the misnomer remains.

This fixes the following dialyzer error:

```
yz_wm_search.erl:84: The call
riak_core_security:check_permission({_,{_,[any()]}},Security::any())
breaks the contract (Permission::permission(),Context::context()) ->
{'true',context()} | {'false',binary(),context()}
```

This is in reaction to riak_core commit
19be73ab706e3aadbda82047d19f07f0ffc43fe7.

### DIFF
--- a/src/yz_wm_search.erl
+++ b/src/yz_wm_search.erl
@@ -82,7 +82,7 @@ resource_forbidden(RD, Ctx, _Perm, {_Resource, undefined}) ->
 resource_forbidden(RD, Ctx=#ctx{security=Security}, Permission,
                    {Resource, Subresource}) ->
     Res = riak_core_security:check_permission(
-            {Permission, {Resource, mochiweb_util:unquote(Subresource)}},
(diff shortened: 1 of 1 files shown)


rzezeski:
To be extra sure I ran the security integration test:

```
Test Results:
yz_security-bitcask: pass
20:44:28.962 [info] Stopped Node... /root/yz-verify/rt/riak-1.4.2/dev/dev1/bin/riak stop ~ wasn't running.
0 Tests Failed
1 Tests Passed
```

@reiddraper If you get a second could you please review. I ping you because you made the core change and you are working on dialyzer stuff anyways.

rzezeski:
@borshop merge

Issue #370, opened 2014-04-28
Sibling objects with curly brace keys are not deleted from YZ when deleted from KV


jhmullen:
If you insert an object into a search-enabled bucket that has curly braces in its key (e.g., '{123}') and that object has siblings, a DELETE issued to that object will remove the object from Riak KV, but NOT from its associated index.  Subsequent sibling creation and deletes will continually grow the quantity of indexed objects in search results.
#### Steps for reproducing:

1) Insert a key with curly braces such as `{123}` into a search-enabled bucket with `allow_mult=true` using the default schema.

2) Insert the same object at least one more time.  

3) Perform an exact match search to observe that numFound=siblingCount (as expected).

4) Delete the object 

5) Perform a GET to observe 404 (correct)

6) Perform search again, noting that numFound=siblingCount instead of 0.  Subsequent runs will increase this number indefinitely.

A simple shell script example is available here:

https://gist.github.com/jhmullen/82c0227ae9b4c9e599b2
#### Expected Behavior:

Deletion of this sibling object from KV should remove all associated indexes in Search.

rzezeski:
The issue is caused by Yokozuna's failure to deal with reserved
characters in the Lucene query syntax [[1]](http://lucene.apache.org/core/2_9_4/queryparsersyntax.html#Escaping%20Special%20Characters).  Not only does the
current code fail to escape them but it also mysteriously URL encodes
the key which is a bug in and of itself.  I have a reproducer and am
working on the fix.

coderoshi:
@rzezeski https://github.com/basho/yokozuna/blob/develop/src/yz_solr.erl#L335 the code does escape the key, to fix the issue with reserved word escapes https://github.com/basho/yokozuna/commit/c147db7f8c9d426a069e639f1e1ab753837da3ad#diff-1933b9092cbac0de03ef337894712bfdR105

rzezeski:
@coderoshi It turns out this is not enough.  The backslash and double
quotes need to be escaped individually.  The binary to list causes
failures for keys with non-latin characters.  Finally the url encoding
should not be done because the keys are not stored url encoded and the
query is being sent as JSON, not url encoded.  I am wrapping up a test
and fix for all this tomorrow.

PR #371, opened 2014-05-08
Fix various issues re deleting documents

rzezeski:
This is a fix for #370. 

Jimmy Mullen, with the help of a user, discovered issues deleting keys
which contain curly braces.  A patch was made recently to wrap the key
in double quotes (Lucene phrase query) but it turns out there were
other issues with the code.
- Don't url encode the key.  The keys are not stored url encoded and
  the query is being sent as a body of JSON, not
  `application/x-www-form-urlencoded`.
- Don't perform a `binary_to_list`.  This breaks keys with non-latin
  characters.
- Don't build up a list when a binary can be used.  This should be
  more efficient than `++`.
- Use the term query parser for future safety.  Just quoting works now
  but future changes like the complex phrase query syntax could cause
  issues.  The use of the term query parser protects the characters
  from parsing (with 2 exceptions noted next).
- Escape any double quote or backslash characters.  The double quote
  must be escaped because double quotes are used to wrap the query.
  Backslash is escaped to prevent Solr from treating it as an escape
  sequence.

Other changes made:
- Modified the sibling test to use a key with various special
  characters and added a case for regular deletion.
- Removed dead code from `yz_solr`.
- Moved `delete_op` type to `yz_solr` and added docs for the various
  types.
- Use cons instead of `++` when building list of fields.

### DIFF
--- a/include/yokozuna.hrl
+++ b/include/yokozuna.hrl
@@ -89,11 +89,6 @@
 -type ring_event() :: {ring_event, riak_core_ring:riak_core_ring()}.
 -type event() :: ring_event().
 
--type delete_op() :: {id, binary()}
-                   | {key, binary()}
-                   | {siblings, binary()}
-                   | {'query', binary()}.
-
 %% @doc The `component()' type represents components that may be
 %%      enabled or disabled at runtime.  Typically a component is
 %%      disabled in a live, production cluster in order to isolate
--- a/riak_test/yz_siblings.erl
+++ b/riak_test/yz_siblings.erl
@@ -1,8 +1,7 @@
 %% @doc Ensure that sibling creations/searching workds.
 -module(yz_siblings).
 -compile(export_all).
--import(yz_rt, [host_entries/1,
-                run_bb/2, search_expect/5,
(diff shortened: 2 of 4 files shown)


rzezeski:
Benchmark results show a potential small dip (1%) in throughput and
raise in median latency.  Building binaries should be more efficient
than list concatenation but that gain is overshadowed by the use of
the two `binary:replace` calls.

| --- | --- | --- | --- |
| current | 1496 | 12.84 ms | 112.48 ms |
| patch | 1482 | 13.21 ms | 110.53 ms |

coderoshi:
- [x] code review
- [x] compile/unit
- [x] dialyzer
- [x] riak tests
- [x] manual testing

```
Test Results:
yz_security-bitcask          : fail
```

Security failure due to expired cert, unrelated to these changes.

coderoshi:
+1 946e618

rzezeski:
Same thing as other issue. Bors wants +1 on last commit.

rzezeski:
@borshop merge

Issue #367, opened 2014-04-23
Increase ibrowse Inactivity Timeout


rzezeski:
After digging into issues #320, #330, and #358 it was discovered that
ibrowse's default `inactivity_timeout` combined with its poor load
balancing algorithm and the default pool and pipeline sizes is causing
unnecessary socket churn.  Even under load the client may not work
Yokozuna enough to fill all 10 pipelines to prevent the connection
from reaching time out.  Increasing the pool or pipeline size just
makes things worse as the ibrowse algorithm first wants to fill the
pool before using pipelines causing it to make a new connection for
almost every request.  My comment on #330 goes into a bit more detail
and includes evidence à la DTrace [[1]](https://github.com/basho/yokozuna/issues/330#issuecomment-40783996).
### Action Items
- [x] verify no socket churn with higher timeout (60s)
- [x] benchmark before/after data load, curious if reduced churn helps

rzezeski:
Pinging @wbrown, let's focus all inactivity timeout research to this issue now.

wbrown:
So, a few of observations that I have now that I've filled up my database and pushed my nodes to the limit.
- The socket timeout appears to really kill Riak `PUT` latency when the throttle is eased back due to my inbound data deduplicating data that's already in Riak.   I see the average `PUT` latency go up to about `65ms`, `99ms` and `216ms` latency.
- Another observation is that with the default sockets of `10`, my searches get bottlenecked by all the `PUT`s being indexed in Yokozuna.  On the client side, I solve a lot of concurrency issues by spinning up multiple socket connections to Riak, and I can do this on a demand basis -- i.e. searches get their own Riak connections.  But Yokozuna indexes and updates are essentially sharing the same connection pool.

This is excaberated by the multithreaded search result paging code that I have.  The basic flow goes:
- Do a small search against Yokozuna with a `maxRows` of `100`.  This helps get small searches done quicker, if they have small result sets.
- Retrieve the `num_found` key, and if it is more than `100`, we divide the remainder into chunks of ranges over `1000` items.
- A certain amount of threaded search workers is spun up, each worker grabs a chunk, and does a streaming search.
- The results are yielded to the search invoker, allowing the resulting keys to be returned as fast as the invoker can process it.

Search performance wise, I get:
- For a search the returns one result, I get an average latency of `13.69ms`.
[...]

PR #375, opened 2014-05-09
Reduce TIME-WAIT load

rzezeski:
This addresses #367 

A combination of ibrowse's inefficient load balancing algorithm and
default socket inactivity timeout of 10 seconds can cause TIME-WAIT
load.  It becomes worse as the pool size of pipeline sizes are
increased.  This patch is a temporary workaround to reduce socket
churn and thus TIME-WAIT load.  It does so by increasing the
inactivity timeout to 600 seconds across the board.  Below is a chart
of showing the amount of socket churn in connections per minute for
the different timeout values both at idle and while under load from
basho bench.  These numbers were calculated by a DTrace script which
counted the number of new connections being accepted on port 8093.

| Timeout | Socket Churn At Idle | Socket Churn Under Load |
| --- | --- | --- |
| 10s | ~59 conns/min | ~29 conns/min |
| 600s | ~9 conns/min | ~6 conns/min |

The timeout is set via application env because ibrowse has the
absolute most complex configuration management code I have ever seen
and this was the easiest way to make sure the timeout is set
correctly.  This is just a workaround until after 2.0 when other HTTP
clients and pools may be tested.  ibrowse seems to have many issues,
this is but just one.

For more background see the following issues:

https://github.com/basho/yokozuna/issues/367
https://github.com/basho/yokozuna/pull/358
https://github.com/basho/yokozuna/issues/330
https://github.com/basho/yokozuna/issues/320

### DIFF
--- a/src/yz_app.erl
+++ b/src/yz_app.erl
@@ -38,6 +38,7 @@ start(_StartType, _StartArgs) ->
     Enabled = ?YZ_ENABLED,
     case yz_sup:start_link(Enabled) of
         {ok, Pid} ->
+            _ = application:set_env(ibrowse, inactivity_timeout, 600000),
             maybe_setup(Enabled),
             {ok, Pid};
         Error ->


rzezeski:
## 600s timeout

Idle:

```
root@oldboy:~# dtrace -s /home/ryan/tcp-conn-rate.d
Monitor rate of ESTABLISHED passive connections over 1 minute............................................................

127.0.0.1:8093 9
```

Load:

```
root@oldboy:~# dtrace -s /home/ryan/tcp-conn-rate.d
Monitor rate of ESTABLISHED passive connections over 1 minute............................................................

127.0.0.1:8093 6
```

## 10s timeout

Idle:

```
root@oldboy:~# dtrace -s /home/ryan/tcp-conn-rate.d
Monitor rate of ESTABLISHED passive connections over 1 minute............................................................

127.0.0.1:8093 59
```

Load:

```
root@oldboy:~# dtrace -s /home/ryan/tcp-conn-rate.d
Monitor rate of ESTABLISHED passive connections over 1 minute............................................................

127.0.0.1:8093 29
```

rzezeski:
@wbrown If you get a chance you might want to give this PR a spin.  It is not a perfect patch but it will have to be enough until after 2.0 is out and we can look at replacing the entire HTTP client.

src/yz_app.erl
```diff
@@ -38,6 +38,7 @@ start(_StartType, _StartArgs) ->
     Enabled = ?YZ_ENABLED,
     case yz_sup:start_link(Enabled) of
         {ok, Pid} ->
+            _ = application:set_env(ibrowse, inactivity_timeout, 600000),
```
coderoshi, line 41: This is 600 secs, since I assume this is ms

rzezeski, line 41: (facepalm) This is what I get for leaving a branch for over a week.  Yea, so 600s is fine too. I actually wanted to set it to infinity but ibrowse is has a bug where if you set the value via app env it won't accept `infinity` (but it does if you set it other ways, but those other ways only work for some connections...ibrowse literally has the most asinine code for its configuration I want to break shit every time I think about it).

rzezeski:
I amended the commit message to say 600 seconds.

coderoshi:
@rzezeski could you post `tcp-conn-rate.d` in a gist? I'd like to test this.

rzezeski:
> @rzezeski could you post tcp-conn-rate.d in a gist? I'd like to test this.

See the end of this comment: https://github.com/basho/yokozuna/issues/330#issuecomment-40783996

coderoshi:
+1 b02c0bd

I saw a similar pattern running this dtrace.

**600s timeout Idle**

```
127.0.0.1:10014 9
```

**600s timeout Load**

```
127.0.0.1:10014 9
```

**10s timeout Idle**

```
127.0.0.1:10014 58
```

**10s timeout Load**

```
127.0.0.1:10014 55
```

I had to make a small change to the script to work in OSX (`tcp:::accept-established` doesn't exist in OSX)

```
#pragma D option quiet

dtrace:::BEGIN
{
    printf("Monitor rate of ESTABLISHED passive connections over 1 minute");
}

tcp:::state-change
/args[2]->tcps_state == "state-established"/
{
    @est[args[1]->ip_saddr, args[1]->sport] = count();
}

profile:::tick-1sec
{
    printf(".");
}

profile:::tick-60sec
{
    printf("\n\n");
    printa("%s:%d %@d\n", @est);
    exit(0);
}
```

rzezeski:
@borshop merge

PR #379, opened 2014-05-12
Switched to generated certs.

lordnull:
yz_security test generates certs, so if they do expire one can simple
clear the r_t scratch dir and re-run the test.


kellymclaughlin:
+1  714534a

PR #381, opened 2014-05-13
Constrain index names to ASCII to duck latin-1/UTF-8 conflicts

macintux:
YOLOPR: no docs, no tests, not even any manual tests run other than compiling verify_name separately.

(I'll add tests to the PR before it's ready for final review. Or let someone who actually has touched YZ do so.)

### DIFF
--- a/src/yz_index.erl
+++ b/src/yz_index.erl
@@ -262,7 +262,8 @@ schema_name(Info) ->
 %%      UTF-8 support is available
 -spec verify_name(index_name()) -> {ok, index_name()} | {error, invalid_name}.
 verify_name(Name) ->
-    case lists:dropwhile(fun(X) -> X < 128 end, Name) =:= "" of
+    case lists:dropwhile(fun(X) -> X < 128 end,
+                         binary_to_list(Name)) =:= "" of
         true ->
             case re:run(Name, "/", []) of
                 nomatch ->   {ok, Name};
--- a/src/yz_index.erl
+++ b/src/yz_index.erl
@@ -56,7 +56,7 @@ create(Name) ->
     create(Name, ?YZ_DEFAULT_SCHEMA_NAME).
 
 %% @see create/3
--spec create(index_name(), schema_name()) -> 
+-spec create(index_name(), schema_name()) ->
                     ok |
                     {error, schema_not_found} |
                     {error, invalid_name}.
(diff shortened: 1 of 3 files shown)


src/yz_index.erl
```diff
@@ -258,12 +258,19 @@ schema_name(Info) ->
...
 %% @doc Verify that the index is a name that Solr can use. Some chars
-%%      are invalid, namely "/".
+%%      are invalid, namely "/" or non-ascii characters until full
+%%      UTF-8 support is available
 -spec verify_name(index_name()) -> {ok, index_name()} | {error, invalid_name}.
 verify_name(Name) ->
-    case re:run(Name, "/", []) of
-        nomatch ->   {ok, Name};
-        {match,_} -> {error, invalid_name}
+    case lists:dropwhile(fun(X) -> X < 128 end,
```
coderoshi: As long as we're being hard nosed, may as well disallow < 32

coderoshi:
- [x] code review
- [x] unit test
- [x] dialyzer (it's broken, but not related to this PR in any way)
- [x] riak test
- [x] documentation (basho/basho_docs@7b9a071)

rzezeski:
FWIW, you can use binary comprehensions instead of converting to a list.

macintux:
Interesting idea, but `<< <<X>> || <<X>> <= Name, X > 127 >> =:= <<"">>` isn't quite as readable, methinks.

src/yz_index.erl
macintux: Your wish is my command, @coderoshi

macintux:
My original idea was that this would be a failsafe PR, to be deployed only if we didn't tackle UTF-8 indexes, but the more I think about it the less likely it seems there'll be time to do the real work. Should we just plan on going with this and worrying about the real fix for a future release?

coderoshi:
@macintux I wrote a unit test for this, which you might want to add to one of the `tests`.

```
should_verify_name_test() ->
    ?assertEqual({ok, <<"just-fine">>}, yz_index:verify_name(<<"just-fine">>)),
    ?assertEqual({error, invalid_name}, yz_index:verify_name(<<"bad/slash">>)),
    ?assertEqual({error, invalid_name}, yz_index:verify_name(<<"out-of-range-", 129>>)).
```

coderoshi:
+1 1f3745f

rzezeski:
Tests should be added to this PR before it is merged.

macintux:
Added Eric's test with a couple of extras.

coderoshi:
+1 04abef5

macintux:
@borshop merge

PR #385, opened 2014-05-19
Fix deletion for many-to-one indexing

### LINKED ISSUE #376: deletion broken for many-to-one indexing
Deletion is broken for many-to-one indexing.  Yokozuna deletes based
on Riak Key only.  If multiple types or buckets point to the same
index then it is possible for the same key to exist multiple times.
The current code would cause all instances of the key to be deleted.

Any deletes performed as queries (i.e. `{key,Key}` or
`{siblings,Key}`) must be updated to include the type and bucket name.

rzezeski:
Fixes #376.

It is possible to point multiple types and buckets to the same index.
The object key alone is not a unique identifier as the same key could
be repeated across types and buckets.

This patch fixes the deletion code to delete on the actual unique key
which is type + bucket + key.  Other changes made include:
- Expand the essential test to verify deletion and sibling cleanup
  work correctly when using the same key across multiple types and
  buckets.  Add unicode and special chars to additional verification.
- Update the AAE test to to pass the full key when making remote call
  to `yz_solr:delete`.
- Use the ?DEFAULT_TYPE macro in case the default type is changed
  before 2.0.0.
- Change `yz_rt:random_keys` to return binary keys.
- Handle legacy objects (pre 2.0.0) in `encode_delete`.
- Make sure to pass the lager transform when compiling the Yokozuna
  driver and make the `build_bb_driver` function more robust in
  general.

### DIFF
--- a/riak_test/aae_test.erl
+++ b/riak_test/aae_test.erl
@@ -48,7 +48,7 @@ confirm() ->
             yz_rt:wait_for_full_exchange_round(Cluster, TS1),
             RepairCountBefore = get_cluster_repair_count(Cluster),
             yz_rt:count_calls(Cluster, ?REPAIR_MFA),
-            Keys = yz_rt:random_keys(?NUM_KEYS),
+            Keys = [{?BUCKET,K} || K <- yz_rt:random_keys(?NUM_KEYS)],
             {DelKeys, _ChangeKeys} = lists:split(length(Keys) div 2, Keys),
             lager:info("Deleting ~p keys", [length(DelKeys)]),
             [delete_key_in_solr(Cluster, ?INDEX, K) || K <- DelKeys],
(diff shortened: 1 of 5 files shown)


rzezeski:
This also addresses my comment on #351.

https://github.com/basho/yokozuna/pull/351#issuecomment-43517810

coderoshi:
Good call on using ?DEFAULT_TYPE. There are a couple other modules that still hard-code <<"default"">>, fwiw:
- src/riak_object_json.erl
- riak_test/yz_schema_admin.erl

coderoshi:
- [x] Code Review
- [x] Compile/Unit Test
- [x] Dialyzer
- [x] Riak Test
- [x] Manual Test

```
Test Results:
0 Tests Failed
18 Tests Passed
```

coderoshi:
+1 35dd283

rzezeski:
@borshop merge

Yes, those other places should be updated too.  I'll change them in my next PR to avoid holding up this merge.

PR #386, opened 2014-05-20
Fix sort param in `search_fold` to guarantee unique values

### LINKED ISSUE #356: map-reduce must sort on type-bucket-key
Make sure the map-reduce support is sorting on type-bucket-key so that
all unique keys are visited. See issue #355 for more information.

rzezeski:
Fixes #356.

The sort parameter must be type/bucket/key, as described in issue
duplicate values.  Other modifications include:
- Make sure to pass the new accumulator after calling `F` in
  `search_fold/9`.
- Increase the page size to 100 so less HTTP requests are made.  This
  should result in lower latency.
- Modify `yz_mapreduce` to use 1000 keys so since the page size was
  increased to 100.
- Add a test case to `yz_mapreduce` to verify that all 1000 unique
  results are returned.
- Add `yz_rt:load_module/2` to be used for loading a module's BEAM
  onto a set of nodes.  This is needed by the new test to have the HOF
  passed to `search_fold` available on the server process.

### DIFF
--- a/src/yokozuna.erl
+++ b/src/yokozuna.erl
@@ -131,8 +131,9 @@ search_fold(Index, Query, Filter, F, Acc) ->
     Params = [{q, Query},
               {fq, Filter},
               {start, Start},
-              {rows, 10},
+              {rows, 100},
               {fl, <<?YZ_RT_FIELD_S,",",?YZ_RB_FIELD_S,",",?YZ_RK_FIELD_S>>},
+              {sort, <<?YZ_RT_FIELD_S," asc, ",?YZ_RB_FIELD_S," asc, ",?YZ_RK_FIELD_S," asc">>},
               {omitHeader, <<"true">>},
               {wt, <<"json">>}],
     {_, Body} = yz_solr:dist_search(Index, Params),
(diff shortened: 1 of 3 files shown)


rzezeski:
If you want to verify that `search_fold` (and thus map-reduce) was broken before this patch then simply apply this diff and run the `yz_mapreduce` test.

```
diff --git a/src/yokozuna.erl b/src/yokozuna.erl
index a3c3682..082910f 100644
--- a/src/yokozuna.erl
+++ b/src/yokozuna.erl
@@ -133,7 +133,7 @@ search_fold(Index, Query, Filter, F, Acc) ->
               {start, Start},
               {rows, 100},
               {fl, <<?YZ_RT_FIELD_S,",",?YZ_RB_FIELD_S,",",?YZ_RK_FIELD_S>>},
-              {sort, <<?YZ_RT_FIELD_S," asc, ",?YZ_RB_FIELD_S," asc, ",?YZ_RK_FIELD_S," asc">>},
+              %% {sort, <<?YZ_RT_FIELD_S," asc, ",?YZ_RB_FIELD_S," asc, ",?YZ_RK_FIELD_S," asc">>},
               {omitHeader, <<"true">>},
               {wt, <<"json">>}],
```

coderoshi:
Reviewing

coderoshi:
- [x] code review
- [x] compile/unit test
- [x] riak test
- [x] mr test fails as expected with sort removed
- [x] dialyzer

```
Test Results:
0 Tests Failed
18 Tests Passed
```

src/yokozuna.erl
```diff
@@ -219,10 +220,10 @@ positions({struct,[X, Y, Z]}) ->
 search_fold([], _, _, _, _, _, _, _, Acc) ->
     Acc;
 search_fold(Results, Start, Params, Positions, Index, Query, Filter, F, Acc) ->
-    F(Results, Acc),
-    Start2 = Start + 10,
+    Acc2 = F(Results, Acc),
+    Start2 = Start + 100,
```
coderoshi, line 224: Magic number

coderoshi:
+1 c2235b6

Might be good to define that magic number, but it's a drop in the bucket for Riak

rzezeski:
@borshop merge

Yea that magic number eventually needs to be replaced I just didn't want to scope creep this PR.

PR #396, opened 2014-06-02
Correctly set Jetty temp directory

### LINKED ISSUE #387: Look at this issue in node_package
https://github.com/basho/node_package/issues/138

I don't feel we should have to create any directories as part of the packaging.  I thought jetty/solr was creating a relative dir.  Needs investigating...

rzezeski:
Fixes #387.

A custom `solr-jetty-context.xml` must be used so that the
`tempDirectory` is set properly.  `jetty.home` is a system property,
not a regular property.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,9 +2,15 @@
 deps
 ebin
 priv/java_lib
-priv/solr*
 priv/conf/lang
 priv/conf/*.txt
+priv/solr/etc/create-solrtest.keystore.sh
+priv/solr/etc/webdefault.xml
+priv/solr/lib
+priv/solr/resources
+priv/solr/start.jar
+priv/solr/webapps
+priv/solr/solr-webapp
 build/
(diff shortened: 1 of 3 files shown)


rzezeski:
You can test this by starting a Riak node without this change and notice that `solr-webapp` is under the root dir of riak (i.e. `dev/dev#` or `rel/riak`).  With this change `solr-webapp` will be under `{root}/lib/yokozuna/priv/solr/solr-webapp`.

coderoshi:
Looks good. Just needs a rebase then +1

rzezeski:
@coderoshi rebased

coderoshi:
+1 fd37257

rzezeski:
@reiddraper what does it mean when borshop fails to merge into its integration branch?  A merge conflict?

rzezeski:
@borshop merge

Issue #395, opened 2014-05-29
Don't list index if it failed to create


rzezeski:
If you create an index with a bad schema Solr will fail to create the core but the index will be listed if you perform `GET /search/index` (or equivalent PB msg).  These two code paths should filter the metadata index list by checking for existence.

See basho/riak#550.

coderoshi:
Closing, PR exists at #397

PR #397, opened 2014-06-02
Filter indexes that do not exist on Solr

coderoshi:
Fix issue #395

> If you create an index with a bad schema Solr will fail to create the core but the index will be listed if you perform GET /search/index (or equivalent PB msg). These two code paths should filter the metadata index list by checking for existence.

### DIFF
--- a/riak_test/yz_index_admin.erl
+++ b/riak_test/yz_index_admin.erl
@@ -96,12 +96,50 @@
 </types>
 </schema>">>).
 
+-define(SCHEMA_FIELDS_DOUBLE,
+        <<"<?xml version=\"1.0\" encoding=\"UTF-8\" ?>
+<schema name=\"test\" version=\"1.5\">
+<fields>
+   <field name=\"_yz_id\" type=\"_yz_str\" indexed=\"true\" stored=\"true\" required=\"true\"  multiValued=\"false\"/>
+   <field name=\"_yz_ed\" type=\"_yz_str\" indexed=\"true\" multiValued=\"false\"/>
+   <field name=\"_yz_pn\" type=\"_yz_str\" indexed=\"true\" multiValued=\"false\"/>
+   <field name=\"_yz_fpn\" type=\"_yz_str\" indexed=\"true\" multiValued=\"false\"/>
+   <field name=\"_yz_vtag\" type=\"_yz_str\" indexed=\"true\" multiValued=\"false\"/>
+   <field name=\"_yz_rt\" type=\"_yz_str\" indexed=\"true\" stored=\"true\" multiValued=\"false\"/>
+   <field name=\"_yz_rk\" type=\"_yz_str\" indexed=\"true\" stored=\"true\" multiValued=\"false\"/>
+   <field name=\"_yz_rb\" type=\"_yz_str\" indexed=\"true\" stored=\"true\" multiValued=\"false\"/>
+   <field name=\"_yz_err\" type=\"_yz_str\" indexed=\"true\" stored=\"true\" multiValued=\"false\"/>
+   <field name=\"a_field\" type=\"_yz_str\" indexed=\"true\" stored=\"true\" />
+   <field name=\"a_field\" type=\"_yz_str\" indexed=\"true\" stored=\"true\" />
+</fields>
+
+ <uniqueKey>_yz_id</uniqueKey>
+
+<types>
+    <fieldType name=\"_yz_str\" class=\"solr.StrField\" sortMissingLast=\"true\" />
+    <fieldType name=\"text_general\" class=\"solr.TextField\" positionIncrementGap=\"100\">
+      <analyzer type=\"index\">
+        <tokenizer class=\"solr.StandardTokenizerFactory\"/>
+        <filter class=\"solr.StopFilterFactory\" ignoreCase=\"true\" words=\"stopwords.txt\" enablePositionIncrements=\"true\" />
+        <filter class=\"solr.LowerCaseFilterFactory\"/>
+      </analyzer>
+      <analyzer type=\"query\">
+        <tokenizer class=\"solr.StandardTokenizerFactory\"/>
+        <filter class=\"solr.StopFilterFactory\" ignoreCase=\"true\" words=\"stopwords.txt\" enablePositionIncrements=\"true\" />
+        <filter class=\"solr.SynonymFilterFactory\" synonyms=\"synonyms.txt\" ignoreCase=\"true\" expand=\"true\"/>
+        <filter class=\"solr.LowerCaseFilterFactory\"/>
+      </analyzer>
+    </fieldType>
+</types>
+</schema>">>).
+
 confirm() ->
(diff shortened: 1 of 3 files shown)


riak_test/yz_index_admin.erl
```diff
@@ -181,6 +209,28 @@ confirm_409(Cluster) ->
...
+%% @doc Test index creation with a broken schema
+confirm_create_index_bad_schema(Cluster) ->
+    Index = <<"test_index_bad_schema">>,
+    HP = select_random(host_entries(rt:connection_info(Cluster))),
+    lager:info("confirm_create_index_bad_schema ~s [~p]", [Index, HP]),
+
+    lager:info("upload schema ~s [~p]", [Index, HP]),
+    SchemaURL = schema_url(HP, Index),
+    SchemaHeaders = [{"content-type", "application/xml"}],
+    {ok, Status1, _, _} = http(put, SchemaURL, SchemaHeaders, ?SCHEMA),
```
rzezeski: You're not actually using the bad schema here.  Furthermore, the bad schema will be rejected by the schema admin since it doesn't contain the `_yz` fields.  How about using the original issue of a repeated field.

riak_test/yz_index_admin.erl
```diff
@@ -181,6 +209,28 @@ confirm_409(Cluster) ->
...
+    ?assertEqual("204", Status1),
+
+    URL = index_url(HP, Index),
+    Headers = [{"content-type", "application/json"}],
+    Body = <<"{\"schema\":\"bad_schema\"}">>,
```
rzezeski: This needs to be `Index`, not `"bad_schema"`.

riak_test/yz_index_admin.erl
```diff
@@ -181,6 +209,28 @@ confirm_409(Cluster) ->
...
+    {ok, Status, _, _} = http(put, URL, Headers, Body),
+    ?assertEqual("400", Status),
+    %% yz_rt:wait_for_index(Cluster, Index),
+    {ok, GetStatus, _, _} = http(get, URL, [], []),
```
rzezeski: The original issues is that the list of _all_ indexes includes names which were not actually created by Solr.  This test should verify both the 404 on the direct get as well as verify the list of indexes does not include the index.

riak_test/yz_index_admin.erl
```diff
@@ -181,6 +209,28 @@ confirm_409(Cluster) ->
...
+    SchemaHeaders = [{"content-type", "application/xml"}],
+    {ok, Status1, _, _} = http(put, SchemaURL, SchemaHeaders, ?SCHEMA),
+    ?assertEqual("204", Status1),
+
+    URL = index_url(HP, Index),
+    Headers = [{"content-type", "application/json"}],
+    Body = <<"{\"schema\":\"bad_schema\"}">>,
+    {ok, Status, _, _} = http(put, URL, Headers, Body),
+    ?assertEqual("400", Status),
+    %% yz_rt:wait_for_index(Cluster, Index),
```
rzezeski: Remove commented code.

riak_test/yz_index_admin.erl
coderoshi: The test of the list is done in `confirm_list`. This part adds a bad index, and `confirm_list` checks that it doesn't appear in the list.

rzezeski: Ah yes, I forgot that's how this test is structured.  The other two comments still stand.  The test passes but it's not actually testing the right thing.

rzezeski:
@coderoshi This needs to be rebased and squashed.

riak_test/yz_index_admin.erl
```diff
@@ -181,6 +219,26 @@ confirm_409(Cluster) ->
...
     ?assertEqual("409", Status2).
 
+%% @doc Test index creation with a broken schema
+confirm_create_index_bad_schema(Cluster) ->
+    Index = <<"test_index_bad_schema">>,
+    Schema = <<"bad_schema">>,
+    HP = select_random(host_entries(rt:connection_info(Cluster))),
+    lager:info("confirm_create_index_bad_schema ~s [~p]", [Index, HP]),
+
+    lager:info("upload schema ~s [~p]", [Index, HP]),
```
rzezeski: This should be `Schema`.

riak_test/yz_index_admin.erl
```diff
@@ -181,6 +219,26 @@ confirm_409(Cluster) ->
...
+    SchemaURL = schema_url(HP, Schema),
+    SchemaHeaders = [{"content-type", "application/xml"}],
+    {ok, Status1, _, _} = http(put, SchemaURL, SchemaHeaders, ?SCHEMA_FIELDS_DOUBLE),
+    ?assertEqual("204", Status1),
+
+    URL = index_url(HP, Index),
+    Headers = [{"content-type", "application/json"}],
+    Body = <<"{\"schema\":\"",Schema/binary,"\"}">>,
+    {ok, Status, _, _} = http(put, URL, Headers, Body),
+    ?assertEqual("400", Status),
```
rzezeski: This will return a `204`, not `400`.  While we are at it we should verify a `404` for `GET /search/index/Index`.

```
20:27:42.242 [error]
================ yz_index_admin failure stack trace =====================
{{assertEqual_failed,[{module,yz_index_admin},
                      {line,239},
                      {expression,"Status"},
                      {expected,"400"},
                      {value,"204"}]},
 [{yz_index_admin,'-confirm_create_index_bad_schema/1-fun-3-',2,
                  [{file,"riak_test/yz_index_admin.erl"},{line,239}]},
  {yz_index_admin,confirm_create_index_bad_schema,1,
                  [{file,"riak_test/yz_index_admin.erl"},{line,239}]},
  {yz_index_admin,confirm,0,
                  [{file,"riak_test/yz_index_admin.erl"},{line,142}]},
  {riak_test_runner,return_to_exit,3,
                    [{file,"src/riak_test_runner.erl"},{line,159}]}]}
=========================================================================
```

coderoshi: Bah merge accident.

I purposefully removed the 404 test, because it would be 404 in any case, even if it were successfully created, and we can't do a wait until. I can add it back, but it didn't seem useful

rzezeski: I think the 404 check is worth it to verify that the index was not created.  A lot of the checks in these tests are obvious but they are there to make sure assumptions are correct and act as documentation as to the semantics of the test.

rzezeski:
Okay, found a better way than sleeping:

``` erlang
    InitFailure = fun(Node) ->
                          lager:info("Waiting for init failure of ~s [~p]", [Index, Node]),
                          {ok,_,S} = rpc:call(Node, yz_solr, core, [status, [{wt,json},{core,Index}]]),
                          {struct,[]} /= kvc:path([<<"initFailures">>], mochijson2:decode(S))
                  end,
    yz_rt:wait_until(Cluster, InitFailure),
```

rzezeski:
+1 e587597

coderoshi:
@borshop merge

PR #400, opened 2014-06-03
Remove obsolete comment

rzezeski:
This comment is no longer valid.  The reason AAE was being triggered
after node restarts was because Riak Test was hard killing (kill -9)
the cluster causing corruption and thus causing AAE to go nuts.

This is related to #248.


coderoshi:
+1 d794ebe

Yes, this seems to delete that comment

rzezeski:
@borshop merge

PR #409, opened 2014-07-01
Fix failure with yz_dt_test due to a disable operation in a map

kellymclaughlin:
It is currently invalid to disable a flag in a newly created map with
no context. The use of disable on a flag is only valid after the map
has been created, written to Riak, and read from Riak. The context is
then established and the operation is able to complete. The fix for
the test for right now is to remove the disable operation on the flag
for the "Russell" map in the test so that no invalid operation are
attempted.

### DIFF
--- a/riak_test/yz_dt_test.erl
+++ b/riak_test/yz_dt_test.erl
@@ -102,7 +102,6 @@ map_update(PB) ->
                             fun(Reg) ->
                                     riakc_register:set(<<"Russell Brown">>, Reg)
                             end},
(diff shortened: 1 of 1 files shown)


kellymclaughlin:
This fix addresses the riak_test failure demonstrated [here](http://giddyup.basho.com/#/projects/riak_ee/scorecards/105/105-1749-yz_dt_test-centos-6-64/53856/artifacts/921043)

andrewjstone:
:+1:  6b7c7b3

Test passes, code looks correct.

kellymclaughlin:
@borshop merge

PR #410, opened 2014-07-02
Ensure uniqueness of _yz_id

### LINKED ISSUE #406: In rare cases _yz_id is not unique
Currently the unique id, `_yz_id` is built like so:

``` erlang
<<BType/binary,"_",BName/binary,"_",Key/binary,"_",Partition/binary>>.
```

Or if a sibling value:

``` erlang
<<BType/binary,"_",BName/binary,"_",Key/binary,"_",Partition/binary,"_",Sibling/binary>>.
```

Full code: https://github.com/basho/yokozuna/blob/develop/src/yz_doc.erl#L52

Given the right combination of type and bucket names, all indexing under the **SAME INDEX**, there could be unique KV objects which collide on `_yz_id`.  This is bad because `_yz_id` is used as the unique id in Solr.  When two writes are done to the same unique id the later one wins.  In the case of a collision this could mean one of the object's indexes would be missing from Solr.

Here is an example of such a combination:

Type1 = `type_bucket`
Bucket1 = `yay`
Type2 = `type`
Bucket2 = `bucket_yay`
Key = `k`
Partition = `0`

ID1 = `type_bucket_yay_k_0`
ID2 = `type_bucket_yay_k_0`

Thanks to @engelsanchez for finding this bug.

coderoshi:
Fixes #406 

Change separator to be an asterisk. Encode types/buckets/keys (TBKs) `*` (%1) and `%` (%2) chars. This will halt any chance of `*` overlapping ids, and also encoding `%` will deal with the case of someone having the literal string `%1` in an id (it would encode to `%21`). And since `*` and `%` are relatively less commonly used characters for TBKs, it will reduce the cases of encoding being occurring.

We also add a numeric prefix to track the given _yz_id version. We can use this to upgrade doc indexes in the future.

Type1 = `type*bucket`
Bucket1 = `yay`
Type2 = `type`
Bucket2 = `bucket*yay`
Key = `k`
Partition = `0`

ID1 = `1*type%2Abucket*yay*k*0`
ID2 = `1*type*bucket%2Ayay*k*0`

### DIFF
--- a/src/yz_doc.erl
+++ b/src/yz_doc.erl
@@ -23,6 +23,8 @@
 -include("yokozuna.hrl").
 
 -define(MD_VTAG, <<"X-Riak-VTag">>).
+-define(YZ_ID_SEP, "*").
+-define(YZ_ID_VER, "1").
 
 %% @doc Functionality for working with Yokozuna documents.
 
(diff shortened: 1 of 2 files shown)


andrewjstone:
:+1: 9ecf80e
Code looks good and yokozuna_essential passes.

seancribbs:
@borshop merge

src/yz_doc.erl
```diff
@@ -72,6 +74,15 @@ make_doc(O, Hash, {MD, V}, FPN, Partition) ->
...
     {doc, lists:append([Tags, ExtractedFields, Fields])}.
 
+%% @private
+%% @doc encode `*' as %1, and `%' as %2. This is so we can reasonably use
+%% the char `*' as a seperator for _yz_id parts (TBK+partition[+sibling]).
+%% This removes the potential case where `*' included in a part can break
+%% _yz_id uniqueness
+-spec encode_doc_part(binary()) -> iolist().
+encode_doc_part(Part) ->
+    re:replace(re:replace([Part],"[%]","%2",[global]), "[*]","%1",[global]).
```
jonmeredith, line 84: Compiling/executing six regular expression for every doc id component seems painful (Two REs, called 3 times)

```
encode_doc_part([], Acc) ->
  lists:reverse(Acc);
encode_doc_part([\* | Rest], Acc) ->
  encode_doc_part(Rest, ["1%" | Acc]);
encode_doc_part([\% | Rest], Acc) ->
  encode_doc_part(Rest, ["2%" | Acc]).
```

Seems equivalent and lighter weight

jonmeredith:
@coderoshi @andrewjstone I have some nitpicky performance concerns - they may well be eaten in the marshaling/HTTP requestery, but still..

Issue #403, opened 2014-06-19
Better Index/Schema Administration [JIRA: RIAK-1688]


rzezeski:
This is a dump of the ideas I have for better index/schema
administration in Yokozuna.
## Current State of Affairs

The current schema and index administration APIs mostly work but allow
too many ways for users to shoot themselves in the foot.  Many users
have already found the API complex and awkward.

Both HTTP and PB allow the storing and modification of schemas.  A
schema is a name along with a blob of XML.  This XML will be handed
off to Solr verbatim.

A user may also create and delete indexes over HTTP or PB.  When
creating an index an N value and schema name may be supplied.  If no
schema is specified the built in default is used named `_yz_default`.
## Steps During Index Creation

This is an overview of the steps taken by Yokozuna during index
creation.  Index creation is async and proceeds in two stages.  The
first is the user request and the second is the creation of the Solr
Cores across the cluster (in Solr, indexes are called Cores).
### user request
1. If the index already exists then just return success, otherwise
   proceed.
2. If the schema exists then proceed, otherwise return an error saying
   the schema doesn't exist.
3. Write an entry to cluster-meta for the index.
4. Return success.

See: `yz_wm_index:maybe_create_index`, `yz_pb_admin:maybe_create_index`, `yz_index:create`
### core creation

Every node has an event server with a periodic tick upon which it
performs various tasks including checking the cluster-meta for new
index entries.  This is the `yz_events` server, the name comes from
the days when Yokozuna used ring events for notification which it no
longer does.

Remember, these steps are occurring on _each_ node, even the one that
serviced the user request.
1. Tick fires.
2. If Solr is up and either a) the cluster-meta hash for yz indexes
   changed or b) this tick represents a "full check" then continue.
3. Grab the list of cores from the local Solr node.
4. Grab the list of indexes from the cluster-meta.
5. Calculate a delta of removed, added and those that stayed the same.
6. Remove indexes that were removed.
7. Combine the added and same, for any that don't exist try to create
   them locally and continue to next step.
8. Create the directory structure needed under data dir and copy
   various files, including the schema file and `solrconfig.xml`.
9. Send a request to Solr to create the core.  Log return from Solr,
   whether success or failure.

Since this second stage is based on a tick per node that means that a)
the actual local creation of the core may occur at different times
across and cluster, b) core creation is async from the user request
and c) if a core creation fails it will retry to infinity until the
failure is fixed (normally this comes in the form of a bad schema).

[...]

Issue #130, opened 2013-06-19
Add schema update + index reloading [JIRA: RIAK-1591]


rzezeski:
After a conversation with someone on IRC it came to my attention that Yokozuna should support schema updates + index reloading sooner rather than later.  The user was performing these basic steps that would better be handled by Yokozuna itself.
1. Modify the schema.
2. Copy the schema to /data/yz/<index name>/conf/<schema name>.xml on each node.
3. Reboot each Riak node to reload the schema.
4. New objects can use new fields.  New fields can be queried on.

This could be done in Yokozuan like so:
1. Modify schema.
2. Update in Yokozuna: `curl -X PUT -H 'content-type: application/json' 'http://host:port/yz/schema/<schema name>' --data-binary @<schema name>.xml`
3. Tell Yokozuna to reload the index: `curl -X POST -H 'content-type: text/plain' 'http://host:port/yz/index/<index name>' RELOAD`

This would use Solr's core RELOAD command and wouldn't require a restart.

arobson:
@rzezeski - so I've been working on getting yokozuna support for an upcoming feature and having to bounce the cluster every time we add a schema or make a schema change is a real concern. Any chance this might land in the 2.0 release proper?

rzezeski:
Yes, this should land in 2.0 proper.

rzezeski:
Some thought needs to be giving to Solr's new managed schema API.

http://wiki.apache.org/solr/SchemaRESTAPI

Should Yokozuna try to mimic this? I'm not sure this plays well with Yokozuna's distributed bits.

Consideration for #219 must also be given. If schemaless mode is used the schema is updating automatically in Solr itself. How do Yokozuna and Solr keep a consistent view of what the schema is for an index?

rzezeski:
The server now has the ability to reload indexes on the fly thanks to #260. But there is still no HTTP, PB, or riak-admin interface yet.

krestenkrab:
This is a real issue for the usability of Yokozuna, as developing anything usually requires modifying the schema as you go along.  We really need at least HTTP access to this, so it can be scripted.

Also, it's a bit unclear how to push an updated schema.  If you just re-PUT an updated schema to the HTTP `/search/schema/schemaName` url then the schema does not get rewritten into the corresponding `indexName/conf` directory for indexes that use the schema.  That would seem like an obvious semantics, I was a bit baffled about this at least.

[...]

PR #427, opened 2014-10-24
WIP: Update ED value to include schema hash

coderoshi:
Adding schema hash in the entropy data value so that changes to the schema are updated by AAE.

Related to #130 and #403.

Also added #436 since I was changing the ED anyway.


AliakseiMat:
In what riak version this feature will be available?

zeeshanlakhani:
@AliakseiMat we're currently working on other work to increase performance (and upgrade Solr to 4.10.4) for 2.2.

PR #438, opened 2014-12-03
Compile bb before misc/bench compile

coderoshi:
A change in basho_bench (basho/basho_bench@8fd27d3) now requires it be compiled before the rest of the bench code can be.


zeeshanlakhani:
:metal:

PR #447, opened 2015-01-27
Handled missing maxScore value from Solr #447 [JIRA: RIAK-1498]

angrycub:
When a sort is set, maxScore is not set causing `kvc:path([<<"maxScore">>]` to return an empty list.  This causes the encode to break as it expects a float.  Setting maxScore to 0.0 is a total bodge to provide a minimal surface area fix.

### DIFF
--- a/src/yz_pb_search.erl
+++ b/src/yz_pb_search.erl
@@ -84,7 +84,10 @@ maybe_process(true, #rpbsearchqueryreq{index=Index}=Msg, State) ->
 				R = mochijson2:decode(Body),
 				Resp = yz_solr:get_response(R),
 				Pairs = yz_solr:get_doc_pairs(Resp),
-				MaxScore = kvc:path([<<"maxScore">>], Resp),
+				MaxScore = case kvc:path([<<"maxScore">>], Resp) of
+					[] -> 0.0;
+					Score -> Score
+				end,
 				NumFound = kvc:path([<<"numFound">>], Resp),
 
 				RPBResp = #rpbsearchqueryresp{


src/yz_pb_search.erl
```diff
@@ -84,7 +84,10 @@ maybe_process(true, #rpbsearchqueryreq{index=Index}=Msg, State) ->
 				R = mochijson2:decode(Body),
 				Resp = yz_solr:get_response(R),
 				Pairs = yz_solr:get_doc_pairs(Resp),
-				MaxScore = kvc:path([<<"maxScore">>], Resp),
+				MaxScore = case kvc:path([<<"maxScore">>], Resp) of
```
coderoshi, line 87: kvc:path accepts a default value. How about: 

``` erlang
kvc:path([<<"maxScore">>], Resp, 0.0),
```

zeeshanlakhani:
good point @coderoshi, as per https://github.com/etrepum/kvc/blob/5565fe51857747662410cc3c06362ebcf48a2f04/src/kvc.erl#L32. @angrycub if you make that update, I'll :+1:.

Basho-JIRA:
Comment for GitHub. 

**_[posted via JIRA by Derek Somogyi]**_

src/yz_pb_search.erl
zeeshanlakhani, line 87: Realizing we'll need to update kvc to a new tag to make this happen. I'm taking a look into it.

angrycub:
@zeeshanlakhani Per our discussion, kvc 1.3.0 doesn't have a path/3 which provides a default. I'll leave it for you to determine if a switch to `kvc:value/3` would work in it's place or if you'd rather upgrade the kvc dependency.

zeeshanlakhani:
Closing, as superseded by https://github.com/basho/yokozuna/pull/448.

PR #448, opened 2015-01-28
test to check on missing maxScore value from Solr #448 [JIRA: RIAK-1498]

zeeshanlakhani:
Addes test to show https://github.com/basho/yokozuna/pull/447#issue-55647850 is not occurring w/ erlang pb client. Also updates kvc lib in case we need defaults. So, all in all, a minor update.

### DIFF
--- a/riak_test/yz_pb.erl
+++ b/riak_test/yz_pb.erl
@@ -23,6 +23,7 @@
    <field name=\"_yz_rk\" type=\"_yz_str\" indexed=\"true\" stored=\"true\" multiValued=\"false\"/>
    <field name=\"_yz_rb\" type=\"_yz_str\" indexed=\"true\" stored=\"true\" multiValued=\"false\"/>
    <field name=\"_yz_err\" type=\"_yz_str\" indexed=\"true\" stored=\"true\" multiValued=\"false\"/>
+   <field name=\"age_i\" type=\"int\" indexed=\"true\" stored=\"true\" multiValued=\"false\"/>
 </fields>
 <uniqueKey>_yz_id</uniqueKey>
 <types>
@@ -43,6 +44,7 @@ confirm() ->
     confirm_multivalued_field(Cluster),
     confirm_stored_fields(Cluster),
     confirm_search_non_existent_index(Cluster),
+    confirm_search_to_test_max_score_defaults(Cluster),
     pass.
 
 select_random(List) ->
@@ -105,6 +107,8 @@ store_and_search(Cluster, Bucket, Key, Body, CT, Search, Params) ->
         lager:info("Search for ~s [~p:~p]", [Search, Host, Port]),
         {ok,{search_results,R,Score,Found}} =
             riakc_pb_socket:search(Pid, BType, Search, Params),
+            ?assertNotEqual(Score, []),
+            ?assertNotEqual(Score, 0),
         case Found of
             1 ->
                 [{BType,Results}] = R,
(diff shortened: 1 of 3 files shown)


cmeiklejohn:
I will take this review.

zeeshanlakhani:
@cmeiklejohn, cool. I've set this back to develop on riak_kv, so things should work as they should when the other develop branches are updated.

cmeiklejohn:
Is there a test for this?

zeeshanlakhani:
This was a real simple edit based on the right types, but I'll add an eunit... and also wait for the update to the develop branch.

cmeiklejohn:
The test will ensure no one removes it in the future.

zeeshanlakhani:
@cmeiklejohn agreed. Will do.

zeeshanlakhani:
Test proves that the erlang client (pb) works w/o the defaults. Adding the test and updating kvc for things like defaults if we need it. @angrycub will keep the current yz_patch and take a further look on what's going on w/ the difference between the java client (where max_score can be an empty list) and erlang client.

riak_test/yz_pb.erl
```diff
@@ -244,3 +247,28 @@ confirm_stored_fields(Cluster) ->
...
+    Body = <<"{\"age_i\":5}">>,
+    Params = [{sort, <<"age_i asc">>}],
+    HP = select_random(host_entries(rt:connection_info(Cluster))),
+    URL = bucket_url(HP, Bucket, "multivalued"),
+    lager:info("Storing to bucket ~s", [URL]),
+    {Host, Port} = HP,
+    %% populate a value
+    {ok, "204", _, _} = ibrowse:send_req(URL, [?CT_JSON], put, Body),
+    %% Sleep for soft commit
+    timer:sleep(1100),
```
cmeiklejohn: Anything better you can do here?  Seems like it could trigger a race.

cmeiklejohn:
Something seems weird here; I'm only seeing the addition of the test and not the code?

riak_test/yz_pb.erl
```diff
@@ -244,3 +247,28 @@ confirm_stored_fields(Cluster) ->
...
+    Search = <<"age_i:5">>,
+    {ok, Pid} = riakc_pb_socket:start_link(Host, (Port-1)),
+    {ok,{search_results,[{Index,Fields}],Score,_Found}} =
+        riakc_pb_socket:search(Pid, Index, Search, Params),
+    ?assertNotEqual(Score, []),
+    ?assertNotEqual(Score, 0),
```
cmeiklejohn: Why not assert what the score should be?  Do we know?

Basho-JIRA:
Updated Note: https://github.com/basho/yokozuna/pull/448#issuecomment-73633332 

riak_test/yz_pb.erl
zeeshanlakhani: I was just keeping to the rest of the set of tests, currently. The initial patch was not due to a race cond.

riak_test/yz_pb.erl
zeeshanlakhani: We can't determine the exact score from solr, but I'll push a line of cleanup right now actually :).

zeeshanlakhani:
I've rebased things to focus the commits on purely adding the test in response to the issue brought up in #447.

zeeshanlakhani:
@cmeiklejohn sanity of commits now preserved, timeout removed.

zeeshanlakhani:
Just needs 2.0 -> develop riak_kv.

zeeshanlakhani:
@borshop: retry 431ff5ec97430a09220924a689993876869b6bb7

cmeiklejohn:
That branch needs to be rebased and squashed before we consider merging.

zeeshanlakhani:
@cmeiklejohn rebased. just need the fix in the mecks for green.

cmeiklejohn:
@borshop: retry

cmeiklejohn:
Bors is confused.  Open a new PR with the newly rebased commit?

zeeshanlakhani:
@cmeiklejohn bors won't retry on pr comments. It will only retry on a commit that has been initially +1'ed. But, I'll open a new PR to make things easier.

PR #452, opened 2015-02-14
Perf/zl/remove unnecessary delete to solr idx when crdt or sc

zeeshanlakhani:
Fixes https://bashoeng.atlassian.net/browse/RIAK-1504. 

Improves performance by avoiding an extra delete operation to remove siblings through Solr on the put path when riak datatypes or strong consistency is used. Initial patch/issue was discovered by @drewkerrigan, https://github.com/basho-labs/yokozuna_perf_patch.

### DIFF
--- a/src/yz_kv.erl
+++ b/src/yz_kv.erl
@@ -468,30 +468,30 @@ is_service_up(Service, Node) ->
 %%
 %% @doc Check if object has 2.0 CRDT datatype entry or property for
 %%      strong consistency.
--spec is_datatype_or_consistent(obj()) -> boolean().
+-spec is_datatype_or_consistent(obj()) -> boolean()|{error, _}.
 is_datatype_or_consistent(Obj) ->
-    case riak_kv_crdt:is_crdt(Obj) of
-        false ->
-            is_strongly_consistent(Obj);
-        true ->
-            true;
-        _ ->
-            false
+    Bucket = riak_object:bucket(Obj),
+    case riak_core_bucket:get_bucket(Bucket) of
+        BProps when is_list(BProps) ->
+            case is_datatype(BProps) of
+                false ->
+                    %% Check if this is a consistent object
+                    lists:member({consistent, true}, BProps);
+                _ ->
+                    true
+            end;
+        {error, _}=Err ->
+            Err
     end.
 
 %% @private
 %%
-%% @doc Check if object has property for strong consistency.
--spec is_strongly_consistent(riak_object:riak_object()) -> boolean().
-is_strongly_consistent(Obj) ->
-    Bucket = riak_object:bucket(Obj),
-    case riak_kv_util:consistent_object(Bucket) of
-        true ->
-            true;
-        _ ->
-            false
-    end.
-
+%% @doc Check if BucketProps has 2.0 CRDT datatype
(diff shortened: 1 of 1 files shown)


Basho-JIRA:
https://github.com/basho/yokozuna/pull/452 

src/yz_kv.erl
```diff
@@ -460,3 +463,131 @@ is_owner_or_future_owner(P, Node, Ring) ->
...
 is_service_up(Service, Node) ->
     lists:member(Service, riak_core_node_watcher:services(Node)).
+
+%% @private
+%%
+%% @doc Check if object has 2.0 CRDT datatype entry or property for
+%%      strong consistency.
+-spec is_datatype_or_consistent(obj()) -> boolean().
+is_datatype_or_consistent(Obj) ->
+    case riak_kv_crdt:value(Obj) of
```
seancribbs: Similar to the `is_strongly_consistent/1` below, you could check the bucket properties only to determine if it's a CRDT. Calling `riak_kv_crdt:value/1` will perform a deserialization of the value, which is extra work if you are not intending to use the result.

zeeshanlakhani: Makes sense. Will do. 

Zeeshan Lakhani
programmer | 
software engineer at @basho | 
org. member/founder of @papers_we_love |
twitter => @zeeshanlakhani

zeeshanlakhani:
@borshop will now fail until https://github.com/basho/riak_kv/pull/1081 is reviewed/merged, but eunit tests pass locally.

src/yz_kv.erl
```diff
@@ -460,3 +463,139 @@ is_owner_or_future_owner(P, Node, Ring) ->
...
+            false
+    end.
+
+
+%% @private
+%%
+%% @doc Set yz_solr:index delete operation(s) on write_reason.
+-spec delete_operation(obj(), put|handoff|anti_entropy, [doc()], bkey(), lp()) ->
+                              []|[{id, _}]|[{siblings, _}].
+delete_operation(Obj, put, Docs, BKey, LP) ->
```
seancribbs, line 501: Can you explain why `handoff` or `anti_entropy` should result in cleanups? Shouldn't the same reasons apply in those cases as for `put`?

zeeshanlakhani, line 501: In a conversation I had w/ @drewkerrigan, it was to keep the delete operation for  `handoff`/`anti_entropy` as cleanup b/c there was an edge case where data-typed keys could actually have a siblings and "one of the test cases the client had was like deleting without using a context and then updating the datatype object."

seancribbs, line 501: Understood. I think we should fix the underlying problem in that case, but that should be ok for now.

zeeshanlakhani, line 501: Agreed, as there's another issue altogether.

seancribbs:
Holding off final review on this for benchmarks.

zeeshanlakhani:
@seancribbs, yep, will post results here comparing `develop` to this branch.

zeeshanlakhani:
@seancribbs will reopen w/ a new PR... as the rebase w/ the changes to dev got a bit funky.

Basho-JIRA:
Updated fix version to 2.0.6 since the GH was included in release notes by Zeeshan.   

**_[posted via JIRA by Patricia Brewer]**_

Issue #402, opened 2014-06-11
Creating a new search index via HTTP responds before the index is available [JIRA: RIAK-1547]


superstructor:
When scripting the creation of schemas, indexes and associated bucket types I noticed that there is a significant delay between the HTTP response for creating an index and the index actually being made available for association with a bucket / bucket-type ?

For example if I run

`curl -v -XPUT http://localhost:8098/search/index/example -H'content-type:application/json' -d'{"schema":"example"}'`

I get a `204` response which means "The server has fulfilled the request..."

Then if I immediately run

`riak-admin bucket-type create example '{"props":{"search_index":"example","allow_mult":false}}'`

The riak-admin command fails because the index is not available yet, on my MacBook Pro (latest iteration, i7 2.6Ghz w/ 16GB RAM) I need to wait between 2 and 5 seconds after the HTTP response before the index is actually available, but is generally unpredictable and obviously depends on the hardware platform / load.

You need to either
1. Preferably block the response until the index is available.

OR
1. Correctly signal that the request is NOT fulfilled by using a `202 Accepted` status code which means "The request has been accepted for processing, but the processing has not been completed."

Great work on Yokozuna for Basho Riak 2.0, really exciting to be working with this in general!

I am using the official OS X Riak 2.0-beta1 package on OS X 10.9.3.

Thanks!

rzezeski:
@superstructor Yes, the index creation is not synchronous.  I also agree it should be made synchronous as time permits after the 2.0 release.

In the meantime you can use `GET /search/index/Foo` to determine if the index is up in Solr.  You need to call it on every node as the ping is only performed locally.

binarytemple:
Referenced in ZD issue - https://basho.zendesk.com/agent/tickets/10045

Basho-JIRA:
Relies on

riak_pb -> https://github.com/basho/riak_pb/pull/111
riak-erlang-client -> https://github.com/basho/riak-erlang-client/pull/207
yokozuna -> https://github.com/basho/yokozuna/pull/463
riak_test -> https://github.com/basho/riak_test/pull/745 

**_[posted via JIRA by Zeeshan Lakhani]**_

Issue #446, opened 2015-01-21
More Informative Logging Around Yokuzuna Indexing Errors [JIRA: RIAK-1491]


Basho-JIRA:
Related to the notes and short-term fixed mentioned in https://bashoeng.atlassian.net/browse/RIAK-1484. 

**[Created in JIRA by Zeeshan Lakhani]**

PR #453, opened 2015-02-17
Wait up to 20 seconds for yz_index:exists before returning from create/3

drewkerrigan:
This change is in response to #402 (RIAK-1547). Typical wait time for yz_index:exists is between 5-12 seconds. I'm not sure if we should define a new response for cases when it takes longer that 20 seconds to create the index because we can't really be sure if it is just going to take longer or if there was an actual issue with the creation.

### DIFF
--- a/src/yz_index.erl
+++ b/src/yz_index.erl
@@ -87,7 +87,8 @@ create(Name, SchemaName, NVal) when is_integer(NVal),
                     {error, schema_not_found};
                 true  ->
                     Info = make_info(SchemaName, NVal),
-                    ok = riak_core_metadata:put(?YZ_META_INDEXES, Name, Info)
+                    ok = riak_core_metadata:put(?YZ_META_INDEXES, Name, Info),
+                    wait_for_index(Name, 20)
             end;
         {error, _} = Err ->
             Err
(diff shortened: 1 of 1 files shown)


zeeshanlakhani:
@drewkerrigan I was looking at a fix for this too, but I don't feel like a random wait interval works here, and it is not configurable. @seancribbs and I have discussed this and plan to put a little time into a succinct solution, as it seems like sync/blocking-create should be initiated differently. The async one should maybe return that 202 as well if we allow both options, but it seems like being sync is preferable form a http/pb standpoint.

drewkerrigan:
@zeeshanlakhani I agree that the time for a wait interval should be configurable; also agree about returning 202 if we decide to preserve the current async capability, but yes - it seems that a sync/blocking create is the expected behavior from the client perspective anyway. With regard to a different approach, the tricky part about this operation is that the longest part of the operation happens entirely within Solr with little visibility from the Riak side. It would be nice to have a cleaner view of Solr's index creation but it also seems easy to over-engineer that problem. Either way I'm looking forward to getting a fix out for this sooner than later since it comes up a lot for clients.

zeeshanlakhani:
@drewkerrigan yeppers. After doing some thinking, I'm going to implement create as synchronous, keeping compatibility w/ all the apis.

Once I get back onto #130 (RIAK-1173), I'll create specific funs for two-phase commit.

Agree w/ your point on more visibility on the Solr side. I'm hoping the work in Errbal (affiliated w/ #446 (RIAK-1491)) will fit nicely in Yokozuna, and we'll have something better than what we have now. :+1: otherwise.

zeeshanlakhani:
Closing... as the new PR will be up this week.

PR #458, opened 2015-02-24
Bugfix/im/handle get ring timeouts

ian-mi:
Fixes issue #437 (RIAK-1567) by adding a catch-all clause to yz_events:handle_info. Also adds logging to yz_events:remove_non_owned_data so that failed calls to yz_cover:get_ring_used can be detected.

### DIFF
--- a/src/yz_events.erl
+++ b/src/yz_events.erl
@@ -180,6 +180,7 @@ remove_non_owned_data(unknown) ->
     %% The ring used to calculate the current coverage plan could not
     %% be determined. In this case do nothing to prevent removing data
     %% that the current coverage plan is using.
+    ?DEBUG("the current ring is unknown, no data can be removed", []),
     ok;
 remove_non_owned_data(Ring) ->
     case yz_solr:cores() of
--- a/src/yz_cover.erl
+++ b/src/yz_cover.erl
@@ -45,14 +45,13 @@
 %% to the caller how to interpret this.
 -spec get_ring_used() -> ring() | unknown.
 get_ring_used() ->
-    try gen_server:call(?MODULE, get_ring_used, 5000) of
-        undefined -> unknown;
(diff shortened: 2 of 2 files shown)


src/yz_events.erl
```diff
@@ -180,7 +183,7 @@ remove_non_owned_data(unknown) ->
     %% The ring used to calculate the current coverage plan could not
     %% be determined. In this case do nothing to prevent removing data
     %% that the current coverage plan is using.
-    ok;
+    ?DEBUG("the current ring is unknown, no data can be removed", []);
```
zeeshanlakhani: You're going to still want the `ok` after the `?DEBUG` statement.

zeeshanlakhani:
@ian-mi have you branched from the most recent version of develop, https://github.com/basho/yokozuna/blob/develop/rebar.config#L17, that contains the ibrowse fix for xref?

zeeshanlakhani:
@ian-mi also... can you write a simple test for this... maybe even just meck it out as a functional test?

ian-mi:
Changed the fix to use a proxy process to make the genserver call in get_ring_used rather than add an additional case to yz_events:handle_info.

bearcage:
:+1: a2e9ce7

zeeshanlakhani:
Great work @ian-mi!

ian-mi:
I was able to trigger the get_ring_used timeout by adding a five second sleep to its yz_cover:handle_call function clause, and verify that these timeouts were causing the yz_events:handle_info crashes. I then applied https://github.com/basho/yokozuna/commit/a2e9ce70cc6d932f6dbba5689f4fa3be2c5bb2b9 and verified that the crashes no longer occurred.

zeeshanlakhani:
@borshop merge

Issue #450, opened 2015-02-10
AAE Failing hourly, with exit value {badarg,46}, [{base64,decode_binary[{base64.erl... [JIRA: RIAK-1557]


ghost:
We are getting these error messages hourly, I assume when trying to rebuild the AAE tables....

```
emulator Error in process <#.####.###> on node 'riak@ip.add.re.ss' with exit value: {{badarg,46},[{base64,decode_binary,2,[{file,"base64.erl"},{line,191}]},{yz_solr,to_pair,1,[{file,"src/yz_solr.erl"},{line,414}]}, {yz_solr,'-get_pairs/1-lc$^0/1-0-',1,[{file,"src/yzsolr.erl"},{line,411}]},{yz_solr,entropy_data,2,[{file...
```

Unfortunately erlang appears to cut off the error logs, so that's all I can provide currently...

Would appreciate any tips on how to proceed in terms of debugging the issue.

zeeshanlakhani:
Hey @boardom. You should be able to see the entire error in the crash.log file. That'd help us take a look at what's going on. If you need more info about Logs in Riak, check out http://docs.basho.com/riak/latest/community/faqs/logs/.

ghost:
Yup. I had already looked there... Only difference is that one prints out the ===ERROR REPORT==== and then the line.

Any other suggestions.

zeeshanlakhani:
@boardom have you tried deleting the `data_dir` directory where AAE trees are stored before adding data again -> http://docs.basho.com/riak/latest/ops/advanced/configs/configuration-files/#Active-Anti-Entropy?

ghost:
We actually pushed updated schema's, which appears to have resolved the isssue... Unfortunate as there's probably a data handling issue somewhere...  Thanks for the help. 

Feel free to close.

zeeshanlakhani:
@boardom yeah... surprised that there was not more descriptive data handling/corruption errors in any of the logs, even moreso, the solr.log if the validation issues occured on the Solr side, which would be my guess here. We have an issue out to better expose these solr errors in the logs, https://github.com/basho/yokozuna/issues/446. 

Thanks.

ghost:
I looked into it a bit a while back, and it stemmed back to something with
erlang and max buffer size on the error reporting/stack trace stuff... I
didn't follow up far enough obviously, but definately odd... Truncating
stack traces is just evil.

stevegaron:
[...]

Issue #436, opened 2014-11-22
AAE fails when Riak key contains spaces


oleksiyk:
The following error occurs during AAE when Yokozuna hits the key with a space (' ') in it:

```
emulator Error in process <0.26807.79> on node 'riak@10.0.1.1' with exit value: {{badmatch,false},[{base64,decode_binary,2,[{file,"base64.erl"},{line,211}]},{yz_solr,to_pair,1,[{file,"src/yz_solr.erl"},{line,414}]},{yz_solr,'-get_pairs/1-lc$^0/1-0-',1,[{file,"src/yz_solr.erl"},{line,411}]},{yz_solr,'-get_pairs/1-lc$^0/1-0-'...
```

As Ryan Zezeski have found in a mailing list discussion the problem is:

> You have found a bug in Yokozuna.
> 
> https://github.com/basho/yokozuna/blob/develop/src/yz_doc.erl#L230
> 
> https://github.com/basho/yokozuna/blob/develop/java_src/com/basho/yokozuna/handler/EntropyData.java#L139
> 
> It foolishly assumes there is no space character used in the type,
> bucket, or key names. As a workaround I think you'll have to make sure
> your application converts all spaces to some other character (like
> underscore) before storing in Riak.

oleksiyk:
I've build Riak from source with this patch: https://github.com/oleksiyk/yokozuna/commit/d0ec38da40ecfb5049945390907b5c46d6afb762

and AAE seem to work nicely now with keys with spaces.

p.s.
However I noticed that I get inconsistent number of results (num_found) for the same query when its repeated. For example the query may return 10 docs, then 7, then 5, then 10 again. And the docs that are missing in the result are always those repaired by AAE and never those that were indexed normal way (by inserting or modifying object in Riak).

oleksiyk:
Can someone please comment if the problem with inconsistent num_found is because my attempt to fix the spaces is wrong or is it some different kind of problem? The AAE has finished and `riak-admin search aae-status` shows that all trees are built and synced but I still get different number of results when repeating the same query.

PR #459, opened 2015-02-26
fix to java w/ version update and update to aae/pb tests for RIAK-1557 i...

zeeshanlakhani:
...nvolving entropy data and spaces in types/buckets/keys.

This fixes up the issue w/ Entropy Data described in https://github.com/basho/yokozuna/issues/450 and https://github.com/basho/yokozuna/issues/436. 
- [x] update version #'s in yz jars within https://github.com/basho/yokozuna/blob/develop/tools/build-jar.sh and https://github.com/basho/yokozuna/blob/develop/tools/grab-solr.sh and upload to s3.

### DIFF
--- a/misc/bench/src/yz_driver.erl
+++ b/misc/bench/src/yz_driver.erl
@@ -9,7 +9,7 @@
 -include_lib("basho_bench/include/basho_bench.hrl").
 -record(state, {default_field, fruits, pb_conns, index, bucket, iurls, surls}).
 -define(DONT_VERIFY, dont_verify).
--define(SPACER, 'foo ').
+-define(SPACER, "testfor spaces ").
 
 -define(M100,   100000000).
 -define(M10,    10000000).
--- a/misc/bench/src/yz_driver.erl
+++ b/misc/bench/src/yz_driver.erl
@@ -9,6 +9,7 @@
 -include_lib("basho_bench/include/basho_bench.hrl").
 -record(state, {default_field, fruits, pb_conns, index, bucket, iurls, surls}).
 -define(DONT_VERIFY, dont_verify).
+-define(SPACER, 'foo ').
 
 -define(M100,   100000000).
 -define(M10,    10000000).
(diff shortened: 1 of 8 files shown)


zeeshanlakhani:
@seancribbs I uploaded yokozuna-2.jar and yokozuna-2.jar.sha to S3.

misc/bench/src/yz_driver.erl
```diff
@@ -9,6 +9,7 @@
 -include_lib("basho_bench/include/basho_bench.hrl").
 -record(state, {default_field, fruits, pb_conns, index, bucket, iurls, surls}).
 -define(DONT_VERIFY, dont_verify).
+-define(SPACER, 'foo ').
```
bearcage: Total nitpick, but mind switching the prefix on space testing keys to something that makes the purpose clear if you're tracking down a problem later?

misc/bench/src/yz_driver.erl
```diff
@@ -159,6 +160,17 @@ run(load_fruit, KeyValGen, _, S=#state{iurls=URLs}) ->
         {error, Reason} -> {error, Reason, S2}
     end;
 
+run(load_fruit_plus_spaces, KeyValGen, _, S=#state{iurls=URLs}) ->
+    Base = get_base(URLs),
+    {Key, Val} = KeyValGen(),
+    Key2 = mochiweb_util:quote_plus(lists:concat([?SPACER, Key])),
```
seancribbs, line 166: Does this even work? `?SPACER` is an atom.

zeeshanlakhani, line 166: It does... but should I just make it a string?

zeeshanlakhani, line 166:
``` erlang
 list_to_binary(lists:concat([' foo bar', binary_to_integer(<<"44">>)])).
```

would then return `<<" foo bar44">>`... which would get formatted correctly on the url calls.

zeeshanlakhani, line 166: string works too... so whichever is preferred :D.

seancribbs, line 166: I think a string would be less surprising, that's all.

zeeshanlakhani, line 166: :+1:

seancribbs:
:+1: ca768f3

```
Test Results:
aae_test-bitcask: pass
0 Tests Failed
1 Tests Passed
```

zeeshanlakhani:
@borshop merge

PR #462, opened 2015-03-07
Update grab-solr.sh [JIRA: RIAK-1587]

nbari:
sh instead  bash in order to be available to compile riak in *nix with no bash


zeeshanlakhani:
I'll double check this w/ a build of riak on nix... but cool stuff.

seancribbs:
Sorry for the confusion here, looks like this is covered by #467

PR #467, opened 2015-03-10
Update grab-solr.sh

nbari:
sh instead of bash and added fetch


zeeshanlakhani:
Closing in favor of https://github.com/basho/yokozuna/pull/475, which builds solr & jars correctly. @nbari, please take a look when you can.

PR #472, opened 2015-03-13
add ability to choose upgrade version for riak_search->yokozuna migratio...

zeeshanlakhani:
...n and get it from giddyup metadata

We're changing our builders to make the `previous` version 2.0.5... but this migration test relies on the 1.4 series, which is the last to use the original riak_search. With the upgrade_version, we can now specify legacy for this test in the test metadata.


seancribbs:
:+1: da36470

zeeshanlakhani:
@borshop merge

Basho-JIRA:
merged into 2.0 branch 

**_[posted via JIRA by Patricia Brewer]**_

Issue #481, opened 2015-04-08
Yokozuna loses entries when YZ AAE trees expire, dealing w/ Default Bucket Types [JIRA: RIAK-1674]


kesslerm:
When AAE trees expire, Yokozuna starts to lose entries. After initially reporting identical numbers, once the YZ AAE trees have been expired the total number of entries as reported by "http://$RIAK_HOST/solr/$INDEX_NAME/select?q=_:_" is less than the total number of keys reported by key listing.

Steps to reproduce:
- Set up a 5 node devrel cluster of Riak-2.0.5; add `search = on` and 
  `anti_entropy.tree.build_limit.per_timespan = 5m` to `riak.conf` on each node, start the nodes and join them into a cluster.
- Add the default yz schema to the test bucket

```
#!/bin/bash
RIAK_HOST="http://127.0.0.1:10018"

test_results_bucket_props=`curl -s "$RIAK_HOST/buckets/Test_Results/props"`
if [[ $test_results_bucket_props  =~ "index" ]]
then
  echo "Index already exists";
  exit
fi

echo "Creating index"
curl -XPUT "$RIAK_HOST/search/index/Test_Results"
sleep 10

echo "Adding index to bucket"
curl -XPUT -H "Content-Type: application/json" "$RIAK_HOST/buckets/Test_Results/props" -d '{"props":{"search_index":"Test_Results"}}'

test_results_bucket_props=`curl -s "$RIAK_HOST/buckets/Test_Results/props"`
if [[ $test_results_bucket_props  =~ "index" ]]
then
  echo "Index added";
fi
```
- Add some test data

```
#!/bin/bash

for i in {0..4999}
do
        uuid=$(uuidgen)
        echo ${uuid}

        curl -XPUT http://127.0.0.1:10018/buckets/Test_Results/keys/${uuid} \
                -H "Content-Type: application/json" \
            -d "{\"uuid\": \"${uuid}\", \"date\": \"$(date "+%FT%T.00000:Z")\"}"
done
```
- Monitor the number of hits reported by Riak and Yokozuna

```
#!/bin/bash
echo "solr:"
curl -XGET "http://127.0.0.1:10018/solr/Test_Results/select?wt=json&q=*:*" 2>/dev/null | json_pp | grep numFound
echo "keys in bucket:"
echo $((`curl -XGET "http://127.0.0.1:10018/buckets/Test_Results/keys?keys=true" 2>/dev/null | json_pp | wc -l` - 4))
```
[...]

PR #482, opened 2015-04-08
update default_schema to return sets in query responses by storing them...

zeeshanlakhani:
Handles/fixes RIAK-1672 (#480) issue with a missing set in the query response... focused in on the _default_schema_ only. 

In regards to missing data, that's due to something awry in the replication, as YZ queries only use R=1, and one of the partitions is missing the data. That's either caused by errors or will be fixed as part of the AAE expiration testing, #481 (RIAK-1674). 

Goes along w/ riak_test pr https://github.com/basho/riak_test/pull/773.

### DIFF
--- a/priv/default_schema.xml
+++ b/priv/default_schema.xml
@@ -95,13 +95,13 @@
 
    <!-- Riak datatypes default fields-->
    <field name="counter" type="int"    indexed="true" stored="true" multiValued="false" />
-   <field name="set"     type="string" indexed="true" stored="false" multiValued="true" />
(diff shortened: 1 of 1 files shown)


zeeshanlakhani:
@seancribbs figure that you take a look at this b/c of https://github.com/basho/yokozuna/commit/43e4a53d4ff7b4ef28bfd945d040784c394ab51d... where sets were set to `stored='false'`. Custom schemas could always fix this too.

fadushin:
+1 dd4a677

fadushin:
@borshop merge

Issue #382, opened 2014-05-14
use mochiglobal for the extractor map (and also, CMD)... check via CAPs [JIRA: RIAK-1673]


rzezeski:
Use mochiglobal for the extractor map instead of the ring.  The
extractor map is something that rarely, if ever, changes.  It should
be stored in something optimized for reads like mochiglobal rather
than pulling the entire ring each time.

PR #483, opened 2015-04-12
move to using mochiglobal to store extractor map

zeeshanlakhani:
- have read-through if map not in mochiglobal then check CMD and then the Ring
- add capability to now move extractor map into CMD (and no longer store it in the ring metadata)

Fixes RIAK-1673 (#382).


fadushin:
Note that the riak_test for this change is in https://github.com/basho/riak_test/pull/779

fadushin:
+1 9203f45

PR #487, opened 2015-04-15
Bugfix/stop index creation loop on bad data related to Zendesk issue.

zeeshanlakhani:


src/yz_index.erl
```diff
@@ -220,33 +212,81 @@ local_create(Name) ->
...
+            lager:info("Unloaded previous instance of index ~s", [Name]);
+        {error, {ok, "400", _, Resp}} ->
+            lager:info("Couldn't unload index ~s prior to creating "
+                       "a new instance of it. This is likely the first "
+                       "time this index has been created. Solr "
+                       "responded with ~s.", [Name, Resp]);
+        {error, UnloadErr} ->
+            lager:error("Couldn't unload index ~s prior to creating "
+                        "a new instance of it: ~p", [Name, UnloadErr])
+    end,
```
seancribbs: I think all clauses here return `ok`, you should be fine without the extra return value.

zeeshanlakhani: right... as lager will do so, cool.

src/yz_index.erl
```diff
@@ -220,33 +212,81 @@ local_create(Name) ->
...
+    local_remove(Name, CoreProps).
+
+-spec local_remove(index_name(), [{core | delete_instance | delete_index | delete_data_dir,
+                                   index_name() | string()}]) -> ok.
+local_remove(Name, CoreProps) ->
+    case yz_solr:core(remove, CoreProps) of
+        {ok, _, _} ->
+            lager:info("Unloaded previous instance of index ~s", [Name]);
+        {error, {ok, "400", _, Resp}} ->
+            lager:info("Couldn't unload index ~s prior to creating "
```
seancribbs, line 282: Might be better as `notice` level.

src/yz_index.erl
```diff
@@ -220,33 +212,81 @@ local_create(Name) ->
...
+    %% impact should be minimal (and somewhat unavoidable; if AAE is
+    %% replacing the whole index, that's what it's going to do anyway).
+    ok = local_remove(Name, [{core, Name}]),
+    case yz_solr:core(create, CoreProps) of
+        {ok, _, _} ->
+            lager:info("Created index ~s with schema ~s",
+                       [Name, SchemaName]),
+            ok;
+        {error, exists} ->
+            lager:info("Index ~s already exists in Solr, "
```
seancribbs: `notice` seems appropriate here as well.

src/yz_index.erl
zeeshanlakhani, line 282: hmm, @seancribbs, the 400 is more _expected_ behavior, as that's the first time you're creating an index, which the most typical and correct use-case for this. So, wouldn't notice refer to something a bit more _significant_ (not an error, I know) in terms of log severity?

seancribbs, line 282: Hmm. I'd love to hear @angrycub and @mrallen1 weigh in on this. It could go either way in my mind.

zeeshanlakhani, line 282: agreed. I'll take @angrycub and @mrallen1's votes and adjust accordingly :).

jadeallenx, line 282: If 400 is expected my take is that its something that should be at a warning level. My thinking around info vs. notice has to do with info -> logging mostly for operational tools or state updates and notice -> human this is something you should take action to deal with

src/yz_index.erl
zeeshanlakhani: Agree here.

src/yz_index.erl
jadeallenx, line 282: Well, after another quick chat with @zeeshanlakhani I think the right level is info - this is something that's expected and done automatically, so we should just inform the user that it was done.

fadushin:
+1 59106f7207c1ba61caf4937afe30780ba211423d

zeeshanlakhani:
@borshop merge

PR #502, opened 2015-06-12
* config and build script updates for pulling in solr 4.10.4 (solr bi…

zeeshanlakhani:
This will probably be only used for patches and not for a 2.0.x release, but here anyway for discussion, etc...
- config and build script updates for pulling in solr 4.10.4 (solr binary already pushed to uploads.basho)
- minor updates to prevent benign but useless warnings in solr logs
- patch update for move to 4.10.4
- move to jdk 7 for building solr 4.10.4, as it's the min version required

(cherry picked from commit 9745d4785479051cd8731aa923de5f64c8f58b8b)


zeeshanlakhani:
https://github.com/basho/riak_test/pull/799 and the other r_t tests that work for 2.0.5 are passing. If you're  running r_t, make sure to change the riakc/http-client versions to match correctly.

zeeshanlakhani:
Closing for now, but this shows can things can work off the 2.0 branch if we need a patch or something else.

PR #506, opened 2015-06-24
Cherry picked logging of deletion of core properties file from hotfix…

fadushin:
… branch.  Also toned down logging in yz_events when there is nothing to be done (once/minute, which is choking logs).

This PR addresses some improvements to login when Yokozuna decides to delete the core.properties file for a given Solr core (index).  It also tones down the logging when yokozuna tries to synchronize its view of the cores that are loaded in solr (via cluster metadata) and what it can scrape off the file system.  The current logging spits out an info message every minute, when no work results in the comparison.  This change makes it so that something is logged only if there is something that needs to be done.

### DIFF
--- a/src/yz_events.erl
+++ b/src/yz_events.erl
@@ -212,9 +212,12 @@ sync_indexes() ->
                                  yz_index:get_indexes_from_meta()),
             {Removed, Added, Same} = yz_misc:delta(IndexSetFromSolr,
                                                    IndexSetFromMeta),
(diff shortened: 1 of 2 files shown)


zeeshanlakhani:
:+1:  0d728a1

zeeshanlakhani:
:+1: 0d728a146143820cad15388797f1b7c86be7ed8e

zeeshanlakhani:
Since https://github.com/basho/yokozuna/pull/507 is merging. This should be caught-up to develop and then just diff the lager case on yz_events.

Issue #525, opened 2015-07-17
Yokozuna fails to re-index the remainder of a partition after a non-indexable document is encountered [JIRA: RIAK-2006]


kesslerm:
When a non-indexable document is encountered during AAE, the yz_exchange_fsm process dies with

```
2015-07-16 09:24:12 =ERROR REPORT====
Error in process <0.6990.0> on node 'dev2@127.0.0.1' with exit value: {{nocatch,{"Failed to index docs",{ok,"400",[{"Content-Type","application/json; charset=UTF-8"},{"Transfer-Encoding","chunked"}],<<106 bytes>>}}},[{yz_solr,index,3,[{file,"src/yz_solr.erl"},{line,199}]},{yz_kv,index,7,[{file,"src/yz_kv.erl"},{line...

2015-07-16 09:24:12.620 [error] emulator Error in process <0.6990.0> on node 'dev2@127.0.0.1' with exit value: {{nocatch,{"Failed to index docs",{ok,"400",[{"Content-Type","application/json; charset=UTF-8"},{"Transfer-Encoding","chunked"}],<<106 bytes>>}}},[{yz_solr,index,3,[{file,"src/yz_solr.erl"},{line,199}]},{yz_kv,index,7,[{file,"src/yz_kv.erl"},{line...
```

At this point the AAE tree build never completes, and consequently no other entries are being re-indexed for the given partition.

Reproduce:
- Set up a bucket with an index.
- Populate the bucket with test-data.
- Verify that Yokozuna returns the correct number of index entries.
- Add one entry that can't be indexed by Solr, e.g. because it has a non-parseable date field. Yokozuna should have logged an error and still return the previous number of entries.
- Stop the cluster, and manually remove the yz/<index_name> directories and the yz_anti_entropy directories from all nodes.
- Restart the cluster and wait for re-indexing to finish. Yokozuna will not be able to reach the full number of index entries again, and return an incoherent set of results. The actual number of missing entries depends on how many entries could be indexed from the affected partition, but will usually be less than the total.

zeeshanlakhani:
@kesslerm never deleting the problematic entry and letting AAE retrigger right?

kesslerm:
@zeeshanlakhani, yes, as long as the problematic entry stays in KV, the issue re-occurs any time AAE runs again.

binarytemple:
Customer [example of issue](https://basho.zendesk.com/agent/tickets/11371), and waiting on patch.

[...]

PR #526, opened 2015-07-17
yz_exchange_fsm: catch exception from yz_kv:index/7, yz_solr:index/3

kesslerm:
Without catching an exception thrown when Solr can't index an entry,
the yz_exchange_fsm process dies for a given partition, which will
never be re-indexed fully.

Fixes https://github.com/basho/yokozuna/issues/525

### DIFF
--- a/src/yz_exchange_fsm.erl
+++ b/src/yz_exchange_fsm.erl
@@ -223,16 +223,25 @@ repair(Partition, {_Reason, KeyBin}) ->
     %% node is owner.
     case yz_kv:local_get(Partition, BKey) of
         {ok, Obj} ->
-            case yz_kv:should_index(Index) of
-                true ->
-                    Ring = yz_misc:get_ring(transformed),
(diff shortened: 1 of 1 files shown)


zeeshanlakhani:
@kesslerm I'm back on Monday. A few things...

1) I'd like to clean this up a bit and maybe impl. the fix w/o a try/catch
2) I'd like to update the aae test here to showcase the fix
3) based on new release work, after 1 + 2, the PR should be opened against 2.0 and forward merged through. We'll probably want this (w/ my changes) in 2.1.2 @JeetKunDoug. 

@kesslerm ok if I close this on Monday and reopen a PR w/ a test targeting 2.0 to fix #525 (RIAK-1376) (RIAK-2006).

kesslerm:
@zeeshanlakhani sure, this was just my proof of concept code, to test how yz_exchange_fsm behaves if it's not crashing. This code solves the immediate problem, but feel free to build some better fix on top of the basic idea.

zeeshanlakhani:
@kesslerm thank you. will do. and thanks for finding this.

PR #527, opened 2015-07-20
Yokozuna fails to re-index the remainder of a partition after a non-indexable document is encountered

zeeshanlakhani:
- update yz_exchange_fsm to catch index error w/ ill-formated/failed document, thank you @kesslerm 
- update aae_test to test for this bug, 
- add a date field to check-up on in fruit_schema.xml

Fixes RIAK-2006 (#525) / https://github.com/basho/yokozuna/issues/525.

Will need to be merged-forward to 2.1 and develop.

### DIFF
--- a/misc/bench/schemas/fruit_schema.xml
+++ b/misc/bench/schemas/fruit_schema.xml
@@ -6,6 +6,8 @@
 
    <field name="text" type="text_ws" indexed="true" stored="false" multiValued="true"/>
 
+   <field name="date_register" type="tdate" indexed="true" stored="true" multiValued="false" />
+
    <field name="_version_" type="long" indexed="true" stored="true"/>
 
    <!-- Entropy Data: Data related to anti-entropy -->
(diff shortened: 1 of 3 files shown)


zeeshanlakhani:
@kesslerm very similar to your set of changes... but I added the test and defined it moreso in scope. 

@kesslerm and @fadushin please review and then merge forwarding will commence.

zeeshanlakhani:
@kesslerm i was able to get the failure (forever-waiting for an AAE exchange to occur) just w/ expire and not needing to stop/remove, which makes sense to me.

kesslerm:
@zeeshanlakhani yes expiring trees will have the same effect of AAE exchanges never happening. The code change overcomes the crash, which is the underlying cause, but it still doesn't mark those entries as non indexable. So each time AAE tries to fix the broken entry again, it will be flagged in the logs. At least the rest of the index will be rebuilt fine now.

kesslerm:
:+1: - repeated log entries about non-indexable KV entries are probably a good thing and alert users to the state of their data.

zeeshanlakhani:
@kesslerm right... i was going to mention that. They should be aware, consistently, that this datum is not handled properly. Compared to the previous PR however, I don't have to reshow the trace.

kesslerm:
Cool. Not showing the trace is a nice change. I just tested your version against my test database and it works fine.

jadeallenx:
For whatever its worth (maybe 2 cents)
+1 d09a6ff

kesslerm:
@zeeshanlakhani it would be good to merge this into the relevant branches before the next release cycle. The bug potentially affects many customers. We may also need patched beams for some.

zeeshanlakhani:
@kesslerm yep. that's what we targeted 2.0 first. Then we'll move this forward to 2.1 and develop. It will be def. be in 2.1.2 and eventually 2.0.7.

zeeshanlakhani:
@borshop merge

JeetKunDoug:
:+1: d09a6ff

zeeshanlakhani:
@borshop merge

PR #529, opened 2015-07-27
Bugfix/fd/riak 1937

fadushin:
This PR changes the signature of the yz_kv:index function to allow clients to pass a Bucket and Key, and to support the case in which the client of this function supplies an encoded Riak Object, as is the case in the write_once put path.  If (but only if) yokozuna is enabled and an encoded Riak Object is passed in, the binary is decoded into a Riak Object and passed into the (normal) index path.  Otherwise, the decoding is skipped, so as not to impose a performance penalty in the write_once path if yokozuna is not enabled.

This PR requires a corresponding fix to riak_kv, which is on the same branch name as this PR.

The yz_pb test has been augmented to include a test for write_once buckets.  The test has been shown to fail without the changes to riak_kv and yokozuna.

Cf. https://github.com/basho/riak_kv/pull/1159

### DIFF
--- a/riak_test/yz_pb.erl
+++ b/riak_test/yz_pb.erl
@@ -43,6 +43,7 @@ confirm() ->
     confirm_admin_index(Cluster),
     confirm_admin_bad_index_name(Cluster),
     confirm_basic_search(Cluster),
+    confirm_w1c_search(Cluster),
     confirm_encoded_search(Cluster),
     confirm_search_to_test_max_score_defaults(Cluster),
     confirm_multivalued_field(Cluster),
@@ -82,11 +83,12 @@ create_index(Cluster, BucketType, Index) ->
 
 create_index(Cluster, BucketType, Index, UseDefaultSchema) when
       is_boolean(UseDefaultSchema) ->
-    create_index(Cluster, BucketType, Index, 3, UseDefaultSchema);
+    create_index(Cluster, BucketType, Index, [{use_default_schema, UseDefaultSchema}]);
 create_index(Cluster, BucketType, Index, Nval) when is_integer(Nval) ->
-    create_index(Cluster, BucketType, Index, Nval, true).
-
-create_index(Cluster, BucketType, Index, Nval, UseDefaultSchema) ->
+    create_index(Cluster, BucketType, Index, [{n_val, Nval}]);
+create_index(Cluster, BucketType, Index, Props) when is_list(Props) ->
+    Nval = proplists:get_value(n_val, Props, 3),
+    UseDefaultSchema = proplists:get_value(use_default_schema, Props, true),
     Node = select_random(Cluster),
     [{Host, Port}] = host_entries(rt:connection_info([Node])),
     lager:info("create_index ~s for bucket type ~s [~p]", [Index, BucketType, {Host, Port}]),
@@ -111,10 +113,13 @@ create_index(Cluster, BucketType, Index, Nval, UseDefaultSchema) ->
     ?assertEqual([{index, Index}, {schema, SchemaName}, NvalT], IndexData),
 
     %% Add the index to the bucket props
-    yz_rt:set_bucket_type_index(Node, BucketType, Index, Nval),
+    yz_rt:set_bucket_type_index(Node, BucketType, Index, [{n_val, Nval} | Props]),
     yz_rt:wait_for_bucket_type(Cluster, BucketType),
     riakc_pb_socket:stop(Pid),
     ok.
+
+create_index(Cluster, BucketType, Index, Nval, UseDefaultSchema) ->
+    create_index(Cluster, BucketType, Index, [{n_val, Nval}, {use_default_schema, UseDefaultSchema}]).
 
 store_and_search(Cluster, Bucket, Key, Body, Search, Params) ->
     store_and_search(Cluster, Bucket, Key, Body, "text/plain", Search, Params).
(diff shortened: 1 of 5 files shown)


riak_test/yz_pb.erl
```diff
@@ -111,11 +113,15 @@ create_index(Cluster, BucketType, Index, Nval, UseDefaultSchema) ->
...
-    yz_rt:set_bucket_type_index(Node, BucketType, Index, Nval),
+    yz_rt:set_bucket_type_index(Node, BucketType, Index, [{n_val, Nval} | Props]),
     yz_rt:wait_for_bucket_type(Cluster, BucketType),
     riakc_pb_socket:stop(Pid),
     ok.
 
+create_index(Cluster, BucketType, Index, Nval, UseDefaultSchema) ->
+    create_index(Cluster, BucketType, Index, [{n_val, Nval}, {use_default_schema, UseDefaultSchema}]).
+
+
```
zeeshanlakhani: nothing important... but just a weird extra space here.

zeeshanlakhani:
@fadushin https://github.com/basho/yokozuna/blob/2.1/riak_test/aae_test.erl#L148 is failing for obvious reasons.

zeeshanlakhani:
@fadushin I'd squash the fix failing tests into one commit and the typo commit into the pb test commit.

src/yz_kv.erl
```diff
@@ -190,8 +190,20 @@ has_indexes(RemoteNode) ->
         _ -> false
     end.
 
-index(Obj, Reason, P) ->
-    case yokozuna:is_enabled(index) andalso ?YZ_ENABLED of
+-spec is_enabled() -> boolean().
+is_enabled() ->
+    yokozuna:is_enabled(index) andalso ?YZ_ENABLED.
+
+index(B, K, Obj, Reason, P) when is_binary(Obj) ->
```
zeeshanlakhani: maybe we can add a dialyzer spec to `index` now as well and a comment about allowing for Bucket and Key to pass through and why the latter version treats it as unnecessary.

zeeshanlakhani:
@fadushin I'd also like the results of a b_b run, if possible, comparing a set of writes against the previous version. Is that cool?

fadushin:
You mean for the w1c path, or just the normal path?  (Or both)  For write once, clearly it will be a LOT slower, with low throughput, as well.  But I suspect that is what is intended/desired.

zeeshanlakhani:
@fadushin for w1c. Just so we have it recorded. 

Also, all other tests pass, will re-run aae test shortly.

src/yz_kv.erl
```diff
@@ -190,8 +190,20 @@ has_indexes(RemoteNode) ->
...
+        true ->
+            RObj = riak_object:from_binary(B, K, Obj),
+            index(B, K, RObj, Reason, P);
+        _ ->
+            ok
+    end;
+index(_B, _K, Obj, Reason, P) ->
+    case is_enabled() of
         true ->
             Ring = yz_misc:get_ring(transformed),
```
zeeshanlakhani, line 214: After looking through the code in a walkthrough (esp. the pars the PR doesn't show, haha), @fadushin and I realized that the `should_index` check should be called before converting the object from binary. This led to a deeper look into some code duplication (around BKey creating, etc...) and whatnot that is happening; so, he's going to be refactoring his changes a bit and just adding a separate index/5 function for binary objects.

src/yz_kv.erl
```diff
@@ -190,6 +190,48 @@ has_indexes(RemoteNode) ->
         _ -> false
     end.
 
+
+%% @doc Index the data supplied in the Riak Object.
+%% The Riak Object should be a deserialized object (suitable for use with the riak_object
+%% module, or a binary, which has been serialized using riak_object:to_binary/1
+-spec index_binary(bucket(), key(), binary(), atom(), integer()) -> ok.
+index_binary(B, K, Obj, Reason, P) when is_binary(Obj) ->
+    BKey = {B, K},
```
zeeshanlakhani: Think you could do this in the try/catch scope.

src/yz_kv.erl
```diff
@@ -190,6 +190,48 @@ has_indexes(RemoteNode) ->
...
+    case yokozuna:is_enabled(index) andalso ?YZ_ENABLED of
+        true ->
+            Ring = yz_misc:get_ring(transformed),
+            case is_owner_or_future_owner(P, node(), Ring) of
+                true ->
+                    T1 = os:timestamp(),
+                    try
```
zeeshanlakhani, line 219: @fadushin thinking we can create a helper to avoid the duplication in index/3 and index/5, no?

src/yz_kv.erl
```diff
@@ -190,35 +190,54 @@ has_indexes(RemoteNode) ->
...
 
+
+%% @doc Index the data supplied in the Riak Object.
+%% The Riak Object should be a deserialized object (suitable for use with the riak_object
+%% module, or a binary, which has been serialized using riak_object:to_binary/1
+-spec index_binary(bucket(), key(), binary(), atom(), integer()) -> ok.
+index_binary(B, K, Obj, Reason, P) when is_binary(Obj) ->
+    index_internal(
+        Reason, P,
+        fun() -> {B, K} end,
```
zeeshanlakhani: I don't mind the passing of functions here, but this may work better as named helper functions (for readability) determined by the Obj guard (is_binary or not)

zeeshanlakhani:
@fadushin, in moving to having an index/3 and index/5, guessing you still plan make updates to the tests and the riak_kv PR; so, I'll leave those out of my comments.

zeeshanlakhani:
@fadushin Running another set of tests.

zeeshanlakhani:
:+1: eeb90d057781463606e378bda4c4373610d1fe2f

zeeshanlakhani:
Will merge this first, then basho/riak_kv#1159. @fadushin will still add b_b results here for preservation.

zeeshanlakhani:
@borshop merge

fadushin:
See https://github.com/basho/yokozuna/wiki/WriteOnceBucketIndexingBug for a discussion of this bug and some basho_bench results.

Issue #324, opened 2014-02-26
Function Clause Error in yz_entropy [JIRA: RIAK-2068]


rzezeski:
There seems to be a missing function head in `yz_entropy` when
iterating tree data. It looks like it can't deal with request
timeouts. It shouldn't be the end of the world if the tree builder
crashes because of this error as it will just be retried later but
might me good to retry the request or fail with a better message
(remember it's okay to fail in Erlang).

Here is a copy of the message from basho/riak_kv#854. The last message
is the issue. I do not believe any of the previous messages are
related.

```
2014-02-21 17:19:12.289 [error] <0.781.0> gen_server <0.781.0> terminated with reason: bad argument in call to erlang:byte_size({r_object,<<"IPDB">>,<<65,1,1,0>>,[{r_content,{dict,5,16,16,8,80,48,{[],[],[],[],[],[],[],[],[],...},...},...}],...}) in riak_kv_index_hashtree:maybe_throttle_build/4 line 460
2014-02-21 17:19:12.289 [error] <0.781.0> CRASH REPORT Process <0.781.0> with 0 neighbours exited with reason: bad argument in call to erlang:byte_size({r_object,<<"IPDB">>,<<65,1,1,0>>,[{r_content,{dict,5,16,16,8,80,48,{[],[],[],[],[],[],[],[],[],...},...},...}],...}) in riak_kv_index_hashtree:maybe_throttle_build/4 line 460 in gen_server:terminate/6 line 744
2014-02-21 17:19:12.289 [error] <0.773.0> Supervisor {<0.773.0>,poolboy_sup} had child riak_core_vnode_worker started with {riak_core_vnode_worker,start_link,undefined} at <0.781.0> exit with reason bad argument in call to erlang:byte_size({r_object,<<"IPDB">>,<<65,1,1,0>>,[{r_content,{dict,5,16,16,8,80,48,{[],[],[],[],[],[],[],[],[],...},...},...}],...}) in riak_kv_index_hashtree:maybe_throttle_build/4 line 460 in context child_terminated
[...]

PR #538, opened 2015-08-11
RIAK-2068 ++

zeeshanlakhani:
- Increase and handle ibrowse _timedout_ for `entropy_data` Solr request  https://github.com/basho/yokozuna/issues/324 to prevent crash during iteration or allowing the build to _pass_ if an error occurs in iterating over the tree or if solr was not pingable. 
- make `solr_request_timeout` configurable, which is used by all Solr localhost http request and add macro timeout to `entropy_data` call.
- clean-up tests and update tools.mk
- make yz_solr:ping use a HEAD request.

### DIFF
--- a/src/yz_index_hashtree.erl
+++ b/src/yz_index_hashtree.erl
@@ -295,7 +295,7 @@ load_built(#state{trees=Trees}) ->
         _ -> false
     end.
 
--spec fold_keys(p(), tree()) -> ok.
+-spec fold_keys(p(), tree()) -> [ok|timeout|not_available].
 fold_keys(Partition, Tree) ->
     LI = yz_cover:logical_index(yz_misc:get_ring(transformed)),
     LogicalPartition = yz_cover:logical_partition(LI, Partition),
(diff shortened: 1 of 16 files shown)


src/yz_index_hashtree.erl
```diff
@@ -307,8 +307,7 @@ fold_keys(Partition, Tree) ->
                 insert(async, IndexN, BKey, Hash, Tree, [if_missing])
         end,
     Filter = [{partition, LogicalPartition}],
-    [yz_entropy:iterate_entropy_data(I, Filter, F) || I <- Indexes],
```
zeeshanlakhani, line 310: So, fold_keys was crashing here when it couldn't get a response from the /entropy_data handler. Now, we iterate through but store an error atom if the latest iteration failed to give a response. Previously, if it crashed, we wouldn't get a `build_finished`... https://github.com/basho/yokozuna/pull/538/files#diff-9e0330457ea26c2e802f9524821758c6R558 ... call. Now, we throw a build_failed if the last iteration is an error (timeout) to requeue a poke in order to try and continue building the tree.

@bsparrow435 and I discussed the idea of only building a partial tree if errors/timeouts occur, but this could cause tons of exchanges if we're getting errors/timeouts that persist and occur early in the iteration cycle, as we'd _finish_ building the tree without many of the keys. Something to continue thinking about though. 

/cc @fadushin @jonmeredith

zeeshanlakhani:
Could use the eyes of @jonmeredith @bsparrow435 and @fadushin I think.

PR #541, opened 2015-08-25
Modified waits on queries to pattern match against 0 results

fadushin:
Branched off 2.0.  A subsequent PR will be a merge into 2.1.

This fixes an intermittently failing test where we have a bad match on the return value from a query, due to a race condition in indexing the data in Solr.  Also modified other test cases that use the same pattern, so that we match against the case where 0 results come back, and return false from the wait-for function, so that we will try again until a non-zero result comes back.

### DIFF
--- a/riak_test/yz_pb.erl
+++ b/riak_test/yz_pb.erl
@@ -240,7 +240,9 @@ confirm_multivalued_field(Cluster) ->
                     1 -> true;
                     0 -> false
                 end;
-            _ -> false
+            {ok, {search_results, [], _Score, 0}} ->
+                lager:info("Search for multivalued_field has not yet yielded data"),
+                false
             end
         end,
     yz_rt:wait_until(Cluster, F),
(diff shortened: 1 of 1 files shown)


riak_test/yz_pb.erl
```diff
@@ -240,7 +240,9 @@ confirm_multivalued_field(Cluster) ->
                     1 -> true;
                     0 -> false
                 end;
-            _ -> false
+            {ok, {search_results, [], _Score, 0}} ->
```
jvoegele, line 243: The "catch all" case has been replaced with the tagged tuple above. Is that guaranteed to always match, or should there be an additional clause that catches anything that might sneak through?

fadushin, line 243: I wanted it to fail in that case with a bad match, on the assumption that matching against {ok, {search_results, [], _Score, 0}} is an indication that the data has not yet been indexed.  In that case, we should wait.  Otherwise, we should fail the test.

jvoegele:
:+1: f60bfac8eaad778bbd13a69e0e62ca472fbfee2f

zeeshanlakhani:
@borshop merge

PR #546, opened 2015-08-31
make sure disabled for yz is treated separately and correctly

zeeshanlakhani:
As per discussions w/ @fadushin and @jonmeredith, this handles separately disabling yz aae correctly. The handling of entropy_mgr calls was actually the same as riak_kv... so... there's still room for cleanup, but this is something.

### DIFF
--- a/src/yz_kv.erl
+++ b/src/yz_kv.erl
@@ -402,10 +402,8 @@ first_partition([{Partition, _}|_]) ->
 
 %% @private
 %%
-%% @doc Get the tree.
+%% @doc Get the tree if yz anti_entropy is enabled.
 %%
-%% NOTE: Relies on the fact that this code runs on the KV vnode
-%%       process.
 -spec get_tree(p()) -> tree() | not_registered.
 get_tree(Partition) ->
     case erlang:get({tree,Partition}) of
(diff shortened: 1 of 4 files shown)


jonmeredith:
I think it also needs something like this

``` erlang
@@ -383,8 +383,13 @@ first_partition([{Partition, _}|_]) ->
 get_tree(Partition) ->
     case erlang:get({tree,Partition}) of
         undefined ->
-            lager:debug("Tree cache miss (undefined): ~p", [Partition]),
-            get_and_set_tree(Partition);
+            case yz_entropy_mgr:enabled() of
+                true ->
+                    lager:debug("Tree cache miss (undefined): ~p", [Partition]),
+                    get_and_set_tree(Partition);
+                false ->
+                    not_registered
+            end;
         Pid ->
             case is_process_alive(Pid) of
                 true ->
```

Otherwise when AAE is disabled, every call contacts the yz_entropy_mgr and tries to get a tree.

zeeshanlakhani:
@jonmeredith good point, verifying w/ a trace on yz_kv... forgot that this was placed there for `update_hashtree`, derp for fast checking.

src/yz_kv.erl
```diff
@@ -402,12 +402,25 @@ first_partition([{Partition, _}|_]) ->
 
 %% @private
 %%
-%% @doc Get the tree.
+%% @doc Get the tree if yz anti_entropy is enabled.
 %%
-%% NOTE: Relies on the fact that this code runs on the KV vnode
-%%       process.
 -spec get_tree(p()) -> tree() | not_registered.
 get_tree(Partition) ->
```
zeeshanlakhani, line 408: broke it up b/c way too may case statements ;).

jonmeredith, line 408: Why not do the check in the get_and_set_tree, that way the common case of having a tree will just be a hit of the pdict, and the disabled case would be a pdict/appenv:get rather than always appenv:get/pdict:get

zeeshanlakhani, line 408: I was just attempting to even avoid https://github.com/basho/yokozuna/blob/bugfix/zl/handle-yz-aae-disabled/src/yz_kv.erl#L424, but probably makes little difference (and in get_and_set, it'd be cleaner)

zeeshanlakhani:
Updated.

PR #550, opened 2015-09-14
Hotfix/zl/2.0.6 2012.08.25 deletes

zeeshanlakhani:
Patch Code branch

### DIFF
--- a/src/yz_kv.erl
+++ b/src/yz_kv.erl
@@ -265,9 +265,12 @@ index(Obj, Reason, Ring, P, BKey, ShortPL, Index) ->
             LI = yz_cover:logical_index(Ring),
             LFPN = yz_cover:logical_partition(LI, element(1, ShortPL)),
             LP = yz_cover:logical_partition(LI, P),
+            {Bucket, _} = BKey,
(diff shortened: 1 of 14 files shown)


src/yz_kv.erl
```diff
@@ -396,6 +405,19 @@ cleanup(_, _) ->
 
 %% @private
 %%
+%% @doc Cleanup tombstones and siblings accordingly... cleanup/3
+%% TODO: make this work for allow_mult=true case
+-spec cleanup([doc()], bkey(), values()) -> [{bkey, bkey()}].
+cleanup([], _BKey, _Values) ->
+    [];
+cleanup([{doc, Fields}|T], BKey, [VH|VT]) ->
```
jonmeredith, line 413: _DO NOT_ fix for patch, but I don't think we need Values at all - could just work on the Fields as is if a tombstone, VH will always be <<>>, so it's redundant.

zeeshanlakhani, line 413: @jonmeredith good point.. actually was already doing this... but will do when i bring this over to the main PR.

jonmeredith:
+1 c23e1f0

zeeshanlakhani:
@borshop merge

PR #553, opened 2015-09-17
handle yz_kv:index/7 for anti_entropy/delete repair calls in the exch…

zeeshanlakhani:
…ange_fsm, like how we handle the same func in yz_kv:index/3 w/ a try/catch


JeetKunDoug:
:+1: 004c31e

zeeshanlakhani:
@borshop merge

PR #555, opened 2015-09-18
hash needs to match kv for tree comparison, force use of obj0 before …

zeeshanlakhani:
…merge


fadushin:
+1 5871aa4

fadushin:
@borshop merge

PR #556, opened 2015-09-18
Include JTS JAR in Solr WAR

jvoegele:
Patch the ivy.xml file in the Solr source code to include a dependency
on com.vividsolutions.jts:1.13 so that the JTS JAR file is included in
the Solr WAR file that is distributed with Yokozuna.


kesslerm:
Rather than using custom built Solr files, couldn't we just add additional elements to the java classpath generated by [yokozuna](https://github.com/basho/yokozuna/blob/develop/src/yz_solr_proc.erl#L185) and copy the dependencies into a suitable location? The "-cp" [option](https://docs.oracle.com/javase/7/docs/technotes/tools/windows/classpath.html) can take a colon (on Unix) separated list of jar files. Alternatively, Yokozuna could set the CLASSPATH environment variable before launching java.

jvoegele:
Hi @kesslerm , I think the problem with that approach is that it necessitates installing the JTS JAR on each node and configuring the class path to point to it on each node. By including the JAR in the Solr distribution that we ship with Yokozuna, we eliminate this operational overhead.

Or did I misunderstand your suggestion?

kesslerm:
Distribution would happen in the same way as the Solr files, just add the necessary jars into the priv directory. And then change the hard-coded class path in Yokozuna. I assume you want this new dependency to become a standard dependency of Yokozuna. What I'd like to avoid is a situation where we start building too many of our own changes into the actual Solr distribution. It'd be great if at some later point in time we could just replace a solr file with an upstream version. Up until now we carry one extra patch, but even that could move out into a separate jar.

JeetKunDoug:
@jvoegele @kesslerm  Although we are familiar and comfortable with including the JTS jars, we'd like to maintain control of which Solr/additional libraries are included/used within Riak, as opening up to any Solr version/3rd-party libs could cause us some fairly large support headaches.

That said, as Riak 2.0 is now a "Long Term Support" branch, @jvoegele can you please retarget this PR onto the 2.1 branch?

zeeshanlakhani:
Actually, probably best for 2.next @JeetKunDoug, per conversations (new feat).

PR #562, opened 2015-09-24
handle fuse events for recovered and make yz_events a gen_event behavior

zeeshanlakhani:
- an idea ;) which still needs a test. 
- Make yz_events into a gen_event behavior.
- Handle fuse events
- Create recovered statistic through track of index state when fuses go from blown->ok via a yz_events ets table.


zeeshanlakhani:
Closing for PR into new branch.

PR #566, opened 2015-09-30
handle fuse events for recovered and make yz_events a gen_event behavior

zeeshanlakhani:
- allows for recover stat
- changes yz_events from gen_server to gen_events behavior and adds sup needs
- includes the move to fuse 2.1.0 and code around removal, which was added recently: https://github.com/jlouis/fuse/pull/8.

### DIFF
--- a/src/yz_events.erl
+++ b/src/yz_events.erl
@@ -93,9 +93,7 @@ handle_event({Index, blown}, S) ->
 handle_event({Index, ok}, S) ->
     handle_index_recovered(Index, up),
     {ok, S};
-handle_event({Index, reset}, S) ->
-    %% TODO: We are currently using reset when an Index is removed, but there
-    %% should be a true removal of fuse eventually.
+handle_event({Index, removed}, S) ->
     handle_index_recovered(Index, removed),
     {ok, S};
 handle_event(_Msg, S) ->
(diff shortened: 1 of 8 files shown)


src/yz_events.erl
```diff
@@ -106,6 +133,17 @@ terminate(_Reason, _S) ->
 %%% Private
 %%%===================================================================
 
+%% @doc Called by {@link yz_general_sup} to create the p ETS table used to
+%% track registered events. Created when we add the handler after the supervisor
+%% is already up for yz_events.
+-spec create_table() -> ok.
+create_table() ->
+    _ = ets:new(?MODULE, [named_table, public, set,
```
zeeshanlakhani: It's been a few weeks since I wrote this, but I remember `protected` not working here. When you review @jvoegele / @fadushin, you can try and see if it works... update the code, yay :dancer:

src/yz_events.erl
```diff
@@ -59,7 +70,26 @@
...
+
+add_callback(Fn) when is_function(Fn) ->
+    gen_event:add_handler(?MODULE, {?MODULE, make_ref()}, [Fn]).
+
+add_sup_callback(Fn) when is_function(Fn) ->
+    gen_event:add_sup_handler(?MODULE, {?MODULE, make_ref()}, [Fn]).
+
+add_guarded_callback(Fn) when is_function(Fn) ->
+    yz_eventhandler_sup:start_guarded_handler(?MODULE, {?MODULE, make_ref()},
+                                          [Fn]).
```
fadushin, line 92: A lot of these add_ calls are not actually used in the code, except current, add_guarded_handler.  We should either remove these, if they are not needed, or document why they are there.

fadushin:
+1 5c837ce

fadushin:
@borshop merge

PR #567, opened 2015-10-06
handle cleanup for delete operations to Solr, fixing up allow-mult=true

zeeshanlakhani:
- rewrite cleanup method into cleanup and cleanup_for_sibs, the latter for handling allow-mult=true objects' sibling reconciliation and tombstones
- make yz_siblings r_t use delete_mode=keep so that we can test how yz handles cleanup for allow-mult=true reconciliation/delete

### DIFF
--- a/riak_test/yz_siblings.erl
+++ b/riak_test/yz_siblings.erl
@@ -4,7 +4,18 @@
 -include_lib("eunit/include/eunit.hrl").
 
 -define(FMT(S, Args), lists:flatten(io_lib:format(S, Args))).
--define(CFG, [{yokozuna, [{enabled, true}]}]).
+-define(CFG, [
+              {riak_kv,
+               [{delete_mode, keep}]},
+              {yokozuna,
+               [{enabled, true}]}
+             ]).
+-define(BODIES,
+        [<<"This is value alpha">>,
+         <<"This is value beta">>,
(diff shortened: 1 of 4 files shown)


riak_test/yz_siblings.erl
```diff
@@ -88,7 +137,8 @@ reconcile_sibs(Cluster, HP, Index, Bucket, EncKey) ->
 delete_key(Cluster, HP, Index, Bucket, EncKey) ->
     lager:info("Delete the key"),
     URL = bucket_url(HP, Bucket, EncKey),
-    {ok, "200", RH, _} = ibrowse:send_req(URL, [], get, [], []),
+    {ok, Status, RH, _} = ibrowse:send_req(URL, [], get, [], []),
+    ?assert(Status == "200" orelse Status == "300"),
```
fadushin, line 141: I am curious why checking for 300 is now required, but I am not completely up on the Riak HTTP API.  Will Riak send back a 300 after a partition heal?

zeeshanlakhani, line 141: cause we have siblings, so we may get multiple pieces content.

fadushin:
+1 cd4badc

zeeshanlakhani:
@borshop merge

PR #577, opened 2015-10-23
Include JTS JAR in Solr WAR

jvoegele:
Patch the ivy.xml file in the Solr source code to include a dependency on `com.vividsolutions.jts:1.13` so that the JTS JAR file is included in the Solr WAR file that is distributed with Yokozuna.

This is #556 rebased on `develop`.


jvoegele:
This should be against `master` instead of `riak/2.0`, so recreating.

zeeshanlakhani:
Closing until we figure out how to handle licensing.

PR #565, opened 2015-09-29
Features+fixes/zl/hotfix 2.0.6 2012.08.25 for 2.0

zeeshanlakhani:
- New code containing fuses, batching, queues
- New work revolved around these updates should be PR'ed into this branch until we can go safely into 2.0

PR #608, opened 2016-02-02
Features+fixes/zl/cuttlefishisms+ibrowse upgrade+fuse handling around ed

zeeshanlakhani:
- Adds fuse/batch settings to cuttlefish w/ a few validations
- More types... one TODO for @fadushin as well!
- Don't attempt to build tree if Indexes are blowing during folding of keys for a partition. 
- Upgrade to ibrowse 4.2.2 -> This causes bors to fail b/c the webmachine dependency too. I'm trying to get access so I can make that change. 
- This also includes the logging fix which is part of #565 now.

### DIFF
--- a/dialyzer.ignore-warnings
+++ b/dialyzer.ignore-warnings
@@ -96,3 +96,4 @@ Unknown functions:
   rt:heal/1
   rt:partition/2
   rt:wait_until_transfers_complete/1
+  yz_solr_orig:index_batch_orig/2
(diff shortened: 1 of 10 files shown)


fadushin:
Do all the riak tests pass with the ibrowse change?

zeeshanlakhani:
@fadushin They all did for me off the 2.0.6 version I tried. I'll re-run everything except AAE, as you fix that one, which again was failing beforehand.

zeeshanlakhani:
@fadushin will re-run through.

Issue #581, opened 2015-11-02
Cleanup/prep YZ startup + shutdown [JIRA: RIAK-2253]


zeeshanlakhani:
- Flush the queue on graceful shutdown in yz_app:prep_stop(). 
- Should also call yokozuna:disable(index) (maybe others?) at that point too and maybe unregister API calls.

Related to need(s) in https://github.com/basho/yokozuna/issues/582.

zeeshanlakhani:
Also, todo: https://github.com/basho/yokozuna/pull/565/files#diff-f42188ad89ac358833c9b82d8622c226R44.

Look at prep_stop on kv: https://github.com/basho/riak_kv/blob/2.0/src/riak_kv_app.erl#L226 + http://learnyousomeerlang.com/the-count-of-applications.

jvoegele:
Write a riak_test to verify startup/shutdown behavior.

jvoegele:
@zeeshanlakhani Can you say more concretely what you mean by "flush the queue"? Is there a specific function to call or a message to send to some process?

zeeshanlakhani:
@jvoegele fyi, if I wasn't explicit before, flush the queue, means flush to Solr (which is flushing by readying a worker/batch), but you probably know that via the singular index flush call in https://github.com/basho/yokozuna/pull/565/files#diff-e664002ad0421c9debe8a6ec2b5732b0R198. But, just being double sure.

PR #610, opened 2016-02-03
Clean up startup/shutdown sequence for yokozuna

jvoegele:
See issue https://github.com/basho/yokozuna/issues/581 for background.
- Ensure that all of the solrqs are drained at shutdown.
- Move shutdown logic to `yz_app:prep_stop` function.
- General cleanup of the startup and shutdown code.
- Add a new riak_test (`yz_startup_shutdown`) that tests all of the above.

### DIFF
--- a/src/rt_intercept_pt.erl
+++ b/src/rt_intercept_pt.erl
@@ -1,0 +1,91 @@
+%% XXX
+%% -------------------------------------------------------------------
+%% TODO: This file was copied from riak_test in order to allow for using
+%% the parse transform from a riak_test whose source code is in the
+%% Yokozuna repository. This file should be deleted if or when a better
+%% mechanism is found for using the rt_intercept_pt parse transform
+%% external to the riak_test repository.
+%% -------------------------------------------------------------------
+%% XXX
+
+-module(rt_intercept_pt).
+-export([parse_transform/2]).
+
+%% This parse transform looks for calls to rt_intercept:add/2, and if found
+%% potentially modifies the second argument. The second argument can be a
+%% list of intercept tuples or a single intercept tuple. An intercept tuple
+%% can have either 2 or 3 elements, but either way, a final element of the
+%% form
+%%
+%%   [{{F,Arity},{[var], fun()}}]
+%%
+%% is transformed into
+%%
+%%   [{{F,Arity},{[{varname, var}], tuple()}}]
+%%
+%% Only the second element of this tuple is modified. In the first form the
+%% fun() is an anonymous interceptor function and [var] represents the list
+%% of free variables used within the function but defined in the context in
+%% which the function is defined. The list of vars is transformed into a
+%% list of 2-tuples of var name and var, while the function is replaced
+%% with its abstract format (which, since we are already dealing with
+%% abstract format, is actually the abstract format of its abstract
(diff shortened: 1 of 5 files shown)


src/rt_intercept_pt.erl
```diff
@@ -0,0 +1,91 @@
+%% XXX
+%% -------------------------------------------------------------------
+%% TODO: This file was copied from riak_test in order to allow for using
+%% the parse transform from a riak_test whose source code is in the
+%% Yokozuna repository. This file should be deleted if or when a better
+%% mechanism is found for using the rt_intercept_pt parse transform
+%% external to the riak_test repository.
+%% -------------------------------------------------------------------
+%% XXX
```
jvoegele, line 9: @zeeshanlakhani Note the temporary hack to get the parse transform to work. ^^^

src/yz_app.erl
```diff
@@ -58,16 +60,46 @@ start(_StartType, _StartArgs) ->
...
     end.
 
+%% @doc Prepare to stop - called before the supervisor tree is shutdown
+prep_stop(State) ->
+    try %% wrap with a try/catch - application carries on regardless,
+        %% no error message or logging about the failure otherwise.
+        lager:info("Stopping application yokozuna.\n", []),
+        ok = riak_api_pb_service:deregister(?QUERY_SERVICES),
+        ok = riak_api_pb_service:deregister(?ADMIN_SERVICES),
+        ok = yz_solrq_drain_mgr:drain(),
```
fadushin: I am not sure we want to pattern match against ok, or if we do, we may want to continue with the disable_components even if the drain fails, times out, errors, or is already in progress.

src/yz_app.erl
```diff
@@ -58,16 +60,46 @@ start(_StartType, _StartArgs) ->
...
+        Type:Reason ->
+            lager:error("Stopping application yokozuna - ~p:~p.\n",
+                        [Type, Reason])
+    end,
+    State.
+
 stop(_State) ->
-    ok = riak_api_pb_service:deregister(?QUERY_SERVICES),
-    ok = riak_api_pb_service:deregister(?ADMIN_SERVICES),
+    lager:info("Stopped application yokozuna.\n", []),
```
fadushin: These lager lines have \newlines appended to them, but I don't think they are necessary (from what little I know about lager.  We should check with @macintux or @mrallen1.  Also, I'd prefer we use ~n, as \n is not platform independent.  (Yes, we don't support Windows today, but who knows?)

zeeshanlakhani: Yep, `\n` is not needed unless you're doing so explicitly between lines.

src/yz_app.erl
jvoegele: Yeah, I considered removing all of the `ok =` pattern matches but left them there in the end since they were in the original `stop` code. What do you think? Should I zap them all? Also, any thoughts on the order of operations in `prep_stop`?

fadushin: I would definitely deregister the PB and WebMachine services (if the latter is do-able) first.  I don't know about the pattern matching on those deregisters, but I know that there are a few non-ok results that could come back.  (Details TBD, as we nail down the desired behaviors)

zeeshanlakhani:
Will be incorporated into another PR.

PR #615, opened 2016-03-01
cleanup guards on api calls and cuttlefish schema & tests

zeeshanlakhani:
@fadushin lmk if i'm missing more. We'll continue to clean-it up. 

Use -> https://github.com/basho/riak_ee/tree/release_2.0.7-zl-update_reltool_for_fuse cause fuse boot, but make sure you update rebar.config for the kv branch, etc...
Use -> https://github.com/basho/riak_test/pull/1005 for _riak_test_
## WIP Things Left

@fadushin 
- [x] Tests for Purge Strategies
- [x] Finish stat renaming


riak_test/yz_solrq_test.erl
```diff
@@ -124,21 +163,348 @@ confirm_requeue_undelivered([Node|_] = Cluster, PBConn, BKey, Index) ->
...
+                    proplists:get_value(<<"_yz_rt">>, Fields),
+                    proplists:get_value(<<"_yz_rb">>, Fields)
+                },
+                proplists:get_value(<<"_yz_rk">>, Fields)
+            }
+        end,
+        SearchResults
+    ).
+
+%% TODO Revisit re-using Indices and Buckets
```
zeeshanlakhani: Can probably move the TODO to where we list out the Bucket macros and mention that we should eventually reuse indexes in this test and change the `delete_mode`.

riak_test/yz_solrq_test.erl
```diff
@@ -124,21 +163,348 @@ confirm_requeue_undelivered([Node|_] = Cluster, PBConn, BKey, Index) ->
...
+        %% Send two messages through each indexq on the solrq, which
+        %% will trip the fuse on both; however
+        %% because fuse blown events are handled asynchronously,
+        %% we need to sleep enough to allow the solrqs to get notified.
+        %%
+        [Index1BKey1] = put_bkey_objects(PBConn, [Index1BKey1]),
+        [Index1BKey2] = put_bkey_objects(PBConn, [Index1BKey2]),
+        [Index2BKey1] = put_bkey_objects(PBConn, [Index2BKey1]),
+        [Index2BKey2] = put_bkey_objects(PBConn, [Index2BKey2]),
+        timer:sleep(500), % wait for notification
```
zeeshanlakhani: I see the message, but anyway we can make this a little more robust?

riak_test/yz_solrq_test.erl
```diff
@@ -124,21 +163,348 @@ confirm_requeue_undelivered([Node|_] = Cluster, PBConn, BKey, Index) ->
...
+            _ ->
+                F()
+        end
+    after
+        %%
+        %% Revert the intercept, and drain, giving time for the
+        %% fuse to reset.  Commit to Solr so that we can run a query.
+        %%
+        intercept_index_batch(Node, index_batch_call_orig),
+        timer:sleep(?MELT_RESET_REFRESH + 1000), %% wait for fuse reset
```
zeeshanlakhani: same as above in terms of robustness, maybe we can listen for the event?

riak_test/yz_rt.erl
```diff
@@ -615,3 +615,6 @@ load_intercept_code(Node) ->
                               "*.erl"]),
     rt_intercept:load_code(Node, [CodePath]).
 
+-spec rpc_call_cluster([node()], Module :: atom(), Function :: atom(), [term()]) -> [term()].
+rpc_call_cluster(Cluster, Module, Function, Args) ->
+    [rpc:call(Node, Module, Function, Args) || Node <- Cluster].
```
jvoegele: Is this any different from Erlang's built-in `rpc:multicall/4`?

http://erlang.org/documentation/doc-5.1/lib/kernel-2.7/doc/html/rpc.html#multicall%4

fadushin: Good catch.  I suspect we should be using riak_core_util:multi_rpc, but I don't think we have riak_core in riak_test's deps.

riak_test/yz_solrq_test.erl
```diff
@@ -55,6 +80,16 @@ confirm() ->
...
     ok = create_indexed_bucket(PBConn, Cluster, ?BUCKET3, ?INDEX3),
+    ok = create_indexed_bucket(PBConn, Cluster, ?BUCKET4, ?INDEX4),
+    ok = create_indexed_bucket(PBConn, Cluster, ?BUCKET5, ?INDEX5),
+    ok = create_indexed_bucket(PBConn, Cluster, ?BUCKET6, ?INDEX6),
+    ok = create_indexed_bucket(PBConn, Cluster, ?BUCKET7, ?INDEX7),
+    ok = create_indexed_bucket(PBConn, Cluster, ?BUCKET8, ?INDEX8),
+    ok = create_indexed_bucket(PBConn, Cluster, ?BUCKET9, ?INDEX9),
+    ok = create_indexed_bucket(PBConn, Cluster, ?BUCKET10, ?INDEX10),
+    ok = create_indexed_bucket(PBConn, Cluster, ?BUCKET11, ?INDEX11),
+    ok = create_indexed_bucket(PBConn, Cluster, ?BUCKET12, ?INDEX12),
```
jvoegele: In #621 I moved this `create_indexed_bucket` function into the `yz_rt` module so that it could be reused by other tests. We'll have to do some conflict resolution at some point.

riak_test/yz_solrq_test.erl
```diff
@@ -124,21 +164,326 @@ confirm_requeue_undelivered([Node|_] = Cluster, PBConn, BKey, Index) ->
...
+    lager:info("confirm_purge_one_strategy ok"),
+    ok.
+
+check_one_purged({{Index1Written, Index1SearchResults},
+                  {Index2Written, Index2SearchResults}} = TestResults) ->
+    Condition =
+        equal(Index1Written, Index1SearchResults)
+        and first_purged(Index2Written, Index2SearchResults)
+        orelse first_purged(Index1Written, Index1SearchResults)
+        and equal(Index2Written, Index2SearchResults),
```
jvoegele: Why the mixture of the short-circuiting `orelse` with the strict `and` (instead of `andalso`)?

fadushin: Good point.  We can use andalso here as well.

fadushin:
+1 5a87989

fadushin:
@borshop merge

PR #616, opened 2016-03-04
Ensure only valid unicode binaries are extracted from datatypes

Vorticity-Flux:
This patch ensures that only valid unicode strings are passed to SOLR.
#### Use case:

CRDT map with a mix of valid unicode and non-unicode binaries is being indexed by yokozuna.
Only fields that contain valid unicode are added to the SOLR schema.
#### Current behaviour:

Indexing fails in mochijson2 due to invalid UTF8.
#### Expected behviour:

Non-indexable (non-unicode) fields should not prevent indexing of other fields that contain valid unicode.

### DIFF
--- a/src/yz_dt_extractor.erl
+++ b/src/yz_dt_extractor.erl
@@ -103,7 +103,7 @@ extract_fields(Name, counter, Value, #state{fields=Fields, field_separator=Sep}=
     State#state{fields=[{FieldName, ?INT_TO_BIN(Value)}|Fields]};
 extract_fields(Name, register, Value, #state{fields=Fields, field_separator=Sep}=State) ->
     FieldName = field_name(Name, register, Sep),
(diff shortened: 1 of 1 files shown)


zeeshanlakhani:
@Vorticity-Flux thank you for this. After some thinking, I think that we should allow for two things. Firstly, we should only skip invalid UTF8 values/elems via an option b/c some users may actually be trying to index invalid UTF8 values/elems and should receive an error. So, I'm going to close this for now, but will reference it in an issue I'm writing up. We'll A) have better error'ing instead of that mochijson2 error you got and B) allow for a configurable option to skip all invalid utf8 values and force the ensure... as you did in your code. 

For now, the fix is doing your patch or just forking the extractor into a new one and using that. Thanks and I'll update you once we have it in for release :).

PR #618, opened 2016-03-07
Use Base64 encoded names for fuses

jvoegele:
Previously we had been directly using the name of each Solr index as the
name of the fuse that was associated with it. However, since the index
names are arbitrary names chosen by users, it is possible that these
names could clash or contain characters that are not legal for fuse
names. Therefore, we now apply a prefix to the fuse name to make sure
that it does not clash, and then apply Base64 encoding to that name to
ensure that it does not contain illegal characters.
- [ ] Do fuse stats continue to collect/work/able-to-call-by-index-name via tests


zeeshanlakhani:
@jvoegele as @fadushin was gonna work on stat-name changes and tests... of note, please make sure/test all fuse stats as well. For example, I don't see a change to https://github.com/basho/yokozuna/blob/feature/jv/flush-queues/src/yz_fuse.erl#L143 either, so that will need fixing, right?

jvoegele:
@zeeshanlakhani The encoding of the fuse name is encapsulated within the `yz_fuse` module and only affects the name of the fuse that we use for interacting with the `fuse` library. All of the other functions continue to use the `Index` name directly, so I don't think we need to change anything about https://github.com/basho/yokozuna/blob/feature/jv/flush-queues/src/yz_fuse.erl#L143 . I am still working separately on the riak_test stuff for the stats, etc.

zeeshanlakhani:
@jvoegele figured... just want to make sure w/ tests of course :).

zeeshanlakhani:
Also, to check that https://github.com/basho/yokozuna/blob/feature-jv-namespaced_fuses/src/yz_fuse.erl#L172 works correctly.

src/yz_fuse.erl
```diff
@@ -26,14 +26,22 @@
 -export([create/1, check/1, check_all_fuses_ok/1, melt/1, remove/1, reset/1]).
 
 %% helpers
--export([fuse_context/0]).
+-export([fuse_context/0, encode_fuse_name/2, decode_fuse_name/1,
+         fuse_name_for_index/1]).
```
fadushin: Do we need to export encode/decode fuse name?  They only seem to be used internally.  I see that fuse_name_for_index is used in the EQC test.  Can we maybe put that in a separate export with a comment that makes it clear that the function is only used in a test?

fadushin:
@zeeshanlakhani I don't see where get_stats_for_index is called:

> find . -type f -not -path "_.eunit" -not -path "_.git*" -exec grep "get_stats_for_index" {} \; -print
> Binary file ./ebin/yz_fuse.beam matches
> ./ebin/yz_fuse.beam
> -export([aggregate_index_stats/2, stats/0, get_stats_for_index/1]).
> -spec get_stats_for_index(atom()) -> ok.
> get_stats_for_index(Index) ->
> ./src/yz_fuse.erl

Should this function be removed all together?  Or am I missing something painfully obvious?

zeeshanlakhani:
Meant for debug/cse  use until clique work is done.

src/yz_fuse.erl
jvoegele: Will do, @fadushin.

On a side note, the Eiffel language had a feature where you could export a function to specific callers. I never saw much utility in it, but it strikes me that functions exported to tests only would be a perfect fit.

rzezeski: It's been a long time since I wrote any Erlang but I believe this should work.

``` erlang
-ifdef(TEST).
-export([foo/0, bar/0]).
-endif.
```

zeeshanlakhani: @jvoegele original Z / @rzezeski is correct.

fadushin:
+1 f0d07d928c3bc5636d44ec749906ff15b842d8e3

fadushin:
@borshop merge

PR #621, opened 2016-03-14
Add yz_fuse_upgrade test

jvoegele:
This test confirms that fuses become associated with indexes that were
created either before or after a rolling upgrade. This commit also
contains a small amount of refactoring to move some useful functions
out of specific tests and into the yz_rt module where they can be reused
by other tests.


fadushin:
+1 0351a29

fadushin:
@borshop merge

PR #624, opened 2016-03-25
cleanup around monitor for sync index proc and refactor flatten/reverse for batchies

zeeshanlakhani:
- monitor process to sync, which calls child process to wait
- refactor reverse/flatten of ops to be closer to ops code

@fadushin testy + looky.

### DIFF
--- a/src/yz_index.erl
+++ b/src/yz_index.erl
@@ -42,6 +42,15 @@
         }).
 
 -type index_info() :: #index_info{}.
+-type create_index_return() ::
+        ok |
+        {error, index_not_created_within_timeout} |
+        {error, schema_not_found} |
+        {error, invalid_name} |
+        {error, core_error_on_index_creation, binary()} |
+        {error, term()}.
+
+-export_type([create_index_return/0]).
 
 
 %%%===================================================================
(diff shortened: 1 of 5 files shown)


fadushin:
One thing I'd like to do before +1'ing the list:reverse changes to yz_solrq_helper is to temporarily modify the mapping from vnodes to Solrqs from the existing BKey hash back to the old partition hash, and then uncomment the portions of the EQC test that check for preservation of key ordering-by-vnode, and see if we are still passing EQC with retries.  That would give some confidence that the list_reverse changes in yz_solrq_helper don't affect ordering off retries.  (And also run the EQC test with many iterations).  I can try this, but if anyone else has time, give a holler.

zeeshanlakhani:
The code changes aren't different then what was already on the flush-queues branch btw, just moved closer to ops. But if you want to update the pulse test to check ordering, go for it :).

Zeeshan Lakhani
programmer | 
software engineer at @basho | 
org. member/founder of @papers_we_love | paperswelove.org
twitter => @zeeshanlakhani

fadushin:
Tests pass when reverting hashing strategy, so I confident the retry ordering is not affected by this change.

fadushin:
+1 7f543a3

fadushin:
@borshop merge

PR #626, opened 2016-04-05
Fix intermittent yz_rs_migration_test failures

nickelization:
The yz_rs_migration_test would occasionally fail during the final query_data step, due to reading slightly fewer values than expected. After much debugging, this turned out to be due to AAE temporarily removing data from YZ in certain situations. To fix it, we can simply disable AAE prior to executing the last part of the test.

To go into a bit more detail, the issue was due to the following sequence of steps occurring:
1. basho_bench begins loading data
2. A YZ AAE exchange begins to execute, and snapshots the KV hashtree
3. basho_bench continues writing data, which (among other things) inserts new hashes into the YZ AAE tree
4. The AAE exchange snapshots the YZ tree, which now has new data that is not present in the KV snapshot
5. The AAE exchange removes the new data from Yokozuna as a `remote_missing` repair
6. The test continues with the basho_bench query step, and can't find all the data that was supposed to have been written, failing the test
7. The next AAE exchange happens and repairs the missing data

I've also added a call to yz_rt:drain_solrqs in the load_data step. I'm not sure if this is necessary in practice, but can't hurt to be safe and make sure all the queues are drained before we turn around and try to query the associated data.

### DIFF
--- a/riak_test/yz_rs_migration_test.erl
+++ b/riak_test/yz_rs_migration_test.erl
@@ -118,6 +118,12 @@ confirm() ->
             restart(Cluster),
             check_for_errors(Cluster),
 
+            %% Disabling anti-entropy here prevents a race condition we would occasionally
+            %% see when AAE kicked in during the load_data step. It was possible for more
+            %% data to be written to YZ between the KV and YZ trees being snapshotted, which
+            %% would then cause the data to be removed from YZ as a remote_missing repair.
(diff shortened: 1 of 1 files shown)


fadushin:
+1 ecda2f2

nickelization:
@borshop merge

JeetKunDoug:
:+1: thanks @nickelization

Issue #378, opened 2014-05-12
range facets with mincount inconsistent [JIRA: RIAK-1694]


rzezeski:
Note: This issue is based off an index card I wrote nearly a year ago
so I don't have an exact example but it should be easy to reproduce if
still an issue.

While building a RICON demo I noticed that setting `facet.mincount` to
1 during a range facet produced inconsistent results across queries
(when the query plan changes).  I believe the cause to be that
mincount is applied to each shard separately rather than for the
combined result.

Attempt to reproduce and determine the root cause.  A fix may require
a patch to Solr upstream.

DSomogyi:
Comment for Jira.

zeeshanlakhani:
@jvoegele as part of #571 / #530, can you test this?

zeeshanlakhani:
@JeetKunDoug @jvoegele this has not be tested in #571 yet... will be added as a separate pr.

zeeshanlakhani:
closing as per #627.

PR #627, opened 2016-04-12
Add testing of facet.mincount to yz_faceted_search test

jvoegele:
This test is to verify that #378 is no longer an issue.

I updated `yz_faceted_search` test to include the `facet.mincount` parameter and also to execute the faceted query multiple times to increase the likelihood of an error occurring. I also tried several different values for the `facet.mincount` parameter, but no error was triggered.

### DIFF
--- a/riak_test/yz_faceted_search.erl
+++ b/riak_test/yz_faceted_search.erl
@@ -127,8 +127,14 @@ verify_field_faceting(Cluster, Index) ->
 verify_query_faceting(Cluster, Index) ->
     HP = yz_rt:host_port(Cluster),
     Params = [{facet, true},
+              {'facet.mincount', 1},
               {'facet.query', "price:[1 TO 29]"},
               {'facet.query', "price:[30 TO 100]"}],
+    %% Verify 10 times because of non-determinism in coverage
+    [verify_query_facets(HP, Index, Params) || _ <- lists:seq(1, 10)],
+    ok.
+
+verify_query_facets(HP, Index, Params) ->
     lager:info("Query faceting: ~p, ~p, ~p", [HP, Index, Params]),
     {ok, "200", _Hdr, Res} = yz_rt:search(HP, Index, "state", "Ohio", Params),
     Struct = mochijson2:decode(Res),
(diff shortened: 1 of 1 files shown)


zeeshanlakhani:
@jvoegele thank you for this. Can we also add a mincount test that tests something other than the default `1`? Just as precaution, double checking. If so, I'd like to get this in for 2.0.7 for our test runs. ^^ @JeetKunDoug @fadushin.

jvoegele:
@zeeshanlakhani I added a test with facet.mincount > 1

riak_test/yz_faceted_search.erl
```diff
@@ -120,15 +120,24 @@ verify_field_faceting(Cluster, Index) ->
                             <<"facet_fields">>,
                             <<"state">>],
                            Struct),
-    ?assertEqual([<<"Ohio">>,4,<<"Kentucky">>,1], StateCounts).
+    %% We expect to see all 4 Ohio restaurants in the search results, but none
+    %% of the Kentucky restaurants because there is only one in the test data
+    %% set, and facet.mincount is set to 3.
+    ?assertEqual([<<"Ohio">>,4], StateCounts).
```
fadushin, line 126: Looks good.  Odd response format from Solr with multiple facet returns, but (shrug) :)

riak_test/yz_faceted_search.erl
```diff
@@ -120,15 +120,24 @@ verify_field_faceting(Cluster, Index) ->
...
 -define(PRICE_RANGE_1, <<"price:[1 TO 29]">>).
 -define(PRICE_RANGE_2, <<"price:[30 TO 100]">>).
 verify_query_faceting(Cluster, Index) ->
     HP = yz_rt:host_port(Cluster),
     Params = [{facet, true},
+              {'facet.mincount', 1},
               {'facet.query', "price:[1 TO 29]"},
               {'facet.query', "price:[30 TO 100]"}],
+    %% Verify 10 times because of non-determinism in coverage
+    [verify_query_facets(HP, Index, Params) || _ <- lists:seq(1, 10)],
```
fadushin, line 137: I realize we use this pattern elsewhere, but I wonder if we should be checking the response, which I believe has the query plan that was used, and wait for the plan to change.  Not a blocker by any means, but we should consider a strategy for doing this test more deterministically in the future.

fadushin:
+1 2901464

fadushin:
@borshop merge

PR #629, opened 2016-04-18
Purge and unblock vnodes when fuses blow, in order to unblock wedged vnodes

fadushin:
This PR covers a number of issues that were uncovered in the process of adding tests for Yokozuna statistics, specifically, issues around blocking of vnodes when we hit the HWM and repercussions of some the renaming of YZ fuses we did.

Code changes:
- Purge and unblock vnodes when fuses blow, in order to unblock wedged vnodes that started waiting for space to clear before the fuse was blown.
- Fuse blown/recovered events are now cached so that we only notify solrqs once -- when the fuse blows and when the fuse recovers.  Repetitive blown events when the slogs have already been notified are suppressed, and ditto for recovered events.
- Fixed an issue with naming of fuses, which was causing the fuse check to fail to detect a blown index when writing to Solr.
- Check to ensure fuses not blown, instead of fuses just being ok, before initiating a hashtree rebuild.  This change prevents a misleading error from being logged during startup.
- Changed the measurement of the drain latency to cover just the drain time, instead of the drain time and the YZ AAE update
- Reverted subtraction of vnode count from HWM config value
- Changed _cnt config settings to _count
- Added a reset operation to yz_stats, and changed name of aae_repairs stat to "detected_repairs"
- Fixed a lager FORMAT_ERROR in yz_doc when the extractor fails

Test Changes:
- Added tests for new stats
- Refactored some common functionality in yz_rt
- Refactored yz_rt:create_indexed_bucket to allow specification of an n_val
- Made N_VAL truely configurable in yz_aae_test


JeetKunDoug:
:+1: d2c2fd5

JeetKunDoug:
@borshop merge

PR #632, opened 2016-04-29
Use correct yokozuna environment variable for Solr entropy data timeout

kesslerm:
Also fix typo in related documentation.

In yokozuna.hrl the variable is called `solr_ed_request_timeout`. Make the cuttlefish mapping match this.


kesslerm:
Rebasing against feature-solr-batching-rebased branch. New pull request there with updated test case.

PR #634, opened 2016-04-29
Batching and asynchronous delivery into Solr

fadushin:
This PR significantly alters the way data is written into Solr from Yokozuna, primarily by decoupling Solr writes from the VNodes from which index operations originate, and queuing and batching operations into Solr, for better performance and throughput.

A detailed list of changes is too extensive to list here, but a detailed description is available in the new batching document, which is on the branch to be merged:

https://github.com/basho/yokozuna/blob/feature-solr-batching-rebased/docs/BATCHING.md

This document will ultimately be merged into 2.0, as well.  Please use this document as an architectural overview as part of the review.


priv/yokozuna.schema
```diff
@@ -59,3 +59,232 @@
...
+%% @doc To 'enable/disable' AAE for Search.
+%%      * active: out-of-sync keys will be repaired in the background
+%%      * passive: out-of-sync keys are only repaired on read
+%%      * active-debug: like active, but outputs verbose debugging
+%%        information
+{mapping, "search.anti_entropy", "yokozuna.anti_entropy", [
+  {datatype, {enum, [active, passive, 'active-debug']}},
+  {default, active},
+  hidden
+]}.
```
fadushin, line 91: Interesting.  Was the ability to turn on/off YZ AAE (only) not in prior versions?

priv/yokozuna.schema
```diff
@@ -59,3 +59,232 @@
...
+  {datatype, {duration, ms}},
+  hidden
+]}.
+
+%% @doc Determines how fuse operator is called: sync or async.
+{mapping, "search.fuse_context", "yokozuna.fuse_ctx", [
+  {default, async},
+  {datatype, {flag, {sync, sync}, {async, async_dirty}}},
+  hidden
+]}.
```
fadushin, line 149: search.fuse_context is missing from the BATCHING.md description.  Add it.

priv/yokozuna.schema
zeeshanlakhani, line 91: You could still disable it via advanced config, but it wasn't schematized, no.

src/yokozuna.erl
```diff
@@ -227,3 +227,14 @@ search_fold(Results, Start, Params, Positions, Index, Query, Filter, F, Acc) ->
...
+-spec ensure_started(Application :: atom()) -> ok.
+%% @doc Start the named application if not already started.
+%%      Pulled from core, as it's not exported currently
+ensure_started(App) ->
+    case application:start(App) of
+	ok ->
+	    ok;
+	{error, {already_started, App}} ->
+	    ok
+    end.
```
fadushin, line 246: I was worried about the lack of a default clause, so I checked the code, and I don't see the ensure_started call being made anywhere.  Am I missing something?

zeeshanlakhani, line 246: We're using fuse from boot now... so it's not needed. I wanted to keep it in as a possible helper down the road though.

src/yz_app.erl
```diff
@@ -29,25 +29,76 @@
...
             Error
     end.
 
+%% @doc Prepare to stop - called before the supervisor tree is shutdown
+prep_stop(State) ->
+    try %% wrap with a try/catch - application carries on regardless,
+        %% no error message or logging about the failure otherwise.
+        lager:info("Stopping application yokozuna.", []),
+        riak_api_pb_service:deregister(?QUERY_SERVICES),
+        riak_api_pb_service:deregister(?ADMIN_SERVICES),
```
fadushin, line 77: Do we need to worry about web machine services, as well as PB services?

zeeshanlakhani, line 77: I didn't get a chance to review, but from my [initial intentions](https://github.com/basho/yokozuna/issues/581), i think we should try to if possible.

@fadushin also look at cleanly stopping the entropy managers maybe?

src/yz_debug.erl
```diff
@@ -0,0 +1,228 @@
+%% -------------------------------------------------------------------
```
fadushin: I have found this module useful for debugging, but it's debatable if we should include it in a release.  Maybe we should move it to our tools repo?

src/yz_entropy.erl
```diff
@@ -22,32 +22,63 @@
...
+                                           entropy_data_limit, 100)}|Filter],
+            case get_entropy_data(Index, Filter2) of
+                {ok, ED} ->
+                    iterate_entropy_data(Index, Filter2, Fun, ED);
+                {Err, _ED} ->
+                    Err
+            end;
         _ ->
-            ok
+            ?NOTICE("Can't ping Solr to start iterating over entropy data"),
```
fadushin: Perhaps let's include the Index in the message for troubleshooting.

src/yz_events.erl
```diff
@@ -59,7 +70,26 @@
...
+    gen_event:start_link({local, ?MODULE}).
+
+add_handler(Handler, Args) ->
+    gen_event:add_handler(?MODULE, Handler, Args).
+
+add_sup_handler(Handler, Args) ->
+    gen_event:add_sup_handler(?MODULE, Handler, Args).
+
+add_guarded_handler(Handler, Args) ->
+    yz_eventhandler_sup:start_guarded_handler(?MODULE, Handler, Args).
```
fadushin: I was wondering if we could actually just use riak_core_eventhandler_sup and riak_core_eventhandler, instead of duplicating them in this code.

As I understand the rationale for this guard pattern: We want a process that will effectively act as a kind of monitor for yz_events, such that when the guard terminates (say, as a normal shut down of yokozuna), it will terminate the event handler.  Similarly, if the event handler is "deleted" (i.e. removed, if I understand the event_handler docs), the guard will get a en_event_EXIT message.

src/yz_events.erl
```diff
@@ -59,7 +70,26 @@
...
+
+add_callback(Fn) when is_function(Fn) ->
+    gen_event:add_handler(?MODULE, {?MODULE, make_ref()}, [Fn]).
+
+add_sup_callback(Fn) when is_function(Fn) ->
+    gen_event:add_sup_handler(?MODULE, {?MODULE, make_ref()}, [Fn]).
+
+add_guarded_callback(Fn) when is_function(Fn) ->
+    yz_eventhandler_sup:start_guarded_handler(?MODULE, {?MODULE, make_ref()},
+                                          [Fn]).
```
fadushin: It looks like add_guarded_handler is the only one of these API calls that is currently be called from outside of yz_events.  I don't know if that really matters, but I thought it was worth noting in the review.

src/yz_events.erl
```diff
@@ -106,12 +148,24 @@ terminate(_Reason, _S) ->
 %%% Private
 %%%===================================================================
 
+%% @doc Called by {@link yz_general_sup} to create the p ETS table used to
+%% track registered events. Created when we add the handler after the supervisor
+%% is already up for yz_events.
+-spec create_table() -> ok.
+create_table() ->
+    _ = ets:new(?ETS, [named_table, public, set,
```
fadushin, line 138: As far as I can tell, all of the functions in this module that use ?ETS are called within the context of the yz_events process.  Can we therefore make this table private?

zeeshanlakhani: @fadushin Sure, if that's the case. Not sure if it was initially.

src/yz_events.erl
zeeshanlakhani: I was just implementing the general api spec for the variations. Personally, I don't mind the additional options, but they may be unnecessary if we can use the core module that this references.

src/yz_events.erl
zeeshanlakhani: Def. give a try. I thought I did that initially, but maybe we couldn't listen for fuse events when it was private?

src/yz_events.erl
zeeshanlakhani: We may be able too. I don't remember why I didn't initially. I def. thought about and am wondering if there was something too specific related to core w/ it, but maybe not. Sadly, been too long for me to remember.

src/yz_entropy.erl
zeeshanlakhani: +1.

src/yz_debug.erl
zeeshanlakhani: probably so... something we can just patch.

src/yz_events.erl
zeeshanlakhani: doing some tests now, probably works. Probably had diff. intentions initially.

zeeshanlakhani, line 138: @fadushin better note: I think it needs to be public to catch the fuse events. Will double check nonetheless.

zeeshanlakhani, line 138: @fadushin `2016-05-02 17:48:16.675 [error] <0.819.0> gen_event yz_events installed in fuse_event terminated with reason: bad argument in call to ets:lookup(ets_yz_events, <<"yz_stat_test">>) in yz_events:cache_index_state/2 line 313` if it's private.

zeeshanlakhani, line 138: and yep, must be public. Not a necessary change.

src/yz_solrq.erl
```diff
@@ -0,0 +1,817 @@
...
+%%          all the solrqs have completed their drains.  This is a signal to resume
+%%          batching, by moving the aux queue to the main queue, setting the drain
+%%          flag to false (finally), and kick-starting the indexq.
+%% @end
+%%
+handle_cast(drain_complete, #state{indexqs = IndexQs} = State) ->
+    NewIndexQs = dict:fold(
+        fun(Index, #indexq{queue = Queue, queue_len = QueueLen, aux_queue = AuxQueue} = IndexQ, IndexQAccum) ->
+            NewIndexQ = IndexQ#indexq{
+                queue = queue:join(Queue, AuxQueue),
```
fadushin: Shouldn't AuxQueue go before Queue?  In general, Queue should be empty when we receive a drain_complete, and there should be an invariant to that effect, but I am not sure how easy it is to guarantee that.

If not, it seems possible to re-arrange the order of Solr updates during a drain, which could cause strange behavior (e.g., a delete before a write, or conversely), though at this layer, any anomalies might be papered over by AAE.

src/yz_solrq.erl
```diff
@@ -0,0 +1,817 @@
...
+%%      during its prepare state, typically as the result of
+%%      a request to drain the queues.
+%%
+%%      This handler will iterate over all IndexQs in the
+%%      solrq, and initiate a drain on the queue, if it is currently
+%%      non-empty.
+%% @end
+%%
+handle_cast({drain, DPid, Token, Partition}, #state{indexqs = IndexQs} = State) ->
+    ?PULSE_DEBUG("drain{~p=DPid, ~p=Token, ~p=Partition}.  State: ~p~n", [debug_state(State), DPid, Token, Partition]),
```
fadushin: This module needs to be wrapped to 80 or 120 columns, or whatever our coding conventions dictate.

src/yz_solrq.erl
```diff
@@ -0,0 +1,817 @@
...
+    IndexQ3 = maybe_start_timer(Index, IndexQ2),
+    update_indexq(Index, IndexQ3, State);
+
+%%
+%% @doc
+%% This queue is being drained.  There are two cases to consider:
+%%    1. The batch succeeded.  In this case, move any data we
+%%       have accumulated in aux_queue to the main queue,
+%%       and remove ourselves from the list
+%%       of remaining indices that need to be flushed.
```
fadushin: comments are out of date.  Need to review comments in this module, as there were several architectural changes between the time the comments were written and behavior was finalized.

src/yz_solrq.erl
zeeshanlakhani: 80!

src/yz_solrq.erl
zeeshanlakhani: @fadushin seems that your thinking is correct to me.

src/yz_events.erl
zeeshanlakhani: #635 handles this.

src/yz_events.erl
zeeshanlakhani: #635 handles this.

src/yz_app.erl
```diff
@@ -29,25 +29,76 @@
...
+            %% Now everything is started, permit usage by KV/query
+            enable_components(),
             {ok, Pid};
         Error ->
             Error
     end.
 
+%% @doc Prepare to stop - called before the supervisor tree is shutdown
+prep_stop(State) ->
+    try %% wrap with a try/catch - application carries on regardless,
```
zeeshanlakhani: @jvoegele also, this comment should go above the try i think.

src/yz_solrq.erl
```diff
@@ -0,0 +1,817 @@
...
+      all_queue_len = 0                :: non_neg_integer(),
+      queue_hwm = 1000                 :: non_neg_integer(),
+      pending_vnodes = []              :: [{pid(), atom()}],
+      drain_info = undefined           :: {pid(), reference(),
+                                           [solrq_message()]} | undefined,
+      purge_strategy                   :: purge_strategy()
+    }
+).
+
+%% TODO: Clean this up w/ proper type naming conventions.
```
zeeshanlakhani: somebody mind getting to this TODO :).

src/yz_exchange_fsm.erl
```diff
@@ -122,27 +143,41 @@ prepare_exchange(start_exchange, S) ->
...
+                {?EXCHANGE_FSM_PID, Self},
+                {?YZ_INDEX_HASHTREE_PARAMS, {YZTree, Index, IndexN}},
+                {?DRAIN_PARTITION, Index}
+            ])
+        end
+    ),
     {next_state, update_trees, S};
 
+update_trees({drain_error, Reason}, S) ->
+    lager:debug("Drain failed with reason ~p", [Reason]),
```
JeetKunDoug: This is done in the review cleanup branch.

This probably needs to be a lager:error call, not just a debug. We want to log when drains failed, especially because they will have caused the yz_solrq_drain_fsm to be killed if the timeout/cancel cycle takes too long. May want to take into account different error types (not log in_progress?) but need to make sure we can understand what's going on when something ends up in the crash log. /cc @kesslerm

src/yz_solrq.erl
fadushin: taken care of in the review branch

src/yz_fuse.erl
```diff
@@ -124,18 +124,19 @@ fuse_context() ->
...
 %% are the arguments to encode_fuse_name.
 -spec decode_fuse_name(encoded_fuse_name())
-                      -> {fuse, Namespace::fuse_namespace(), Name::fuse_name()}.
-decode_fuse_name(Name) ->
-    Fuse = base64:decode(atom_to_binary(Name, latin1)),
-    {fuse, _Namespace, _Name} = binary_to_term(Fuse).
+                      -> {Namespace::fuse_namespace(), Name::fuse_name()}.
+decode_fuse_name(Name0) ->
+    Name = atom_to_list(Name0),
+    Index = string:chr(Name, $:),
```
lehoff, line 139: If there is no : in the Name it will fail, but that seems to be okay. That shouldn't happen.

src/yz_solrq_worker.erl
```diff
@@ -351,7 +351,7 @@ handle_cast(drain_complete, #state{indexqs = IndexQs} = State) ->
     NewIndexQs = dict:fold(
         fun(Index, #indexq{queue = Queue, queue_len = QueueLen, aux_queue = AuxQueue} = IndexQ, IndexQAccum) ->
             NewIndexQ = IndexQ#indexq{
-                queue = queue:join(Queue, AuxQueue),
+                queue = queue:join(AuxQueue, Queue),
```
lehoff: Isn't the original `queue:join(Queue, AuxQueue)` the right solution?

AuxQueue accumulates while the drain is ongoing and once that is done what will be in the Queue?

If it is leftovers from the drain, then AuxQueue should go at the end. Or could it be other things?

fadushin: Actually I think you are right, @lehoff .  My original reasoning above is faulty.

AuxQueue contains the items that have been accumulated while the drain is in progress.  Queue technically should be empty, as this code is called while the worker is in the draining state, but only after all entries in all queues have been successfully drained, unless they have been cancelled due to a time out.  So technically it is possible for Queue to be non-empty, and if so, it contains data that failed to drain, and which is therefore older than the accumulated events in AuxQueue.  Nice catch.

src/yz_fuse.erl
fadushin, line 139: I agree.  It should crash and burn if there is not at least one separator.

fadushin:
Superseded by #648

PR #639, opened 2016-05-09
Post-review cleanup of Batching branch.

JeetKunDoug:
Review changes from @lehoff @JeetKunDoug @zeeshanlakhani and Fred Dushin.  Changes include:
- Renamed solrq -> solrq_worker and made solrq an API facade, and removed (most) API functions from yz_solrq_sup, allowing it to focus on its responsibilities as a supervisor
- Cleaned up the handling of drain cancellation, and added a stat to measure the number of times a drain cancel times out
- Improved testing coverage around drain cancellation
- modified handling of timer cancellation by adding a maybe_cancel_timer instead of a recursive call to handle_batch
- Some dialyzer fixes
- Removed yz_debug and moved some functionality to yz_perf, which is not part of the distribution or build
- Improved data type check


src/yz_solrq.erl
```diff
@@ -18,800 +18,241 @@
...
-
-%% @doc purge entries depending on purge strategy if we are over the HWM
-maybe_purge(Indices, PBIStrategy, State) ->
-    case over_hwm(State) of
-        true ->
-            purge(Indices, PBIStrategy, State);
-        _ ->
-            State
+-spec child_count(atom()) -> non_neg_integer().
+child_count(ChildType) ->
```
JeetKunDoug: child_count and do_child_resize won't work as-is because they are using ?MODULE, which used to be yz_solrq_sup. Thinking the best thing to do is to move the implementation of these back to yz_solrq_sup, and keep these functions here but just pass through to the supervisor.

src/yz_solrq.erl
```diff
@@ -18,800 +18,241 @@
...
+                   Type == ChildType]).
+
+-spec do_child_resize(NewSize :: pos_integer(),
+                      OldSize :: non_neg_integer(),
+                      SetTupleFun :: fun((pos_integer()) -> ok),
+                      RegnameFun :: fun((non_neg_integer()) -> regname()),
+                      ChildSpecFun :: fun((regname()) -> {regname(), pid()}))
+                      -> size_resps().
+do_child_resize(NewSize, OldSize, SetTupleFun, RegnameFun, ChildSpecFun) ->
+    case NewSize of
```
JeetKunDoug: we should move the cases into the function clauses with when-clauses.

src/yz_solrq.erl
zeeshanlakhani: good point @JeetKunDoug.

src/yz_solrq.erl
JeetKunDoug: FYI - Torben and I will fix/push the above child_count/do_child_resize and the refactor to function heads.

fadushin:
+1 9092e50

JeetKunDoug:
@borshop merge

PR #644, opened 2016-05-23
Fix deadlock between yz_entropy_mgr and yz_index_hashtree

JeetKunDoug:
- Add timeout to yz_entropy_mgr:get_lock
- If lock acquisition times out, added yz_entropy_mgr:release_lock, which casts a release message back to make sure we don't hold the lock after the timeout.
- Make yz_index_hashtree:expire a cast rather than call, as that's the way it's really used (and only for testing/diagnostics).
- Removing ill-advised code that attempted to synchronize yz_events with yz_index_hashtree - test had bad configuration options so it would fail, but the code itself handles the situation properly.

### DIFF
--- a/include/yokozuna.hrl
+++ b/include/yokozuna.hrl
@@ -258,6 +258,9 @@
             app_helper:get_env(riak_kv, anti_entropy_build_limit, {1, 3600000})
         )).
 
+-define(YZ_ENTROPY_LOCK_TIMEOUT,
+    app_helper:get_env(?YZ_APP_NAME, anti_entropy_lock_timeout, 10000)).
+
 -type hashtree() :: hashtree:hashtree().
 -type exchange() :: {p(), {p(), n()}}.
 -type exchange_mode() :: automatic | manual.
--- a/src/yz_index_hashtree.erl
+++ b/src/yz_index_hashtree.erl
@@ -109,7 +109,7 @@ clear(Tree) ->
 
 %% @doc Expire the tree.
 expire(Tree) ->
-    gen_server:call(Tree, expire, infinity).
+    gen_server:cast(Tree, expire).
 
 %% @doc Terminate the `Tree'.
 stop(Tree) ->
@@ -208,10 +208,6 @@ handle_call(destroy, _From, S) ->
     S2 = destroy_trees(S),
     {stop, normal, ok, S2};
 
-handle_call(expire, _From, S) ->
-    S2 = S#state{expired=true},
-    {reply, ok, S2};
-
 handle_call(_Request, _From, S) ->
     Reply = ok,
     {reply, Reply, S}.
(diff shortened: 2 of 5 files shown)


include/yokozuna.hrl
```diff
@@ -258,6 +258,9 @@
             app_helper:get_env(riak_kv, anti_entropy_build_limit, {1, 3600000})
         )).
 
+-define(YZ_ENTROPY_LOCK_TIMEOUT,
+    app_helper:get_env(?YZ_APP_NAME, anti_entropy_lock_timeout, 10000)).
```
zeeshanlakhani: @fadushin and I have liked using macros in defaults like this. I know it's not *that* important, but I think it's easier to look for it going forward.

src/yz_index_hashtree.erl
```diff
@@ -587,9 +588,8 @@ build_or_rehash(Tree, Locked, Type, #state{index=Index, trees=Trees}) ->
         {true, build} ->
             lager:debug("Starting YZ AAE tree build: ~p", [Index]),
             Indexes = yz_index:get_indexes_from_meta(),
-            InSync = yz_events:in_sync_with_metadata(),
```
zeeshanlakhani, line 590: Of note in our _big_ rebase, we should remove the old commit et. al, related to metadata sync.

src/yz_index_hashtree.erl
```diff
@@ -639,7 +640,15 @@ get_all_locks(Type, Pid) ->
     %% file page cache but the bg manager stuff is kind of convoluted
     %% and there isn't time to figure this all out for 2.0. Thus,
     %% Yokozuna will not bother with the Solr lock for now.
-    ok == yz_entropy_mgr:get_lock(Type, Pid).
+    try yz_entropy_mgr:get_lock(Type, Pid) of
+        ok -> true;
+        Other ->
+            lager:debug("Could not get lock: ~p", [Other]),
+            false
+    catch exit:{timeout,_} ->
```
zeeshanlakhani, line 648: I wish we have a more "common" abstraction for this specific exit, since it's possible to catch this for any handle_call w/ a timeout; since it's also caught above.

zeeshanlakhani:
I think this should be rebased into 2 commits (cast + config)? @JeetKunDoug.

zeeshanlakhani:
Realized that you, @JeetKunDoug, had the macro in the hrl already, coolio.

zeeshanlakhani:
+1 415c3bc

JeetKunDoug:
@borshop merge

src/yz_index_hashtree.erl
lehoff, line 648: Would it be so bad if we just had the yz_index_hashtree process die when it cannot get the lock?
The hashtree would be persisted to disk on termination.
That would allow the entropy_mgr to do its clean-up of held locks.

PR #645, opened 2016-05-23
Fix errors in yokozuna shutdown sequence

jvoegele:
- Call `riak_core_node_watcher:service_down(yokozuna)` in `prep_stop` so that the node that is going down is not included in coverage plans.
- Handle timeout error when attempting to disable Yokozuna components during shutdown. When this error is not handled index operations continue to queue in the node that is going down, eventually leading to errors when they are sent off to Solr.
- Test that these changes have the desired effect in `yz_startup_shutdown` test.

### DIFF
--- a/riak_test/yz_startup_shutdown.erl
+++ b/riak_test/yz_startup_shutdown.erl
@@ -5,6 +5,7 @@
 -include_lib("eunit/include/eunit.hrl").
 -compile({parse_transform, rt_intercept_pt}).
 
+-define(CLUSTER_SIZE, 4).
 -define(MAX_SESSIONS, 11).
 -define(MAX_PIPELINE_SIZE, 9).
 -define(CONFIG,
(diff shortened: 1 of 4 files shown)


riak_test/yz_startup_shutdown.erl
```diff
@@ -167,3 +177,101 @@ verify_ibrowse_config([Node1|_] = Cluster) ->
...
+
+-define(QUERY_FAILURE_MESSAGE, "IOException occured when talking to server").
+%% @doc Assert that there are no log messages on any node in `Cluster' that
+%% indicate failed queries. Failed queries are a symptom of yz_app not
+%% properly notifying the riak_core_node_watcher that it is shutting down.
+assert_no_query_failures(Cluster) ->
+    [assert_not_in_logs(Node, ?QUERY_FAILURE_MESSAGE) || Node <- Cluster].
+
+start_background_processes(HP, Bucket, Index) ->
+    QueryPid = repeat_until_stop(query_fun(HP, Index, "*", "*"), [{delay, 300}]),
```
jvoegele: Add comment explaining why we use a delay of 300

JeetKunDoug:
+1 2617101

JeetKunDoug:
+1 7845f67

zeeshanlakhani:
@borshop merge

PR #646, opened 2016-05-24
Bring sidejob back for stats

JeetKunDoug:
Writing to Exometer can slow down the batching to Solr. Rather than do this synchronously while batching information to Solr, use sidejob to make the exometer updates asynchronous. Also, build a simple fuse_stats_plugin implementation so that Fuse can also use Sidejob for stats updates.

Requires basho/sidejob#14 in order to pass tests.

### DIFF
--- a/src/yz_fuse_stats_sidejob.erl
+++ b/src/yz_fuse_stats_sidejob.erl
@@ -1,0 +1,40 @@
+%% -------------------------------------------------------------------
+%% Copyright (c) 2016 Basho Technologies, Inc.  All Rights Reserved.
+%%
+%% This file is provided to you under the Apache License,
+%% Version 2.0 (the "License"); you may not use this file
+%% except in compliance with the License.  You may obtain
+%% a copy of the License at
+%%
+%%   http://www.apache.org/licenses/LICENSE-2.0
+%%
+%% Unless required by applicable law or agreed to in writing,
+%% software distributed under the License is distributed on an
+%% "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+%% KIND, either express or implied.  See the License for the
+%% specific language governing permissions and limitations
+%% under the License.
+%%
+%% -------------------------------------------------------------------
+%%
+%% @doc yz_fuse_stats_sidejob is a module implementing for offloading
+%%      statistics updates to a background worker pool maintained by sidejob
+%%      so that the calls to exometer don't slow down indexing work.
+%%      All operations are pass-through to yz_stat to keep the "how" of stats
+%%      updates in one place.
+
+-module(yz_fuse_stats_sidejob).
+-behaviour(fuse_stats_plugin).
+-export([init/1, increment/2]).
+
+%% @doc Initialize exometer for `Name'.
+-spec init(Name :: atom()) -> ok.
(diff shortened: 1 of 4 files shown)


JeetKunDoug:
@borshop retry

src/yz_fuse_stats_sidejob.erl
```diff
@@ -0,0 +1,40 @@
...
+%%      updates in one place.
+
+-module(yz_fuse_stats_sidejob).
+-behaviour(fuse_stats_plugin).
+-export([init/1, increment/2]).
+
+%% @doc Initialize exometer for `Name'.
+-spec init(Name :: atom()) -> ok.
+init(Name) ->
+    yz_stat:initialize_fuse_stats(Name),
```
fadushin, line 33: I don't have a particularly strong feeling about this, but if we want to "contain" callouts to the fuse library in one place, should move the initialize_fuse_stats function to yz_fuse, instead of it being in yz_stat?

src/yz_stat.erl
```diff
@@ -84,6 +82,11 @@ register_stats() ->
 get_stats() ->
     riak_core_stat:get_stats(?APP).
 
+%% @doc Initialize the Fuse stats subsystem for a given fuse.
+-spec initialize_fuse_stats(Name :: atom()) -> ok.
+initialize_fuse_stats(Name) ->
+    fuse_stats_exometer:init(Name).
+
```
fadushin, line 89: Ditto comment above about moving fuse calls to yz_fuse.

src/yz_fuse_stats_sidejob.erl
zeeshanlakhani, line 33: @fadushin i think that's a fair point.

JeetKunDoug, line 33: So I thought about that, but it was "1/2 stats, 1/2 fuse" and since this is really about _stats_ it felt like it belonged in the yz_stats module, not yz_fuse. Would be OK if I'm outvoted, but that's why I put them in stats and not yz_fuse.

zeeshanlakhani, line 33: I had similar internal debates @JeetKunDoug. i mean, if we eventually build a generic interface for fuse... then we'd want it inside a fuse module for any app needing fuse stats?

fadushin, line 33: Yeah my only interest was in containing use of fuse_\* to a single module.  6 of 1.

JeetKunDoug, line 33: So if we want to do that, we'll need to have a separate yz_stat_worker:update clause that calls into yz_fuse rather than routing everything through yz_stats, which also seemed like more cognitive dissonance than I wanted... So, if we have 6 of 1 I'm going to leave it the way it is. LMK if there's a strong enough opinion the other way and we can discuss via a higher-bandwidth communications medium.

JeetKunDoug:
Will revisit moving fuse stats stuff to its own module if/when we have more subsystems using fuse.

zeeshanlakhani:
+1 4a766b3

JeetKunDoug:
@borshop merge

JeetKunDoug:
@borshop merge

PR #650, opened 2016-06-09
Mark new Cuttlefish elements as commented

jvoegele:
New search.queue.\* configuration elements were added to the Yokozuna Cuttlefish schema. Although these new config elements had default values associated with them, they would still cause certain tests to fail. Specifically, tests that performed a "downgrade" operation would fail because the downgraded node would still have the config file with the new elements and would therefore fail to start due to validation errors. Therefore, we need to mark the new config elements as commented in Cuttlefish so that tests that perform a downgrade can function properly.


fadushin:
+1 c50320b

fadushin:
@borshop merge

PR #651, opened 2016-06-14
Added handling of JSON encoding errors

fadushin:
(e.g., if operations contain non-unicode binary data) and treating such bad data in the same way we handle bad requests on the Solr side, in other words, drop the message, and keep chugging along on the rest of the messages, and make sure the object associated with the bad data makes its way into the YZ AAE tree.  Also changed debug logs for failures to error, so that the customer will actually see the error, and added code to handle the case where we try to purge an entry with an empty queue.

A test for this condition has been added to riak_test.  See https://github.com/basho/riak_test/pull/1092


jvoegele:
+1 10a35c4fd9de68c41908f919896ac2e81f7d4bb9

fadushin:
@borshop merge

PR #653, opened 2016-06-17
safe yz event tick

fadushin:
Wrapped the tick handling callback in a try-catch in order to prevent crashes if Solr is unavailable.


JeetKunDoug:
+1 b6a4f9d

PR #654, opened 2016-06-17
Fixed test failures due to setting default draining to false

fadushin:
Added draining to tests that need it and fixed yz_rt to wait for an index to be created by calling yz_index:exists instead of (just) yz_solr:ping


JeetKunDoug:
+1 d72951a

PR #655, opened 2016-06-24
WIP - Use `riak_object:hash` instead of duplicating the logic

JeetKunDoug:
DO NOT MERGE - for discussion/review.

### DIFF
--- a/src/yz_kv.erl
+++ b/src/yz_kv.erl
@@ -89,9 +89,7 @@ get(C, Bucket, Key) ->
 %% @doc calculates the hash of a riak object, returns binary
 -spec hash_object(riak_object:riak_object()) -> binary().
 hash_object(Obj) ->
-    Vclock = riak_object:vclock(Obj),
-    Obj2 = riak_object:set_vclock(Obj, lists:sort(Vclock)),
-    Hash = erlang:phash2(term_to_binary(Obj2)),
+    Hash = riak_object:hash(Obj),
     term_to_binary(Hash).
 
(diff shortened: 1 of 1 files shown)


src/yz_kv.erl
```diff
@@ -89,9 +89,7 @@ get(C, Bucket, Key) ->
 %% @doc calculates the hash of a riak object, returns binary
 -spec hash_object(riak_object:riak_object()) -> binary().
 hash_object(Obj) ->
-    Vclock = riak_object:vclock(Obj),
-    Obj2 = riak_object:set_vclock(Obj, lists:sort(Vclock)),
-    Hash = erlang:phash2(term_to_binary(Obj2)),
+    Hash = riak_object:hash(Obj),
     term_to_binary(Hash).
```
JeetKunDoug: Just a reminder to drop the term_to_binary here if we move it into riak_object:hash as discussed in KV.

bsparrow435:
This is now ready for merge and depends on https://github.com/basho/riak_kv/pull/1473

nickelization:
Nice, I'm glad we're de-duplicating this code. To go one step further, I'm wondering if we even need to have a `yz_kv:hash_object` function at all, or if maybe that's an unnecessary layer of indirection. Should we update the few places where we call `yz_kv:hash_object` to just call `riak_object:hash` directly?

nickelization:
+1 b29e85b

nickelization:
@borshop merge

PR #656, opened 2016-07-07
Merge 2.0 into develop 2.2

fadushin:
This PR includes a large number of merges from 2.0.7 into the develop-2.2 branch.  While there were a relatively large number of merge conflicts in this merge, most of them were inconsequential.  However, we should take a closer look at the interface between riak_kv_vnode and yz_kv, as that interface changed slightly in 2.1 to accommodate indexing in the write_once path (where riak_objects are still encoded as binaries).  We should check to verify that performance is not seriously degraded as part of this merge.


jvoegele:
+1 b655675fbfdeadf9a244b778f787c2417f2f135b

fadushin:
@borshop merge

PR #657, opened 2016-07-12
add gset support

binarytemple-external:
@russelldb another gset pull request


zeeshanlakhani:
@binarytemple-bet365 maybe it wasn't mentioned in the HC, but I don't think this will hit the 2.0 series _officially_, but I'll let @JeetKunDoug, @fadushin refer to which branch it should hit off before we review.

binarytemple-external:
Thanks @zeeshanlakhani

PR #660, opened 2016-07-18
depend on fuse to unbreak make devrel

hmmr:
Because of https://github.com/basho/riak/pull/812, `make devrel` now needs fuse, which it cannot find as no deps in riak TS seems to depend on fuse.

It would, in theory, be more appropriate to rebase riak_ts-develop onto whatever is current in yokozuna, but because yokozuna is not used in riak TS, we should be good with this surgical change.

Let @javajolt the branch master decide on the fate of this PR (or do a rebase/merge).


hazen:
:+1: d86142e

hmmr:
@borshop merge

hazen:
@borshop merge

PR #662, opened 2016-07-22
Add throttling to AAE repairs

jvoegele:
https://bashoeng.atlassian.net/browse/RIAK-2626

As per RIAK-2626, we needed throttling of Yokozuna's AAE repairs to prevent overload due to the large number of key exchanges that will result from the new hashing algorithm. Therefore, we've added configurable throttling behavior Yokozuna's AAE subsystem, using the new `riak_core_throttle` module introduced in https://github.com/basho/riak_core/pull/849 .

This throttling behavior will introduce a delay between AAE repair operations, the length of which is determined by the current level of "load" in Yokozuna. The load is defined as the current value of the `yz_solrq:queue_total_length/0` function. The mapping between the level of load and the corresponding time delay (referred to as "limits") is configurable with the `yokozuna.aae_throttle_limits` configuration element defined in `yokozuna.schema`.

### DIFF
--- a/include/yokozuna.hrl
+++ b/include/yokozuna.hrl
@@ -263,11 +263,11 @@
 -define(YZ_ENTROPY_LOCK_TIMEOUT,
     app_helper:get_env(?YZ_APP_NAME, anti_entropy_lock_timeout, 10000)).
 
--define(YZ_ENTROPY_THROTTLE_KEY, yokozuna_aae).
+-define(YZ_ENTROPY_THROTTLE_KEY, aae_throttle).
 -define(YZ_ENTROPY_THROTTLE_LIMITS_KEY, aae_throttle_limits).
 -define(YZ_ENTROPY_THROTTLE_DEFAULT_LIMITS,
         [{-1,0}, {500,10}, {1000,50}, {5000,250}, {10000,1000}, {50000,3000}]).
--define(YZ_ENTROPY_THROTTLE_KILL_KEY, yokozuna_aae_throttle_kill_switch).
+-define(YZ_ENTROPY_THROTTLE_ENABLED_KEY, aae_throttle_enabled).
 
 -type hashtree() :: hashtree:hashtree().
 -type exchange() :: {p(), {p(), n()}}.
--- a/include/yokozuna.hrl
+++ b/include/yokozuna.hrl
@@ -263,6 +263,10 @@
 -define(YZ_ENTROPY_LOCK_TIMEOUT,
     app_helper:get_env(?YZ_APP_NAME, anti_entropy_lock_timeout, 10000)).
 
+-define(YZ_ENTROPY_THROTTLE_KEY, yokozuna_aae).
+-define(YZ_ENTROPY_THROTTLE_DEFAULT_LIMITS,
+        [{-1,0}, {500,10}, {1000,50}, {5000,250}, {10000,1000}, {50000,3000}]).
+
 -type hashtree() :: hashtree:hashtree().
 -type exchange() :: {p(), {p(), n()}}.
 -type exchange_mode() :: automatic | manual.
--- a/include/yokozuna.hrl
+++ b/include/yokozuna.hrl
@@ -264,8 +264,10 @@
     app_helper:get_env(?YZ_APP_NAME, anti_entropy_lock_timeout, 10000)).
 
 -define(YZ_ENTROPY_THROTTLE_KEY, yokozuna_aae).
+-define(YZ_ENTROPY_THROTTLE_LIMITS_KEY, yokozuna_aae_throttle_limits).
 -define(YZ_ENTROPY_THROTTLE_DEFAULT_LIMITS,
         [{-1,0}, {500,10}, {1000,50}, {5000,250}, {10000,1000}, {50000,3000}]).
+-define(YZ_ENTROPY_THROTTLE_KILL_KEY, yokozuna_aae_throttle_kill_switch).
 
 -type hashtree() :: hashtree:hashtree().
 -type exchange() :: {p(), {p(), n()}}.
--- a/include/yokozuna.hrl
+++ b/include/yokozuna.hrl
@@ -264,7 +264,7 @@
     app_helper:get_env(?YZ_APP_NAME, anti_entropy_lock_timeout, 10000)).
 
 -define(YZ_ENTROPY_THROTTLE_KEY, yokozuna_aae).
--define(YZ_ENTROPY_THROTTLE_LIMITS_KEY, yokozuna_aae_throttle_limits).
+-define(YZ_ENTROPY_THROTTLE_LIMITS_KEY, aae_throttle_limits).
 -define(YZ_ENTROPY_THROTTLE_DEFAULT_LIMITS,
         [{-1,0}, {500,10}, {1000,50}, {5000,250}, {10000,1000}, {50000,3000}]).
 -define(YZ_ENTROPY_THROTTLE_KILL_KEY, yokozuna_aae_throttle_kill_switch).
--- a/priv/yokozuna.schema
+++ b/priv/yokozuna.schema
@@ -104,7 +104,7 @@
 }.
 
 %% @doc Whether the throttle for Yokozuna active anti-entropy is enabled.
-{mapping, "anti_entropy.throttle", "yokozuna.aae_throttle_kill_switch", [
+{mapping, "search.anti_entropy.throttle", "yokozuna.aae_throttle_kill_switch", [
   {default, on},
   {datatype, {flag, off, on}},
   hidden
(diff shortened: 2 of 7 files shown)


jvoegele:
The build currently fails because it depends on changes in https://github.com/basho/riak_core/pull/849 that are not visible here.

src/yz_entropy_mgr.erl
```diff
@@ -502,6 +503,7 @@ tick(S) ->
     S3 = lists:foldl(fun(_,SAcc) ->
                              maybe_poke_tree(SAcc)
                      end, S2, lists:seq(1,10)),
+    update_throttle(S3),
```
bsparrow435, line 508: I'm not understanding why this line is necessary. Why is the throttle being updated here?

jvoegele, line 508: You can't tell just by looking at the diff here, but that call to `update_throttle` is actually part of the `tick` function that is called periodically (I think it's once every second). So basically we update the throttle delay time based on the current level of load, and this happens on every timer tick.

bsparrow435, line 508: I didnt even see it wasn't in the init. That makes sense, thought you were updating it immediately after the init_throttle. Makes sense.

bsparrow435:
Code looks good.

src/yz_entropy_mgr.erl
```diff
@@ -515,6 +517,35 @@ maybe_poke_tree(S) ->
...
+        true ->
+            Load = calculate_current_load(State),
+            riak_core_throttle:set_throttle_by_load(?YZ_ENTROPY_THROTTLE_KEY,
+                                                    Load);
+        false ->
+            ok
+    end.
+
+calculate_current_load(_State) ->
+    yz_solrq:queue_total_length().
```
JeetKunDoug: As discussed, let's see if we can use the stat here rather than calling queue_total_length, which will end up doing a gen_server:call to each solrq. See`riak_kv_entropy_manager:get_max_local_vnodeq` for an example. Additionally, and outside this PR, we should probably make queue_total_length use riak_core_util:pmap or something similar so we don't have to wait for each q_worker to respond before calling the next one.

include/yokozuna.hrl
```diff
@@ -263,6 +263,12 @@
 -define(YZ_ENTROPY_LOCK_TIMEOUT,
     app_helper:get_env(?YZ_APP_NAME, anti_entropy_lock_timeout, 10000)).
 
+-define(YZ_ENTROPY_THROTTLE_KEY, yokozuna_aae).
```
JeetKunDoug: With the addition of AppName to the riak_core_throttle APIs, we can change this to `aae_enabled` since it'll be prefixed with `yokozuna` as an application name rather than having to be unique.

include/yokozuna.hrl
```diff
@@ -263,6 +263,12 @@
...
+-define(YZ_ENTROPY_THROTTLE_LIMITS_KEY, yokozuna_aae_throttle_limits).
```
JeetKunDoug: Same as above (make this just `aae_throttle_limits`)

include/yokozuna.hrl
```diff
@@ -263,6 +263,12 @@
...
+-define(YZ_ENTROPY_THROTTLE_DEFAULT_LIMITS,
+        [{-1,0}, {500,10}, {1000,50}, {5000,250}, {10000,1000}, {50000,3000}]).
+-define(YZ_ENTROPY_THROTTLE_KILL_KEY, yokozuna_aae_throttle_kill_switch).
```
JeetKunDoug: `KILL` becomes `ENABLED` like in core/kv PRs.

priv/yokozuna.schema
```diff
@@ -103,6 +103,46 @@
   end
 }.
 
+%% @doc Whether the throttle for Yokozuna active anti-entropy is enabled.
+{mapping, "search.anti_entropy.throttle", "yokozuna.aae_throttle_kill_switch", [
```
JeetKunDoug: kill_switch -> enabled.

priv/yokozuna.schema
```diff
@@ -103,6 +103,46 @@
...
+  {default, on},
+  {datatype, {flag, off, on}},
```
JeetKunDoug: Need to switch the On/Off defaults when we switch the `kill_switch` to `enabled` - I'm assuming we're mapping `search.anti_entropy.throttle" value of`on`to`yokozuna.aae_throttle_kill_switch`of`false`by the order of the`{flag, off, on}` tuple above, but not sure. Please confirm and make necessary changes.

src/yz_entropy_mgr.erl
```diff
@@ -515,6 +517,35 @@ maybe_poke_tree(S) ->
...
+    riak_core_throttle:init(?YZ_ENTROPY_THROTTLE_KEY,
+                            ?YZ_APP_NAME,
+                            {?YZ_ENTROPY_THROTTLE_LIMITS_KEY,
+                             ?YZ_ENTROPY_THROTTLE_DEFAULT_LIMITS},
+                            {?YZ_ENTROPY_THROTTLE_KILL_KEY, false}),
+    ok = riak_core_throttle:set_throttle(?YZ_ENTROPY_THROTTLE_KEY,
+                                         InitialThrottle).
+
+update_throttle(State) ->
+    case riak_core_throttle:is_throttle_enabled(?YZ_ENTROPY_THROTTLE_KEY) of
```
JeetKunDoug: Refactor to function heads in a `maybe_update_throttle` function?

test/yokozuna_schema_tests.erl
```diff
@@ -50,6 +50,8 @@ basic_schema_test() ->
     cuttlefish_unit:assert_config(Config, "yokozuna.solrq_drain_enable", false),
     cuttlefish_unit:assert_config(Config, "yokozuna.ibrowse_max_sessions", 100),
     cuttlefish_unit:assert_config(Config, "yokozuna.ibrowse_max_pipeline_size", 1),
+    cuttlefish_unit:assert_config(Config, "yokozuna.aae_throttle_kill_switch", false),
```
JeetKunDoug: `kill_switch` -> `enabled`, switch default, etc.

test/yokozuna_schema_tests.erl
```diff
@@ -127,6 +134,10 @@ override_schema_test() ->
     cuttlefish_unit:assert_config(Config, "yokozuna.solrq_drain_enable", false),
     cuttlefish_unit:assert_config(Config, "yokozuna.ibrowse_max_sessions", 101),
     cuttlefish_unit:assert_config(Config, "yokozuna.ibrowse_max_pipeline_size", 11),
+    cuttlefish_unit:assert_config(Config, "yokozuna.aae_throttle_kill_switch",
```
JeetKunDoug: `kill_switch` -> `enabled`, switch default, etc.

src/yz_entropy_mgr.erl
```diff
@@ -547,7 +549,14 @@ update_throttle(State) ->
     end.
 
 calculate_current_load(_State) ->
-    yz_solrq:queue_total_length().
+    case yz_stat:get_stat([queue, total_length]) of
+        Num when is_integer(Num) ->
+            Num;
+        Unexpected ->
+            lager:error("Unexpected value for statistic [queue, total_length]: ~p",
```
JeetKunDoug: Not sure how frequently this is happening, but my guess would be only before the stat is initialized? If so, I'd make this a lager:debug since the end-user isn't going to care (I think).

include/yokozuna.hrl
```diff
@@ -263,6 +263,12 @@
 -define(YZ_ENTROPY_LOCK_TIMEOUT,
     app_helper:get_env(?YZ_APP_NAME, anti_entropy_lock_timeout, 10000)).
 
+-define(YZ_ENTROPY_THROTTLE_KEY, aae_throttle).
+-define(YZ_ENTROPY_THROTTLE_LIMITS_KEY, aae_throttle_limits).
+-define(YZ_ENTROPY_THROTTLE_DEFAULT_LIMITS,
+        [{-1,0}, {500,10}, {1000,50}, {5000,250}, {10000,1000}, {50000,3000}]).
```
jvoegele, line 269: I want to point out that these numbers are complete guesses. We need to do some testing and tuning to make sure that they are effective.

src/yz_stat.erl
```diff
@@ -82,6 +83,18 @@ register_stats() ->
 get_stats() ->
     riak_core_stat:get_stats(?APP).
 
+%% @doc Return the value for the stat with the given `Name', or `undefined' if
+%% not found.
+-spec get_stat([atom(), ...]) -> undefined | term().
+get_stat(Name) when is_list(Name) ->
+    Path = [riak_core_stat:prefix(), ?APP] ++ Name,
```
JeetKunDoug: Use ?PFX macro here?

JeetKunDoug:
+1 bec3557

jvoegele:
@borshop merge

PR #668, opened 2016-08-05
Solr 4.10 upgrade/downgrade

fadushin:
Added support for upgrade to Solr 4.10, along with downgrade back to prior versions of Riak that only support 4.7.

This PR also adds an option to remove a node from a cover plan, so that the node can continue indexing (or reindexing data) while queries can still function across the rest of the cluster.

This PR includes cuttlefish config for disabling a node from cover plans, as well as a command-line option for enabling or disabling the node from cover plans on the fly.

The upgrade/downgrade procedure more or less follows the procedure described at:

https://github.com/basho/rfc/tree/fd-solr-4.10-upgrade-downgrade

The newly added yz_solr_upgrade_downgrade test requires changes to riak_test covered under:

https://github.com/basho/riak_test/pull/1120

In order to build a version of Riak or Riak EE that supports the riak-admin search dist-query option,  use:

https://github.com/basho/riak/pull/867
https://github.com/basho/riak_ee/pull/404


src/yz_solr_proc.erl
```diff
@@ -309,3 +330,10 @@ solr_jvm_opts() ->
     app_helper:get_env(?YZ_APP_NAME,
                        solr_jvm_opts,
                        ?YZ_DEFAULT_SOLR_JVM_OPTS).
+
+-spec solr_is_up(#state{}) -> #state{}.
```
nickelization: Incorrect return type in the spec here.

nickelization:
Added some comments on some of this code last week, but looks like they've been lost due to a rebase. Will go through and copy ones back in if/where they are relevant.

include/yokozuna.hrl
```diff
@@ -263,6 +263,9 @@
 -define(YZ_ENTROPY_LOCK_TIMEOUT,
     app_helper:get_env(?YZ_APP_NAME, anti_entropy_lock_timeout, 10000)).
 
+-define(YZ_ENABLE_DIST_QUERY,
```
nickelization, line 266: I know this pattern is already being used in other places in Yokozuna, but I don't like it. Defines without any arguments like this are normally used for constants, so when I see places where something like this is used, I wouldn't normally expect it to expand out to a function call that dynamically fetches the current config.

Plus we only use it in two places anyway, so it feels like we might as well just inline the app_helper:get_env call directly in the code.

riak_test/yz_rt.erl
```diff
@@ -291,6 +293,37 @@ load_data(Cluster, Bucket, YZBenchDir, NumKeys, Operations) ->
...
     load_data(Cluster, Bucket, YZBenchDir, NumKeys, [{load_fruit, 1}]).
 
+
+-spec write_data(cluster(), bucket(), non_neg_integer(), non_neg_integer()) ->
+    [ok | {error, term()}].
+write_data(Cluster, Bucket, Start, Count) ->
+    ValueGenerator = fun(Key) -> {Key, "text/plain"} end,
+    write_data(Cluster, Bucket, Start, Count, ValueGenerator).
+
+-spec write_data(cluster(), bucket(), non_neg_integer(), non_neg_integer(), fun((non_neg_integer()) -> ok | term()))  ->
```
nickelization, line 303: I wonder if this should just go in `rt` instead of `rt_yz`? Seems like it's not necessarily specific to Yokozuna, and might be useful in other tests as well.

riak_test/yz_rt.erl
```diff
@@ -795,18 +838,21 @@ load_intercept_code(Node) ->
                               "*.erl"]),
     rt_intercept:load_code(Node, [CodePath]).
 
--spec rolling_upgrade(cluster() | node(),
+rolling_upgrade(Cluster, Version, UpgradeConfig, WaitForServices) ->
+    rolling_upgrade(Cluster, Version, UpgradeConfig, WaitForServices, fun rt:no_op/1).
+
+        -spec rolling_upgrade(cluster() | node(),
```
nickelization: ^ Indentation ^

riak_test/yz_solr_upgrade_downgrade.erl
```diff
@@ -0,0 +1,459 @@
...
+    ]},
+    {yokozuna, [
+        {enabled, true},
+        {anti_entropy_tick, 500},
+        {anti_entropy_build_limit, {100, 1000}},
+        {anti_entropy_concurrency, 8}
+    ]}
+]).
+
+%% these indices is never upgraded
```
nickelization: %s/is/are/

riak_test/yz_solr_upgrade_downgrade.erl
```diff
@@ -0,0 +1,459 @@
...
+%% - index-a1 is not upgraded and is not written to after upgrade
+%% - index-a1 is not upgraded and is written to after upgrade
```
nickelization: I think this second comment should say index-a2, right?

riak_test/yz_solr_upgrade_downgrade.erl
```diff
@@ -0,0 +1,459 @@
...
+-define(INDEX_B3, <<"index-b3">>).
+-define(BUCKET_TYPE_B3, <<"bucket-type-b3">>).
+-define(BUCKET_B3, {?BUCKET_TYPE_B3, <<"bucket-name-b3">>}).
+
+%% index-c is created after the upgrade
+-define(INDEX_C, <<"index-c">>).
+-define(BUCKET_TYPE_C, <<"bucket-type-c">>).
+-define(BUCKET_C, {?BUCKET_TYPE_C, <<"bucket-name-c">>}).
+
+-define(LUCENE_MATCH_VERSION_OPEN, "<luceneMatchVersion>").
```
nickelization: LUCENE_MATCH_VERSION_(OPEN|CLOSED) don't appear to be used anymore. Fred mentioned to me in Friday that these are indeed leftover from a previous version of the code, and can be removed.

riak_test/yz_solr_upgrade_downgrade.erl
```diff
@@ -0,0 +1,459 @@
...
+%% data in order for the downgraded cluster to behave properly (except for
+%% index-a, which is not touched as part of the upgrade).
+%%
+confirm() ->
+    %%
+    %% Build a "previous" cluster using the configuration defined above
+    %%
+    Cluster = rt:build_cluster(lists:duplicate(
+        ?CLUSTER_SIZE,
+        {previous, ?CONFIG}
```
nickelization: Instead of using "current" and "previous", I think we would be better off explicitly specifying the Riak versions we want to test. Otherwise I assume this test will break when current and previous start referring to newer code.

riak_test/yz_solr_upgrade_downgrade.erl
```diff
@@ -0,0 +1,459 @@
...
+
+verify_data(Cluster, Index, NumKeys) ->
+    ok = yz_rt:verify_num_match(yokozuna, Cluster, Index, NumKeys),
+    ExpectedSolrMatch = (?N_VAL * NumKeys * length(Cluster)) div ?CLUSTER_SIZE,
+    ok = yz_rt:verify_num_match(solr, Cluster, Index, ExpectedSolrMatch),
+    ok.
+
+augment_config(ApplicationKey, Key, Value, Config) ->
+    ApplicationConfig = proplists:get_value(ApplicationKey, Config),
+    ApplicationDict = dict:from_list(ApplicationConfig),
```
nickelization: Why convert things to and from dict here instead of just working with the proplist directly?

riak_test/yz_solr_upgrade_downgrade.erl
```diff
@@ -0,0 +1,459 @@
...
+    lager:info("Moving yz data directory for index ~s out of the way...", [IndexName]),
+    YZTempPath = io_lib:format("~s/yz/~s/data", [DataDir, IndexName]),
+    ok = file:rename(YZTempPath, io_lib:format("~s-~s", [YZTempPath, TimestampStr])).
+
+mv_yz_aae_dir(DataDir, TimestampStr) ->
+    lager:info("Moving yz AAE directory out of the way..."),
+    YZTempPath = io_lib:format("~s/yz_anti_entropy", [DataDir]),
+    ok = file:rename(YZTempPath, io_lib:format("~s-~s", [YZTempPath, TimestampStr])).
+
+timestamp_str() ->
```
nickelization, line 412: Maybe put this in the `rt` module, in case someone else wants to use it for another test in the future?

priv/yokozuna.schema
```diff
@@ -338,6 +338,19 @@
...
   {validators, ["positive_integer"]}]}.
 
+%% @doc Enable or disable this node in distributed query plans.  If enabled,
+%% this node will participate in distributed Solr queries.  If disabled, the
+%% node will be excluded from yokozuna cover plans, and will therefore never
+%% be consulted in a distributed query.  Note that this node may still be used
+%% to execute a query.  Use this flag if you have a long running administrative
+%% operation (e.g., reindexing) which requires that the node be removed from
+%% query plans, and which would otherwise result in inconsistent search results.
+{mapping, "search.dist_query.enable", "yokozuna.enable_dist_query", [
```
nickelization: I feel like this should just be "search.dist_query", not "search.dist_query.enable"; the extra ".enable" seems redundant, given that it's already a flag that we're setting to "on"/"off". We have many other flags in Riak that follow the format of simply saying "X = on" instead of "X.enable = on" and I think we should stick with that convention.

jvoegele:
Nick has made some good points in his review notes. I don't have anything further to add, so I'll give a provisional `+1` on the presumption that Nick's comments will be addressed as appropriate.

priv/yokozuna.schema
nickelization: Actually, even better would be to use `{datatype, {flag, enabled, disabled}}`. That way it's super clear what's happening in the config, e.g.:

`search.dist_query = enabled`

riak_test/yz_rt.erl
fadushin, line 303: rt already has a version of generating data, and this one has specific support for mime types.  We could, but I don't think it's necessary.

riak_test/yz_solr_upgrade_downgrade.erl
fadushin: Well, we have that problem, generally, yes.  But I also don't want to tie us to a specific version, as the builders, for example, may not have the latest 2.0 or 2.1 devrels.

riak_test/yz_solr_upgrade_downgrade.erl
fadushin, line 412: Perhaps.  I tend to move things to that module if I know I have >1 use of it.

priv/yokozuna.schema
fadushin: Does that preclude 'on' and 'off' or 'true' and 'false'?  I thinks CSEs prefer that flexibility.  But I agree we should remove the .enable

riak_test/yz_rt.erl
nickelization, line 303: Ah okay, gotcha. Makes sense.

riak_test/yz_solr_upgrade_downgrade.erl
nickelization: Hmm okay, I guess we're kind of stuck either way. I know in general that can be a problem, but I figured this test might be a special case, since we're specifically covering the 4.7/4.10.4 upgrade, which means we'll specifically have to be upgrading from pre-2.2 to >= 2.2. I guess there isn't really a good solution though, so we may as well leave it like this.

nickelization:
+1 625d324

JeetKunDoug:
@borshop merge

PR #670, opened 2016-08-12
Replace dist_query CLI command with config change callbacks

nickelization:
Rather than adding a custom command for enabling/disabling dist-query on a node, we can simply use the builtin clique "set" command to directly update the setting, and we can use config callbacks to ensure the new setting takes effect immediately on the live node.

Note: these changes require the latest version of clique from the develop branch.


jvoegele:
+1 79aa59d

jvoegele:
@borshop merge

PR #672, opened 2016-08-23
Fix Sinusoidal performance issue and refactor Solrq/Helper/Vnode relationship.

JeetKunDoug:
- Sinusoidal Performance issue fix:
  Added a callback to yz_index_hashtree:update exactly the same as the
  callback added on the KV side in 2.0.7, because once the snapshot is
  completed, we needed to (and can) allow the Solr queues to continue
  writing to Solr.
- Refactor to partition/index per queue:
  By making Solr queues responsible for only a single {partition, index}
  pair we greatly simplify the code of the queue and the helper. Rather
  than a pool of queues, and a pool of helpers, we now have one
  queue/helper pair per partition/index. Beyond code clarity, this
  provides many benefits:
  - We can now drain by partition, which means we can do parallel
    exchanges again.
  - If we no longer own a partition, we can simply stop that pair
    (which now has its own supervisor as well).
  - All the code dealing with `dict`s of indexq records is now gone - each
    solrq has only a single index/partition, and the indexq record is now
    gone (collapsed into the main state record).
  - Additional refactoring around the interaction between helper and
    worker, simplifying handling of draining further.
- Things we tried in fixing the performance issue that didn't work:
  - Throttling batches to Solr based on current in-flight queue lengths.
    This just allowed the queues to back up more.
  - Attempted to introduce backpressure by using the `riak_core_throttle`
    module in a pre-commit hook in `riak_kv_put_fsm`. This failed to resolve
    the issue, and in fact made it worse as we would hold up a large number
    of requests and then dump them all into KV and the Solr queues.


riak_test/yokozuna_essential.erl
```diff
@@ -79,6 +79,7 @@ confirm() ->
             verify_deletes(Cluster2, ?INDEX, ?NUM_KEYS, KeysDeleted),
             ok = test_escaped_key(Cluster2),
             verify_unique_id(Cluster2, PBConns),
+            %% verify_delete_index(Cluster2, PBConns),
```
nickelization: Do we want this to be commented? If so, it would be nice to also have a comment explaining when/why someone might want to uncomment this part of the test. If not, perhaps we should uncomment it :).

JeetKunDoug: Yes - I need to get this test passing. It was added to validate that we stop the appropriate solrq pairs when the index for which they were started is removed, but got lost as we continued to refactor things.

riak_test/yz_solrq_test.erl
```diff
@@ -312,33 +310,22 @@ confirm_no_contenttype_data(Cluster, PBConn, BKey, Index) ->
...
     confirm_purge_none_strategy(Cluster, PBConn,
-                                {?BUCKET11, ?INDEX11},
-                                {?BUCKET12, ?INDEX12}),
+                                {?BUCKET11, ?INDEX11}),
     ok.
 
-confirm_purge_one_strategy(Cluster, PBConn, Bucket1Index1, Bucket2Index2) ->
-    PurgeResults = do_purge(Cluster, PBConn, Bucket1Index1, Bucket2Index2,
+confirm_purge_one_strategy(Cluster, PBConn, Bucket1Index1) ->
+    PurgeResults = do_purge(Cluster, PBConn, Bucket1Index1,
```
nickelization: Minor formatting nitpick, but with the extra argument deleted, this doesn't need to be split across two lines anymore.

riak_test/yz_solrq_test.erl
```diff
@@ -347,27 +334,22 @@ check_one_purged({{Index1Written, Index1SearchResults},
     ?assertEqual(Condition, true),
     ok.
 
-confirm_purge_idx_strategy(Cluster, PBConn, Bucket1Index1, Bucket2Index2) ->
-    PurgeResults = do_purge(Cluster, PBConn, Bucket1Index1, Bucket2Index2,
+confirm_purge_idx_strategy(Cluster, PBConn, Bucket1Index1) ->
+    PurgeResults = do_purge(Cluster, PBConn, Bucket1Index1,
```
nickelization: And another place we could consolidate to one line.

riak_test/yz_solrq_test.erl
```diff
@@ -448,48 +402,51 @@ first_purged([_H|T] = _Written, Searched) ->
 %%
 do_purge([Node|_] = Cluster, PBConn,
          {Bucket1, Index1},
```
nickelization: Since we're removing `Bucket2` and `Index2`, could we remove the 1s that are leftover in all the variable names in this function?

src/yz_kv.erl
```diff
@@ -141,8 +141,15 @@ get_md_entry(MD, Key) ->
 -spec get_index(bkey()) -> index_name().
 get_index({Bucket, _}) ->
     BProps = riak_core_bucket:get_bucket(Bucket),
+    get_index_from_bucket_props(BProps).
+
+get_index_from_bucket_props(BProps) ->
     proplists:get_value(?YZ_INDEX, BProps, ?YZ_INDEX_TOMBSTONE).
 
+is_search_enabled_for_bucket(BucketProps) ->
+    yz_kv:get_index_from_bucket_props(BucketProps) =/=
```
nickelization: Not super important, but we don't need to qualify this call with the module name - could save a tiny bit of verbosity and remove it if you wanted.

src/yz_solrq_helper.erl
```diff
@@ -285,23 +270,21 @@ get_ops_for_entry_action(Action, _ObjValues, LI, P, Obj, BKey,
...
+    case yz_solr:index_batch(Index, prepare_ops_for_batch(Ops)) of
+        ok ->
+            yz_stat:index_end(Index, length(Ops), ?YZ_TIME_ELAPSED(T1)),
+            ok;
+        {error, {Reason, _Detail}} = Err when Reason =:= badrequest; Reason =:= bad_data ->
             ?DEBUG("batch for index ~s failed.  Error: ~p~n", [Index, Err]),
-            case Err of
-                {_, Reason, _} when Reason =:= badrequest; Reason =:= bad_data ->
+            yz_stat:index_fail(),
                     handle_bad_entries(Index, Ops, Entries);
```
nickelization: ^ Indentation ^

src/yz_solrq_helper.erl
```diff
@@ -321,30 +304,33 @@ handle_bad_entries(Index, Ops, Entries) ->
...
-                              ?ERROR("Updating a single Solr operation failed for index ~p with bad request.", [Index]),
-                              true;
-                          _ ->
-                              ?ERROR("Updating a single Solr operation failed for index ~p with error ~p", [Index, Err]),
-                              yz_fuse:melt(Index),
-                              false
-                      end
-              end
-      end, Ops).
+  lists:takewhile(fun(Op) ->
```
nickelization: It seems like we normally standardize on 4 spaces per level of indentation in our Erlang code, but indentation here and elsewhere in this PR is only using 2 spaces. I feel like we should probably stick with 4 spaces to keep things consistent.

src/yz_solrq_queue_pair_sup.erl
```diff
@@ -0,0 +1,91 @@
...
+%% specifications.
+%%
+%% @end
+%%--------------------------------------------------------------------
+-spec(init(Args :: term()) ->
+    {ok, {SupFlags :: {RestartStrategy :: supervisor:strategy(),
+        MaxR :: non_neg_integer(), MaxT :: non_neg_integer()},
+        [ChildSpec :: supervisor:child_spec()]
+    }} |
+    ignore).
```
nickelization: In practice we never return `ignore` from this function, so we could perhaps tighten up the type spec and remove that as a possibility. (Same goes for the spec for `start_link` above - we don't need to spec out all the possible return values of `supervisor:start_link`, only the ones we might actually return in practice.)

src/yz_solrq_queue_pair_sup.erl
```diff
@@ -0,0 +1,91 @@
...
+    WorkerChild = worker_spec(Index, Partition),
+
+    {ok, {SupFlags, [HelperChild, WorkerChild]}}.
+
+%%%===================================================================
+%%% Internal functions
+%%%===================================================================
+
+
+helper_spec(Index, Partition) ->
```
nickelization, line 84: Could we factor out the common code from `helper_spec` and `worker_spec`? Not critical really, but I hate seeing so much common code copied and pasted like that between two functions that are right next to each other.

src/yz_solrq_sup.erl
```diff
@@ -20,70 +20,112 @@
...
-
+-spec start_queue_pair(Index::index_name(), Partition::p()) -> ok.
+start_queue_pair(Index, Partition) ->
+    validate_child_started(
+        supervisor:start_child(?MODULE, queue_pair_spec({Index, Partition}))).
+
+active_queues() ->
+    PairSups = find_pair_supervisors(),
+    PairChildren = lists:flatten([supervisor:which_children(Sup) || Sup <- PairSups]),
+    Workers = [{Index, Partition} || {{worker, Index, Partition}, _Child, _Type, Modules} <- PairChildren, Modules == [yz_solrq_worker]],
```
nickelization: We should break up the super long lines here and in `find_pair_supervisors` below.

src/yz_solrq_worker.erl
```diff
@@ -171,89 +167,91 @@ cancel_drain(QPid) ->
...
-    {noreply, NewState} = handle_cast(drain_complete, State),
+    NewState = handle_drain_complete(State),
     {reply, ok, NewState};
-handle_call(all_queue_len, _From, #state{all_queue_len=Len} = State) ->
+handle_call(all_queue_len, _From, #state{queue = Queue} = State) ->
+    Len = queue:len(Queue),
     {reply, Len, State};
 handle_call(reload_appenv, _From, State) ->
     {reply, ok, read_appenv(State)}.
-handle_cast({request_batch, Index, HPid}, State) ->
```
nickelization, line 254: This clause gets deleted, but then doesn't appear to be replaced by a new one like I'd expect. It looks like `request_batch/2` will end up causing the worker to crash since there's no `handle_cast` clause that will match.

src/yz_solrq_drain_mgr.erl
```diff
@@ -65,35 +65,46 @@ init([]) ->
     schedule_tick(),
```
nickelization, line 65: Not related to this changeset, but just something I noticed during reviews: the implementation of `schedule_tick` down below uses the `?MODULE` macro to send a message to ourselves. This will work fine, but we already define a `?SERVER` macro to avoid the code being coupled to the module name, so we should probably use that instead.

src/yz_solrq_sup.erl
```diff
@@ -20,70 +20,112 @@
...
-                        Name <- tuple_to_list(yz_solrq:get_solrq_helper_tuple())],
+
+    QueueChildren = [queue_pair_spec(IndexPartition) ||
+                        IndexPartition <- required_queues()],
     %% Using a one_for_all restart strategy as we write data to a hashed worker,
     %% which then uses a random helper to send data to Solr itself - if we want to
     %% make this one_for_one we will need to do more work monitoring the processes
     %% and responding to crashes more carefully.
-    {ok, {{one_for_all, 10, 10}, [DrainMgrSpec | HelperChildren ++ QueueChildren]}}.
+    {ok, {{one_for_all, 10, 10}, [DrainMgrSpec | QueueChildren]}}.
```
nickelization: Can we remove the above comments and make this a `one_for_one` supervisor? It seems like we should be able to, since each worker now has its own personal helper which are running together under a `one_for_all` supervisor.

nickelization: I think we could also rename `QueueChildren` to something like `PairSupChildren` for clarity as to what's actually being started here.

src/yz_solrq_worker.erl
```diff
@@ -171,89 +167,91 @@ cancel_drain(QPid) ->
...
-    IndexQ2 = maybe_request_worker(Index,
-                                   IndexQ#indexq{batch_min = Min,
-                                                 batch_max = Max,
-                                                 delayms_max = DelayMS}),
-    OldParams = {IndexQ#indexq.batch_min,
-                 IndexQ#indexq.batch_max,
-                 IndexQ#indexq.delayms_max},
-    {reply, {ok, OldParams}, update_indexq(Index, IndexQ2, State)};
+handle_call({set_index, Min, Max, DelayMS}, _From,
+            State0) ->
```
nickelization: Probably could just put the above function head on one line instead of splitting it. Also missing a space before the equals sign on the next line below.

src/yz_solrq_worker.erl
```diff
@@ -393,110 +385,58 @@ code_change(_OldVsn, State, _Extra) ->
...
--spec timer_running(#indexq{}) -> boolean().
-timer_running(#indexq{timer_ref = undefined}) -> false;
-timer_running(_) -> true.
+%%
+%% @doc
+%% When we have to retry it does not matter what draining state we are in,
+%% we simply have to requeue and continue.
+%% end
+%%
+handle_batch({retry, Undelivered},
```
nickelization: Could un-break this line.

test/yz_solrq_eqc.erl
```diff
@@ -141,15 +141,16 @@ prop_ok() ->
...
+                    {Helper, IBrowseKeys, MeltsByIndex},
                     begin
                         reset(), % restart the processes
                         unlink_kill(yz_solrq_worker_0001),
                         unlink_kill(yz_solrq_helper_0001),
                         unlink_kill(yz_solrq_eqc_fuse),
                         unlink_kill(yz_solrq_eqc_ibrowse),
-                        {ok, SolrQ} = yz_solrq_worker:start_link(yz_solrq_worker_0001),
-                        {ok, Helper} = yz_solrq_helper:start_link(yz_solrq_helper_0001),
+                        %%{ok, SolrQ} = yz_solrq_worker:start_link(yz_solrq_worker_0001),
```
nickelization: Should we delete these commented lines? Just from skimming, it looks like maybe they're not needed anymore.

test/yz_solrq_eqc.erl
```diff
@@ -159,11 +160,11 @@ prop_ok() ->
                         wait_for_vnodes(Pids, timer:seconds(20)),
                         timer:sleep(500),
                         catch yz_solrq_eqc_ibrowse:wait(expected_keys(Entries)),
-                        {SolrQ, Helper,  yz_solrq_eqc_ibrowse:keys(), melts_by_index(Entries)}
+                        {Helper,  yz_solrq_eqc_ibrowse:keys(), melts_by_index(Entries)}
                     end,
                     ?WHENFAIL(
                         begin
-                            eqc:format("SolrQ: ~p\n", [SolrQ]),
+                            %% eqc:format("SolrQ: ~p\n", [SolrQ]),
```
nickelization: I think we could likely just delete this as well, instead of commenting it out.

src/yz_solrq_helper.erl
JeetKunDoug: Yea - I think @lehoff's editor is set for 2 spaces... will try to clean up where I can.

src/yz_solrq_queue_pair_sup.erl
JeetKunDoug: true 'dat

src/yz_solrq_queue_pair_sup.erl
JeetKunDoug: Funny - I have gone both ways with this (actually undid some of the "make this common"), but now that it's complete again, I see why getting it combined makes sense. Will do something like this:

``` erlang

helper_spec(Index, Partition) ->
    child_spec(helper, yz_solrq_helper, Index, Partition).

worker_spec(Index, Partition) ->
    child_spec(worker, yz_solrq_worker, Index, Partition).

child_spec(Type, Module, Index, Partition) ->
    {{Type, Index, Partition}, {Module, start_link, [Index, Partition]}, 
     permanent, 5000, worker, [Module]}.
```

src/yz_solrq_drain_mgr.erl
JeetKunDoug: will fix while I'm here.

src/yz_solrq_sup.erl
JeetKunDoug: Agree to both comments above. @lehoff thoughts on going from one_for_all to one_for_one here? I think we had discussed this when we made the `yz_solrq_queue_pair_sup` responsible for the pair?

test/yz_solrq_eqc.erl
JeetKunDoug: More importantly, we need to revisit this test altogether since we've changed the supervision structure enough that it doesn't pass any more. Forgot that this needs to be run with a `make clean && make pulse` otherwise the test doesn't run.

src/yz_solrq_worker.erl
JeetKunDoug: I swear I left a note here already, but it's gone, so here we go again.
Nice catch - we actually don't do the same dance with the helper we used to do, so the whole request_batch path here and in the helper is no longer used. I've removed it completely at this point.

fadushin:
+1 5b0bd74

fadushin:
+1 ce4439c

JeetKunDoug:
@borshop merge

PR #675, opened 2016-09-09
Changes job control settings to be commented.

fadushin:
This PR changes the job control settings added in 2.2 so that they are commented with their associated default values. Doing so reduces friction in the (unlikely) case that a new install is downgraded to a previous version.


jvoegele:
+1 df4a3cd

fadushin:
@borhop merge

fadushin:
Bors is gone fishing.  yolo-merging.

PR #676, opened 2016-09-14
Fixed yz_solr_upgrade_downgrade test 

fadushin:
so that it just copies out the downgraded solrconfig.xml, instead of trying to do surgery on the 4.10
version.


paulhenrich:
+1 a58e8f7

PR #681, opened 2016-09-19
Clear out solr-webapp directory when setting up yz_temp dir

paulhenrich:
On upgrade the existence of this directory and its cached solr classes
can cause yokozuna to fail to start. To prevent this, we double-check
the state of the temp directory on start.

I'm going to add a commit to change the `yz_solr_upgrade_downgrade` test to use `del_dir_recursively`, but wanted to open this so I could get some feedback on the code.


ian-mi:
I'm concerned about relying on the order of output from filelib:wildcard as I don't see the output order mentioned in the filelib documentation. Additionally, this method will be inefficient as it traverses the directory structure multiple times but it seems like this is unavoidable when using erlang's file path based operations. Still it may be better to rewrite this a true recursive function. Additionally, we might want to be careful to only delete regular files and to not follow symlinks.

paulhenrich:
Good points, especially re lack of documented order guarantee. I'll take another pass. Thanks for looking @ian-mi

fadushin (commented):
Let's also make sure that if filename:join(TempDir, "solr-webapp") doesn't exist, this is effectively a no-op.

paulhenrich:
@ian-mi @fadushin I implemented the recursive version, which worked, and then on feedback from @JeetKunDoug got rid of the explicit recursion in favor of explicitly sorting the directories from `wildcard`. It has the same effect and is a bit clearer.

I also tested manually with a number of directory structures including ones with symlinks to outside directories. I also updated the test to stop doing this step. Without my addition the test then fails. When I add back my call to del_dir_recursive, the test passes, showing that the implementation now does the right thing.

src/yz_solr_proc.erl
```diff
@@ -276,10 +276,32 @@ ensure_data_dir(Dir) ->
...
+    del_dir_recursive(Path, filelib:is_dir(Path)).
+
+-spec del_dir_recursive(string(), boolean()) -> ok | {error, enotdir}.
+del_dir_recursive(DirPath, true) ->
+    FullPaths = [filename:join(DirPath, FilePath) ||
+                    FilePath <- filelib:wildcard("**", DirPath)],
+    {Dirs, Files} = lists:partition(fun filelib:is_dir/1, FullPaths),
+    lists:foreach(fun file:delete/1, Files),
+    %% Delete directories sorted longest to shortest to ensure we delete leaf
+    %% directories first.
```
fadushin, line 292: sneaky

JeetKunDoug:
+1 1982a33

JeetKunDoug:
So we're on "thumbs" now - I'm going to rebase this while @paulhenrich is in transit so it picks up the thumbs.yml stuff needed to get thumbs happy (I think).

JeetKunDoug:
+1

JeetKunDoug:
Hrm - maybe Thumbs doesn't know about this because the PR existed before the Thumbs.yml was here? @ooshlablu @davidx any help would be appreciated.

davidx:
@JeetKunDoug Yes,  It looks like it didn't find a config when it first ran. At this time it'd be best  to expect the correct behavior on new prs only.

davidx:
@JeetKunDoug Also, the incoming pr branch 'ph-bugfix-clear-solr-temp-on-start' doesn't contain the config yet, (which i just merged very recently :)
if you were to push the addition of the .thumbs.yml to this pr, it would trigger a build status report

paulhenrich:
I rebased again so it should now have .thumbs.yml

paulhenrich:
Make test passes for me locally. Not sure what's going on with Thumbs.

JeetKunDoug:
Looks like test failures are due to changes in riak_kv and riak_core that have been merged - yz_solrq_eqc will probably need to set up a few more things.

JeetKunDoug:
+1

davidx:
Great 👍     Because merge:false in the .thumbs.yml it won't automerge.  it also is currently configured to need 2 people other than the author, to say +1 , hence it is not counting additional +1s.

davidx:
It is now in a state where it would otherwise automerge the PR, having all passing build steps and 2 nonauthor +1s

JeetKunDoug:
So, because I pushed something, I'm now an author? Or not?

davidx:
no, paulhenrich is the author, your review is counted for sure. see show/hide(1 of 2 code reviews from org basho)

ian-mi:
+1

davidx:
there's no additional messaging until the merge occurs, which won't occur until the safeguard merge:false is set to merge:true.  If an additional push occurred, you'd see the build status include the reviewers, which is currently 3.

JeetKunDoug:
cool - at this point, just going to merge this one...

PR #682, opened 2016-09-21
Encapsulated default values for several new configuration variables into macros.

fadushin:
This PR was motivated by inconsistent default values for configuration variables in cuttlefish and in the code (and with different defaults in some parts of the code!).  This can cause unwanted behavior when cuttlefish is not being used, for example, in an upgrade case where the user is still using the older app.config style configuration.  Technically, we should also be prepared for the case where the library is NOT being used in the context of cuttlefish configuration, even through practically speaking cuttlefish is (almost) always in effect.

This allows any unspecified values take on consistent defaults.  There is still a point of management between the defaults in code and the cuttlefish defaults, but these can now be tracked in two files.  Also removed some stale macro definitions that were no longer used.

### DIFF
--- a/include/yokozuna.hrl
+++ b/include/yokozuna.hrl
@@ -494,12 +494,13 @@
 %%%===================================================================
 
 -define(SOLRQ_BATCH_MIN, solrq_batch_min).
+-define(SOLRQ_BATCH_MIN_DEFAULT, 10).
 -define(SOLRQ_BATCH_MAX, solrq_batch_max).
+-define(SOLRQ_BATCH_MAX_DEFAULT, 500).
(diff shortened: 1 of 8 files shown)


JeetKunDoug:
+1

nickelization:
The PR and commit description suggest that this is just a refactor, moving the defaults into macros. But it looks like the defaults are actually being changed significantly in a couple cases. Is this unintentional, or just unmentioned?

fadushin:
@nickelization Please see the re-write of the PR text.

nickelization:
+1

JeetKunDoug:
+1 0220cca

JeetKunDoug:
Merging as Thumbs is not yet configured for auto-merge.

PR #684, opened 2016-09-28
Update defaults for soft commit and batching flush interval to more closely match pre-batching behavior

JeetKunDoug:
This commit makes the combination of Solr's `softCommit` and our `solrq_batch_flush_interval` add up to 1 second, which was previously the interval for the `softCommit` before we introduced batching to Yokozuna. This will make a default installation more closely match the previous behavior of 1 second maximum delay between `put` and searchability.


fadushin:
+1

paulhenrich:
+1

PR #685, opened 2016-09-29
Removed the verify_no_repair_for_non_indexed_data test.

fadushin:
The reasoning for the removal of this test is as follows:

Due to non-atomicity of snapshotting of the riak_kv AAE hash tree and the draining of solrqs, there is always a window of time where data can be written into a queue AFTER a KV hash tree has been snapshotted but before a drain initiates, resulting in data showing up in a YZ AAE hash tree but NOT in the KV snapshot.  This results in a "false" repair -- data showing up in Solr that is not in KV, which subsequently gets removed, and then re-added at the next repair for that partition.

We are tolerant of this race condition, as it is rare and will result in a small amount of spurious repair.  The test, therefore, is testing a condition we cannot guarantee, and the test does fail intermittently for this reason.  We have therefore decided to remove this test from the suite.

### DIFF
--- a/riak_test/yz_aae_test.erl
+++ b/riak_test/yz_aae_test.erl
@@ -130,8 +130,6 @@ aae_run(Cluster, Bucket, Index) ->
 
             verify_no_indefinite_repair(Cluster),
 
-            verify_no_repair_for_non_indexed_data(Cluster, Bucket, PBConns),
-
             verify_count_and_repair_after_error_value(Cluster, Bucket, Index,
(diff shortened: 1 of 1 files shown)


hazen:
@fadushin Would it be possible to change the test to query for data availability? Or do you feel we already have enough coverage?

hazen:
:+1:

bsparrow435 (approved):

hazen:
No auto merge. 😢

PR #686, opened 2016-09-29
Removing the logging of DOWN messages from processes we are monitoring.

fadushin:
The rationale for this change is as follows:

If the Info passed in the DOWN message is 'normal', then there is nothing to log, and it's just an annoyance to the user.
Otherwise, if the Info is "non-normal", it should be picked up by SASL logging elsewhere, so logging it is not providing any additional information for diagnostics.


ian-mi:
+1

ooshlablu:
:+1:

PR #693, opened 2016-10-10
When destroying trees, do not update them first. [JIRA: RIAK-2812]

JeetKunDoug:
Updating a large hashtree can take several minutes, and in the case of
expired tree rebuilds, all we're going to do is delete the files right
after the update anyway. This fixes an issue where the hashtree would
be blocked for the entire time the clear/update was happening.


bashopatricia:
create jira issue

fadushin:
+1

PR #696, opened 2016-10-13
Fix node crash issue caused by unhandled non-error [JIRA: RIAK-2891]

paulhenrich:
When deleting an index, we were only considering 200 a success.
However, if we tried to delete an index that didn't exist, solr
would return a 404, which we didn't handle. Now we consider that
a success as well.


jvoegele:
+1 f313804

bashopatricia:
create jira issue

bsparrow435:
@paulhenrich 👍

javajolt:
:+1:

hazen:
:+1: cd571de

ooshlablu:
@javajolt @bsparrow435 @paulhenrich @jvoegele see my updates to the thumbs comments above. This is good to go

PR #697, opened 2016-10-21
Fix issues with drain messages and pending processes

JeetKunDoug:
This branch contains fixes for two issues:
1) We were sending `drain_complete` and `cancel_drain` messages to all solrqs, rather than just those for a particular partition.
2) While there is a solrq worker per {partition, index} pair, the `yz_exchange_fsm` can _also_ directly index data to Solr without going through the vnode. This caused us to overwrite the `pending_vnode` field in the `yz_solrq_worker`'s `state` record if an exchange was ongoing and both the vnode and the exchange FSM blocked due to the queue growing to the high water mark. Reinstated the `pending_processes` list, rather than just a single `pid()`, and handled sending the messages to unblock appropriately.


hazen:
Weird.  `ecaec7007afa4c2f8f73e96bbe4f858a70a35737` passes all tests by hand

PR #700, opened 2016-11-01
Performance fixes for Solr integration [JIRA: RIAK-2955]

JeetKunDoug:
# Several major performance improvements for Solr integration:
- Remove Delete By Query calls whenever possible as it's a performance killer on Solr
- Split hashtree update out of the "Drain Timeout" window as we were killing hashtree builds unnecessarily.
- Fix issue where `yz_index_hashtree:fold_keys` was flooding the yz_index_hashtree mailbox by implementing the "90 async/1 sync" call pattern as other `index` and `delete` callers used.

## Removed usages of Delete By Query
Prior to this fix, whenever we would add a document to Solr that might contain siblings, we would add a delete by query operation to the batch we send to Solr (in a single HTTP request).  This had a significant impact on performance, especially when objects being updated had a lot of siblings.  (We found that deleting by query would contend for locks with the segment scheduler, and in general adding deletesByQuery to high workloads seems like a bad idea).

The fix involves changing deleteByQuery to deleting documents in Solr by their document ids, which has much better performance than deleting by query..  The challenge is that when adding an object, we didn't have the sibling vtags to generate a document id.  We therefore made a change to riak_kv (basho/riak_kv#1520) which, when updating an object, also sends the old object being replaced.  From that old object, we can extract any siblings and generate associated document ids to delete in Solr.

## Hashtree update/drain timeout split
The process of doing a KV->Yokozuna hastree comparison takes several steps, including draining the relevant Solr queues before snapshotting the Yokozuna hashtree so that its state can be as close as possible to the KV hashtree. In order to not block the exchange indefinitely, a "Drain Timeout" was included in this process. However, as initially developed, the drain was not the only thing being measured in this timeout window. It also included the hashtree:update call, which, on a heavily-loaded system, can take several minutes. This would cause the "drain" to time out and be forcibly cancelled, even though the drain itself had only taken a few hundreds of milliseconds.

`yz_drain_fsm`, `yz_drain_mgr`,  and `yz_exchange_fsm` were was reworked to split the Solr drain from the hashtree update, resolving the issue of drain timeouts under normal operating conditions.

## Hashtree asyc/sync change
During performance testing, it was discovered that the `yz_index_hashtree:fold_keys` 
function could swamp the mailbox of the `yz_index_hashtree` so that other processes 
could not make progress. This commit makes `fold_keys` use the same
"90 async + 1 sync" call pattern that `yz_kv:index` used. That logic
(which lived in `yz_kv`) has been moved to `yz_index_hashtree` and
shared by the new `index` and `delete` calls that do not take an explicit
"call mode" parameter.
Future work is scheduled to figure out how to make this pattern more generic 
(as it's used in several places) and to eliminate the use of the caller's
process dictionary for storing hashtree tokens.

## Minor improvements included in this Pull request

* Added logging for clear and exchange trees, for audit of administrative operations.
* Modified the "Will Repair" log in the yz_exchange_fsm to track the "direction" of repair, viz., whether the repair resulted in a delete of Solr data or an add/update to Solr.
* Corrected the list of `required_queues` in `yz_solrq_sup` to include both _current_ and _future_ indexes the node owned. This prevented premature shutdown of a solrq worker/helper pair that was started by writes to a _future_ vnode (one in the process of being handed off to the node in question).
* Remove exometer statistics when an index is removed. Before, if an index was later re-added, the fuse creation would fail, eventually causing the node to crash. Test updated in 315b7464fa9996e56237de21050fd5f7b1f4eb6d
* Monitor `yz_solrq_drain_fsm` from the queues being drained in case the drain_fsm crashes. Before, it was possible for a queue to get stuck in `wait_for_drain_complete` state if the drain fsm crashed before the drain complete messages were sent.

## Review Checklist:

1. [X] Observability
    - Observability was not changed in this PR - stats already existed for the Solr operations, and there were no major changes in logging required.
1. [X] Performance
    - This PR has been driven by performance testing conducted on AWS over the course of several weeks, and has been shown to improve performance dramatically (In our testing, 4X when writing siblings in large numbers).
1. [X] Fault-tolerance
    - Added some additional monitoring around the `yz_solrq_drain_fsm` to prevent queues from being stuck in `wait_for_drain_complete` state.
1. [X] Testing
    - Updated many tests to account for different behavior.
1. [X] Code Clarity/Quality
    - Wherever possible, code was refactored along the way and reviewed on an ongoing basis for clarity and quality.
1. [X] Code Documentation
    - Updated `batching.md` and associated docs. These can be used to dive External Documentation changes as well.
1. [X] External Documentation
    - Removed the `purge_all` strategy from `search.queue.high_watermark.purge_strategy` options as it no longer makes sense to use (each `yz_solrq_worker` now manages a single index/partition pair.)
    - Removed `search.queue.worker_count` and `search.queue.helper_count` as they are no longer applicable.
    - Added `search.dist_query` setting to riak.conf which will remove the node in question from distributed search results. This setting can also be changed via `riak-admin` by issuing the `riak-admin set search.dist_query=off` or `riak-admin set search.dist_query=on` commands at the command line.  Setting this value in `riak.conf` is useful when you are restarting a node which was removed from search queries with the above-mentioned `riak-admin` feature, to prevent the node from being included in search queries until it is fully spun up. This may be necessary if a re-indexing operation is under way, or, in the case of very large Solr indexes, the time it takes for Solr to read it's indexes off disk and be prepared for query.

1. [X] Release Notes
    - Make sure to mention the addition/removal of config settings above.
    - Added a strictly internal setting `{yokozuna, anti_entropy_max_async}` that mirrors a similar setting in riak_kv (and will take the KV value if the yokozuna value isn't provided), but the setting in riak_kv isn't documented and is essentially never changed, so this should not have to be called out in release notes as we doubt anyone would ever change it.


src/yz_entropy_mgr.erl
```diff
@@ -217,11 +217,13 @@ handle_cast({exchange_status, Pid, Index, {StartIdx, N}, Status}, S) ->
     {noreply, S2};
 
 handle_cast(clear_trees, S) ->
+    lager:info("Clearing YZ hashtrees and stopping all current exchanges..."),
```
bsparrow435: If these logs are going to go in, should remove the "..."

src/yz_solrq_helper.erl
```diff
@@ -186,62 +188,102 @@ update_solr(Index, LI, Entries) when ?YZ_SHOULD_INDEX(Index) ->
             %% yz_index:add_index/1 on 1st creation or diff-check.
             %% We send entries until we can ask again for
             %% ok | error, as we wait for the tick.
-            send_solr_ops_for_entries(Index, solr_ops(LI, Entries),
-                                      Entries)
+            LogicalPartition = logical_partition(LI, Entries0),
+            %% Entries = get_latest_entries_for_keys(Entries0),
```
bsparrow435: Can remove this line

src/yz_solrq_helper.erl
```diff
@@ -382,10 +425,77 @@ get_reason_action(Reason) when is_tuple(Reason) ->
...
--spec prepare_ops_for_batch(solr_ops()) -> solr_entries().
-prepare_ops_for_batch(Ops) ->
+-define(QUERY(Bin), {struct, [{'query', Bin}]}).
+
+-spec prepare_ops_for_batch(index_name(), non_neg_integer(), solr_ops()) -> solr_entries().
+prepare_ops_for_batch(_Index, _LogicalPartition, Ops) ->
     %% Flatten combined operators for a batch.
-    lists:flatten(Ops).
+    _FlattenedOps = lists:flatten(Ops).
+    %% Just for this experiment
```
bsparrow435: Can be removed

src/yz_solrq_helper.erl
```diff
@@ -382,10 +425,77 @@ get_reason_action(Reason) when is_tuple(Reason) ->
...
+        {q, Query}, {wt, json}, {fl, "_yz_id"}, {rows, Rows},
+        {sort, "_yz_id asc"}, {'cursorMark', CurrentCursor}]),
+    %% lager:info("FDUSHIN> JSonResponse: ~p", [JSonResponse]),
+    NextCursor = kvc:path(<<"nextCursorMark">>, JSonResponse),
+    NumFound = kvc:path([<<"response">>, <<"numFound">>], JSonResponse),
+    case {Iteration, NumFound =< Rows} of
+        {1, true} ->
+            Func(JSonResponse, Accum);
+        _ ->
+            lager:info("FDUSHIN> recursive call.  NumFound: ~p Iteration: ~p", [NumFound, Iteration]),
```
bsparrow435: Testing logging, noting so it will be removed

src/yz_exchange_fsm.erl
```diff
@@ -34,6 +34,8 @@
                 built :: integer(),
                 timeout :: pos_integer()}).
 
+-type repair_count() :: {non_neg_integer(), non_neg_integer()}.
```
lehoff: Is it painfully obvious what the two elements of the `repair_count()` tuple is for?
Consider naming the type variables:

```
-type repair_count() :: {SomeCount :: non_neg_integer(), AnotherCount :: non_neg_interger()}.
```

src/yz_exchange_fsm.erl
```diff
@@ -222,6 +218,34 @@ exchange_bucket_kv(Tree, IndexN, Level, Bucket) ->
...
+        fun(KeyDiff, InnerAccum) ->
+            repair_fold_func(Index, KeyDiff, InnerAccum)
+        end,
+        Accum,
+        KeyDiffs
+    ).
+
+-spec repair_fold_func(p(), keydiff(), repair_count()) ->
+    repair_count().
+repair_fold_func(Index, {remote_missing, _KeyBin} = KeyDiff, {YZDeleteCount, YZRepairCount} = Accum) ->
```
lehoff: I would split this into two functions:

```
repair_fold_func(Index, KeyDiff, Accum) ->
    update_repair_func_accum(repair(Index, KeyDiff), KeyDiff, Accum).

update_repair_func_accum(full_repair, _KeyDiff={remote_missing, _KeyBin}, {YZDeleteCount, YZRepairCount}) ->
    {YZDeleteCount + 1, YZRepairCount};
update_repair_func_accum(full_repair, _KeyDiff, {YZDeleteCount, YZRepairCount}) ->
    {YZDeleteCount, YZRepairCount + 1};
update_repair_func_accum(_RepairType, _KeyDiff, Accum) ->
    Accum.
```

src/yz_solrq_helper.erl
```diff
@@ -251,20 +293,21 @@ get_ops_for_entry_action(Action, _ObjValues, LI, P, Obj, BKey,
 %%      allowing it to update to tree to prevent forever repair) or it reaches
 %%      a Solr Internal error of sorts, for which we stop list-processing and
 %%      use the success-length to segment the entries list for `AAE-updates'.
--spec send_solr_ops_for_entries(index_name(), solr_ops(), solr_entries()) ->
+-spec send_solr_ops_for_entries(index_name(), non_neg_integer(), solr_ops(), solr_entries()) ->
```
lehoff: Should `non_neg_integer()` really be `p()`?
If not, please introduce a suitably named type alias.

src/yz_solrq_helper.erl
```diff
@@ -382,10 +425,77 @@ get_reason_action(Reason) when is_tuple(Reason) ->
...
+get_doc_ids(Index, Query) ->
+    fold_query(Index, Query, undefined, <<"*">>, 100, fun doc_ids_from_response/2, [], 1).
+
+fold_query(_Index, _Query, Cursor, Cursor, _Rows, _Func, Accum, _Iteration) ->
+    Accum;
+fold_query(Index, Query, _PreviousCursor, CurrentCursor, Rows, Func, Accum, Iteration) ->
+    JSonResponse = search(Index, [
+        {q, Query}, {wt, json}, {fl, "_yz_id"}, {rows, Rows},
+        {sort, "_yz_id asc"}, {'cursorMark', CurrentCursor}]),
+    %% lager:info("FDUSHIN> JSonResponse: ~p", [JSonResponse]),
```
lehoff: This one should go...

src/yz_entropy_mgr.erl
```diff
@@ -217,11 +217,13 @@ handle_cast({exchange_status, Pid, Index, {StartIdx, N}, Status}, S) ->
...
     clear_all_exchanges(S#state.exchanges),
     clear_all_trees(S#state.trees),
     {noreply, S};
 
 handle_cast(expire_trees, S) ->
+    lager:info("Expiring YZ hashtrees..."),
```
bsparrow435: Same comment on "..."

src/yz_solrq_helper.erl
```diff
@@ -395,3 +505,50 @@ terminate(_Reason, _State) ->
...
+%% @doc Entries is [{BKey, Obj, Reason, P, ShortPL, Hash}]
+%% These are in chronological order - we need to get the latest for each BKey
+get_latest_entries_for_keys(Entries) ->
+    EntryDict = lists:foldl(fun entry_given_previous_entries/2, dict:new(), Entries),
+    dict:fold(fun add_value_to_list/3, [], EntryDict).
+
+add_value_to_list(_Key, Value, List) ->
+    [Value | List].
+
+entry_given_previous_entries(Entry, PreviousEntries) ->
```
lehoff: This would be simpler if we leverage `dict:update/4` along these lines:

```
entry_given_previous_entries(Entry, PreviousEntries) ->
    NewReason = get_reason_from_entry(Entry),
    BKey = get_bkey_from_entry(Entry),
    dict:update(BKey,
                        fun (PreviousEntry) ->
                               PreviousReason = get_reason_from_entry(PreviousEntry),
                               case choose_entry(NewReason, PreviousReason) of
                                   new -> Entry;
                                  _  -> PreviousEntry
                               end
                         end,
                         Entry,
                         PreviousEntries).
```

Having written this I would even say that we should let `choose_entry/2` take Entries as argument and rename the existing `choose_entry/2` to `compare_entry_reasons/2`

```
choose_entry(Entry, PreviousEntry) ->
     NewReason = get_reason_from_entry(Entry),
     PreviousReason = get_reason_from_entry(PreviousEntry),
     case compare_entry_reasons(NewReason, PreviousReason) of
                                   new -> Entry;
                                  _  -> PreviousEntry
                               end.
```

And then we get:

```
entry_given_previous_entries(Entry, PreviousEntries) ->
    BKey = get_bkey_from_entry(Entry),
    dict:update(BKey,
                        fun (PreviousEntry) ->
                              choose_entry(Entry, PreviousEntry)
                        end,
                        Entry,
                        PreviousEntries).
```

lehoff (dismissed):

src/yz_solrq_helper.erl
JeetKunDoug: Ends up it's `lp()`. Will address.

JeetKunDoug: even better, the parameter isn't really needed - will clean up dead code.

JeetKunDoug (commented):
Updated per review comments.

lehoff:
+1

bashopatricia:
create jira issue

davidx:
thumbot retry

davidx:
thumbot retry

docs/BATCHING.md
```diff
@@ -66,11 +56,10 @@ In the undesirable cases where Solr becomes unresponsive (e.g., data corruption,
 
 If a Solr core has become unresponsive and the specified error threshold has been traversed, and if, in addition, the high water mark has been exceeded, then the yokozuna batching system has a mechanism for automatically purging enqueued entries, so as to allow vnodes to continue servicing requests, as well as to allow update operations to occur for indices that are not in a pathological state.
 
-The yokozuna batching subsystem supports 4 different purge strategies:
+The yokozuna batching subsystem supports 3 different purge strategies:
 
 * `purge_one` (default behavior): Purge the oldest entry from a randomly selected indexq structure among the set of search indexes which have crossed the error threshold;
```
JeetKunDoug: This no longer is "a randomly selected indexq structure" as there are A) no indexq structures and B) it'll always purge from the {index, partition} to which the new entry is written.

docs/BATCHING.md
```diff
@@ -66,11 +56,10 @@ In the undesirable cases where Solr becomes unresponsive (e.g., data corruption,
...
 * `purge_index`: Purge all entries from a randomly selected indexq structure among the set of search indexes which have crossed the error threshold;
```
JeetKunDoug: Same as above - will purge all of the entries for the {index, partition} pair. ? is, should it actually purge the _whole index_ or is the current behavior of only purging the one solrq_worker "correct enough"? I believe it is, as this really previously would purge all the data for a single indexq structure _for a single worker_ as it does today.

JeetKunDoug (commented):
Just a few minor nits - I can actually take care of these myself at this point if you're otherwise busy.

JeetKunDoug:
+1 edf1413

PR #703, opened 2016-11-10
Prevent fallback vnodes' solrqs from stopping if they are still needed

JeetKunDoug:
Before, we used the "current and next" vnode list according to the ring
to determine which partitions were needed for solrq start/stop. This
change simply asks `riak_core_vnode_manager` which `riak_kv_vnode`
vnodes are currently running, and uses that list to figure out which
queues we need to have running.
This should prevent issues where fallback vnodes currently running on a
node would create many short-lived solrqs, which would start on demand
when an index call was made that needed them, and stop every tick.

### DIFF
--- a/src/yz_solrq_sup.erl
+++ b/src/yz_solrq_sup.erl
@@ -104,14 +104,14 @@ validate_child_started(Error) ->
     throw(Error).
 
 required_queues() ->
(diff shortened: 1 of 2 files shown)


fadushin:
+1

PR #706, opened 2016-11-23
Prefer relative URL in image in yz batching doc

lucafavatella:


lucafavatella:
BTW without this change the hardcoded URL "https://raw.githubusercontent.com/basho/yokozuna/docs/yz-batching-overview.png" returns 404.

fadushin:
Thank you, Luca!!  I will back port these to the 2.0 and develop-2.2 branches, as well, as those branches are still under active development and support.

+1

PR #709, opened 2016-11-30
Update resource link to currently integrated Solr version

lucafavatella:


JeetKunDoug:
Thanks for the update.

PR #716, opened 2016-12-12
Removed explicit YZ dependency in riak_kv

fadushin:
This PR is coupled with https://github.com/basho/riak_kv/pull/1571, which removes the explicit compile- and run-time dependency on yokozuna from riak_kv_vnode.

In the Yokozuna repo, this means adding indexing operations now expected by riak_kv_vnode, and consolidating them in the yokozuna module, as a reasonable entry point into the application.  This PR also Adds hidden configuration into cuttlefish which will configure riak_kv to send indexing operations through this module.

C.f. https://github.com/basho/riak_kv/pull/1571 for PR checklist.

### DIFF
--- a/src/yokozuna.erl
+++ b/src/yokozuna.erl
@@ -27,6 +27,26 @@
 %%% API
 %%%===================================================================
 
+%% @doc Index a Riak object, given a reason and partition under which
+%%      the object is being indexed.  The object pair contains the new
+%%      and old objects, in the case where a read-before-write resulted
+%%      in an old object we can use to delete siblings before updating.
+-spec index(object_pair(), write_reason(), p()) -> ok.
+index(RObjPair, Reason, Idx) ->
+    yz_kv:index(RObjPair, Reason, Idx).
+
+%% @doc Index a Riak object encoded as a n erlang binary.  This function
+%%      is typically called from the write-once path, where there is no
+%%      old object to
+-spec index_binary(bucket(), key(), binary(), write_reason(), p()) -> ok.
+index_binary(Bucket, Key, Bin, Reason, P) ->
+    yz_kv:index_binary(Bucket, Key, Bin, Reason, P).
+
+%% @doc Determine whether a bucket is searchable, based on its properties.
+-spec is_searchable(riak_kv_bucket:props()) -> boolean().
+is_searchable(BProps) ->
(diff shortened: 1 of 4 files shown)


src/yokozuna.erl
```diff
@@ -27,6 +27,26 @@
 %%% API
 %%%===================================================================
 
+%% @doc Index a Riak object, given a reason and partition under which
+%%      the object is being indexed.  The object pair contains the new
+%%      and old objects, in the case where a read-before-write resulted
+%%      in an old object we can use to delete siblings before updating.
+-spec index(object_pair(), write_reason(), p()) -> ok.
```
jvoegele: I like these doc comments and specs. However, I think they would be even better if they were done as callback definitions in a behavior module, as per my comment on https://github.com/basho/riak_kv/pull/1571

test/eqc_util.erl
```diff
@@ -113,6 +113,7 @@ start_mock_components(Options) ->
     riak_core_ring_manager:setup_ets(test),
     setup_mockring(RingSize),
     riak_kv_entropy_info:create_table(),
+    riak_core_capability:start_link(),
```
jvoegele, line 116: Was this change necessitated by the new `index_module` configuration parameter?

fadushin: Actually no, this fixes the EQC test, which is currently broken on develop

src/yokozuna.erl
```diff
@@ -23,6 +23,51 @@
...
+    update_binary/5,
+    requires_existing_object/1,
+    should_handoff/1
+]).
+
+%%%===================================================================
+%% riak_kv_update_hook behavior
+%%%===================================================================
+
+%% @doc see riak_kv_update_hook:update/3
```
jvoegele, line 40: Maybe use the `@see ...` tag instead of `@doc see ...` here?

http://erlang.org/doc/apps/edoc/chapter.html#ftag-see

jvoegele: Of course, we don't actually generate any HTML docs from these comments (do we?) so this is totally discretionary.

nickelization:
This all looks good. Once the corresponding riak_kv PR is merged, we can tell Thumbs to retry and make sure everything else passes.

fadushin:
@thumbot retry

fadushin:
@thumbot merge

PR #732, opened 2017-03-14
Added handling for upgrades from 2.0.8 and later releases

fadushin:
This PR adds handling for upgrade from Riak 2.0.8 and later (2.0.x) releases.

Riak 2.0.8 added a minor change to the default solrconfig.xml, which changed the hash of the config file, which prevented automatic updates.   This PR detects the hash for that specific version of the default config file, and does an upgrade, accordingly, if that version is detected.  This special case logic should ONLY be required for these two cases, and should not impose an ongoing maintenance burden.

The PR also fixes a regression in the rolling upgrade functionality in the upgrade/downgrade riak test, which now passes when run against a 2.0.x and 2.2.x devrel with fixes for the `riak-admin` tool.

### DIFF
--- a/riak_test/yz_rt.erl
+++ b/riak_test/yz_rt.erl
@@ -95,6 +95,7 @@
     riak_pb/1,
     rolling_upgrade/2,
     rolling_upgrade/4,
+    rolling_upgrade/5,
     run_bb/2,
     schema_url/2,
     search/4,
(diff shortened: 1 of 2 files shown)


bsparrow435:
+1 looks good

JeetKunDoug:
+1

JeetKunDoug:
@thumbot merge

PR #733, opened 2017-03-15
Fixes for failing riak tests.

fadushin:
yz_crdt was moved over from riak_test but did not include a call to yz_rt:load_intercept_code(Cluster), so the intercepts never loaded properly.

yz_extractors was failing to converge on AAE.  Bumped up sweep tick and concurrency to make AAE tree rebuilds occur more quickly, and forced version 1 hash trees at start.


JeetKunDoug:
+1

PR #739, opened 2017-08-25
Added workaround for missing Solr and yokozuna artifacts on AWS

fadushin:
This PR allows one to specify an alternate location for Solr and Yokozuna artifacts via the ARTIFACT_URL_PREFIX environment variable.  These artifacts have traditionally been located on AWS, but which are currently not present or accessible, can be provided by other parties.

Example:
export ARTIFACT_URL_PREFIX="https://files.tiot.jp/riak"
make


Bob-The-Marauder:
+1

PR #741, opened 2017-12-07
Fix broken eqc test

russelldb:
Also remove riak_search, and make some deps changes.

The main change is to the test though

"Many runs of `make test` resulted in `test/hashtree_ce.eqc` this patch
fixes the somehow unhandled case of hash differences for the same
key. The rareness of this event suggests the generators need tuning for
a smaller keyspace."


nickkeers:
+1.

Tests + dialyzer pass

PR #746, opened 2018-04-05
remove -XX:+UseStringCache

ghost:
The UseStringCache option is removed in Java 8 (default installation for the last year).
As a consequence the JVM crashes on startup, and consequently crashes Riak.
I suggest we leave this tuning for the operator to implement if they deem necessary and stick with (non-crashing) defaults.


russelldb:
+1 from @jplock is good enough for me. Merging.

russelldb:
tests fail, a little late, I know. Kind of assumed the +1 from @jplock meant he'd run them. PR incoming

jplock:
Sorry @russelldb, no I hadn’t ran the tests. I don’t use yokozuna I was just in favor of removing this directive as it’s deprecated and removed in later JDK’s.

PR #758, opened 2019-10-02
Yokozuna/KV isolation

martinsumner:
Cherry pick of https://github.com/basho/yokozuna/commit/b71590608fc8b1950c19a824fc88ed498bf41c2b

This adds the hook code so that yokozuna can be used with the isolated hooks in https://github.com/basho/riak_kv/tree/fd-remove-yz-dep-develop29

This a feature that was previously merged in to develop by basho old.  It allows for the removal of yokozuna references from riak_kv.  This will then allow in the future for riak_kv to be packaged without yokozuna in environments where solr is not used, and there is no future intention of using it.

The combined branches pass all yoko_all tests (except the solr_upgrade_downgrade test which tests and upgrade from a legacy solr version - but this also fails in develop-2.9).  Packaging riak without yokozuna passes all kv_all riak_test tests.

### DIFF
--- a/priv/yokozuna.schema
+++ b/priv/yokozuna.schema
@@ -332,6 +332,14 @@
   hidden
 ]}.
 
+%% @doc Set the riak_kv update hook to be the yokozuna module, which
+%%      implements the `riak_kv_update_hook` behavior
+{mapping, "search.update_hook", "riak_kv.update_hook", [
+    {datatype, atom},
+    {default, yokozuna},
+    hidden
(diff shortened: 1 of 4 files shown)


martinsumner:
@llelf do you still maintain an interest in yoko?  Do you want to review this change, and the related PR in riak_kv?

The intention is to move this into Riak KV 2.9.1 and then roll into 3.0

llelf (approved):

PR #759, opened 2019-10-09
Dialyzer fixes

ThomasArts:
Note that `make dialyzer` does complex stuff with dialyzer that does not work.

But `./rebar3 dialyzer` does work.
I believe we need to cleanup the `Makefile` to not use the "manual" way of invoking dialyzer on all subprojects created for riak_test.


martinsumner (approved):
All looks good

PR #767, opened 2020-04-29
Develop 3.0 292 solrupdate

martinsumner:
This brings in the uplift of the solr version, and some changes to allow riak_test to run on OTP 20+.

However, there still are riak_test failures.  The failures appear to fall into three broad categories:

- handling of unicode has changed (in OTP perhaps) so all tests of non-ascii keys/buckets/values are failing;
- changes in URL endpoints between solr versions
- other deprecated features (e.g. tests which include Javascript map/reduce)

I think this should still be merged in, then we need to make the decision whether to:

- exclude yokozuna from riak from develop-3.0 due to broken tests, but provide a branch with yokozuna which can be built from source
- include yokozuna in riak and invest the effort to fix the broken tests


martinsumner:
There is also an issue with needing to include basho_bench as a dep in order to get riak_tests to run (many yokozuna riak_tests require basho_bench).

martincox (approved):
As far as the Erlang stuff goes, which looks to be the usual OTP20+ / relx fixups, looks good to me. Can't really comment on the Java / Solr side of things, but happy for it to be merged.

PR #768, opened 2020-07-09
Replace shasum with sha1sum and sha512sum

martincox:
`/home/martincox/code/riak/_build/default/lib/yokozuna/tools/grab-solr.sh: line 62: shasum: command not found
===> Hook for compile failed!
`
https://github.com/basho/yokozuna/blob/8ac890cdb6f2859d35eed174a194f6112a0b530d/tools/grab-solr.sh#L62

I was seeing build errors in centos for yoko and it looked to be from the use of shasum, which as far as I can tell, isn't present in rhel/centos. I've replaced the calls with the appropriate sha*sum call.

### DIFF
--- a/tools/common.sh
+++ b/tools/common.sh
@@ -18,8 +18,8 @@ mk_sha()
 
     if type sha512t256 &>/dev/null; then
         sha512t256 $file > $file.sha512
-    elif type shasum &>/dev/null; then
-        shasum -a 512256 $file > $file.sha512
+    elif type sha512sum &>/dev/null; then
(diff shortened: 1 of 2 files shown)


martinsumner:
Riak develop 3.0 should point at this branch (https://github.com/basho/yokozuna/tree/develop-3.0-292-solrupdate) and have it commented out - so I'm surprised its causing you build problems.

I'm fine with the change, but I guess we need to get the solrupdate merged into 3.0 as well

martinsumner (approved):

PR #769, opened 2020-07-16
wget fails with certificate verification error.

sendtopms:
When Riak with Yoko build from develop-3.0 branch, wget fails with certificate verification error.


martincox:
Thanks for this. 👍

